PR #51, opened 2022-07-05
Section name validation

Billy99:
If you pass a bad section name when loading a program, an error is returned,
but the program metadata is never cleaned up, so subsequent calls fail.

Validate the section name before inserting to the list of extension programs.

Resolves: 31


dave-tucker (approved):
Thanks @Billy99! This looks great.

PR #55, opened 2022-07-05
WIP: Try and fix stale Maps and Dispatcher Programs

### LINKED ISSUE #56: Dispatcher program + some loaded maps are left orphaned
After Loading and unloading an xdp program which defines a map, such as the one in #53 , the dispatcher program is left attached to the specified interface and the program's map is orphaned. 

To recreate the stale map scenario simply load and unload a BPF program 
which defines a new map. 

An easy way to recreate is to checkout https://github.com/redhat-et/bpfd/pull/53, 
run the go example, exit it while leaving `bpfd` running, you'll see that the 
`stats` program is gone but it's map `xdp_stats_map` is left orphaned even after closing 
the FD. 

```
[astoycos@localhost bpfd]$ sudo bpftool map list name xdp_stats_map
733: percpu_array  name xdp_stats_map  flags 0x0
        key 4B  value 16B  max_entries 5  memlock 4096B
        pids bpfd(644765)
```

Clearly bpfd is still owning/ maintaing a reference to this even though the program that 
defined it (the xdp counter prog) no longer exists. 

https://github.com/redhat-et/bpfd/pull/55  Is a first attempt at fixing this however I don't know if it's the right method.

astoycos:
Tries to fix #56

Currently even when there are no programs
to load the dispatcher program and any previously defined maps are left
orphaned. This takes a stab
at ensuring we remove it when there's no more programs
left to be managed.

This PR manually tries to remove the dispatcher program if no more programs are left to be managed. 
It also tries to manually close all the map FD's defined by any program upon unloading. 

Currently the issue #56 still persists even with these fixes, I'm not 100% sure what's going on.

### DIFF
--- a/bpfd/src/server/bpf.rs
+++ b/bpfd/src/server/bpf.rs
@@ -16,6 +16,7 @@ use nix::{
     },
 };
 use uuid::Uuid;
+use nix::unistd::close;
 
 use crate::server::{
     config::{Config, XdpMode},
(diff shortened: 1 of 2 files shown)


dave-tucker:
Closing in favour of #58

PR #53, opened 2022-07-05
Fix go example

astoycos:
fixes #39 

This PR updates the gocounter example. Mainly it 
- Moves the bpf program's source code to this repo
- Integrates it with `cilium/ebpf`'s buit in build structure for compiling with `go generate` 
- Updates the bpf program to count bytes and packet count 
- Updates the go client to use mTLS 
- Updates the `certificates.sh` script to generate a cert explicitly for the go-counter example 

Also it fixes a bug where the xdp dispatcher program wasn't actually calling any attached programs due to a one off error (I can move that commit into it's own PR if desired EDIT::: Moved to -> https://github.com/redhat-et/bpfd/pull/54 This needs to merge first.

PR #60, opened 2022-07-06
xtask: Always build ebpf in release profile

dave-tucker:
Signed-off-by: Dave Tucker <dave@dtucker.co.uk>


Billy99:
@dave-tucker #53 didn't remove the `#[cfg(debug_assertions)]` and `#[cfg(not(debug_assertions))]` logic in https://github.com/redhat-et/bpfd/blob/main/bpfd/src/main.rs#L27. Since you are always building with `--release`, this needs to be fixed as well.

Issue #52, opened 2022-07-05
`unload` the last program and system hangs


Billy99:
Running in a F36 VM, if two programs are added to an interface and then 'unloaded`, I lose access to the VM and need to reboot it.

```
$ sudo ./target/debug/bpfctl load -p xdp -i ens3 --priority 55 -s "xdp" /home/bmcfall/src/xdp-tutorial/basic01-xdp-pass/xdp_pass_kern.o
f7fe8625-5453-4eca-bb0b-c9130f15d6ea
$ sudo ./target/debug/bpfctl load -p xdp -i ens3 --priority 60 -s "xdp" /home/bmcfall/src/xdp-tutorial/basic01-xdp-pass/xdp_pass_kern.o
909edd07-f023-4dd1-9fee-16b0975bffd6

$ sudo ./target/debug/bpfctl list -i ens3
ens3
xdp_mode: skb

0: f7fe8625-5453-4eca-bb0b-c9130f15d6ea
	name: "xdp"
	priority: 55
	path: /home/bmcfall/src/xdp-tutorial/basic01-xdp-pass/xdp_pass_kern.o
1: 909edd07-f023-4dd1-9fee-16b0975bffd6
	name: "xdp"
	priority: 60
	path: /home/bmcfall/src/xdp-tutorial/basic01-xdp-pass/xdp_pass_kern.o

$ sudo ./target/debug/bpfctl unload -i ens3 f7fe8625-5453-4eca-bb0b-c9130f15d6ea
$ sudo ./target/debug/bpfctl unload -i ens3 909edd07-f023-4dd1-9fee-16b0975bffd6
packet_write_wait: Connection to 192.168.122.118 port 22: Broken pipe
```

Billy99:
Further testing on latest changes still fails. Switched to a different interface (not main interface so I can still access the system). When the last program is removed, the dispatcher is still loaded. Added a log to `bpfd` to show when a program was removed (similar to the add log, which I tweaked the text on):
```
$ sudo ./target/debug/bpfd
12:07:37 [INFO] bpfd::server: [bpfd/src/server/mod.rs:50] Listening on [::1]:50051
12:07:45 [INFO] bpfd::server::bpf: [bpfd/src/server/bpf.rs:132] Program added: 1 programs attached to veth12fa8e3
12:08:01 [INFO] bpfd::server::bpf: [bpfd/src/server/bpf.rs:156] Program removed: 0 programs attached to veth12fa8e3
```
[...]

PR #61, opened 2022-07-07
`unload` the last program, dispatcher not unloaded

Billy99:
When the last program is removed, the dispatcher is still loaded. In the
dispatcher, `num_progs_enabled` is still one, but without a `ctx`, so `prog0()`
is returning `XDP_DISPATCHER_RETVAL` causing all packets on that interface to
drop.
    
Resolves: #52

### DIFF
--- a/bpfd/src/server/bpf.rs
+++ b/bpfd/src/server/bpf.rs
@@ -94,7 +94,7 @@ impl<'a> BpfManager<'a> {
             .ok_or_else(|| BpfdError::SectionNameNotValid(section_name.clone()))?;
         let id = Uuid::new_v4();
         let next_available_id = if let Some(prog) = self.programs.get(&iface) {
-            prog.len()
+            prog.len() + 1
         } else {
             self.programs.insert(iface.clone(), HashMap::new());
             1
--- a/bpfd/src/server/bpf.rs
+++ b/bpfd/src/server/bpf.rs
@@ -126,7 +126,7 @@ impl<'a> BpfManager<'a> {
         let _old_links = self.attach_extensions(&iface, &mut dispatcher_loader)?;
         self.update_or_replace_dispatcher(iface.clone(), dispatcher_loader)?;
         info!(
-            "{} programs attached to {}",
+            "Program added: {} programs attached to {}",
             self.programs.get(&iface).unwrap().len(),
             &iface,
         );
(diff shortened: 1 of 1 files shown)


bpfd/src/server/bpf.rs
```diff
@@ -149,7 +149,15 @@ impl<'a> BpfManager<'a> {
...
             // Keep old_program until the dispatcher has been reloaded
             let _old_program = programs.remove(&id).unwrap();
+            info!(
+                "Program removed: {} programs attached to {}",
+                programs.len(),
+                &iface,
+            );
             if programs.is_empty() {
+                // Keep old_program until the dispatcher has been reloaded
+                let _old_programs = self.programs.remove(&iface).unwrap();
```
dave-tucker, line 159: The comment is incorrect - copy/pasted?
We don't need to keep `_old_programs` in scope as it's just an empty hashmap.
`self.programs.remove(&iface);` will work just fine.

bpfd/src/server/bpf.rs
```diff
@@ -333,6 +341,11 @@ impl<'a> BpfManager<'a> {
         }
         Ok(())
     }
+
+    fn remove_dispatcher(&mut self, iface: String) -> Result<(), BpfdError> {
```
dave-tucker, line 345: Since this is one line we may as well just inline it to save adding another function.

bpfd/src/server/bpf.rs
```diff
@@ -94,7 +94,7 @@ impl<'a> BpfManager<'a> {
             .ok_or_else(|| BpfdError::SectionNameNotValid(section_name.clone()))?;
         let id = Uuid::new_v4();
         let next_available_id = if let Some(prog) = self.programs.get(&iface) {
-            prog.len()
+            prog.len() + 1
```
dave-tucker, line 97: Good catch!

dave-tucker (changes requested):

dave-tucker:
> @dave-tucker This is probably wrong based on our discussion, but I can't figure out what is suppose to be freed when it goes out of context. When the last element of a HashMap is removed, as far as I can tell, it "Keeps the allocated memory for reuse." And not sure why the dispatcher would automatically be unloaded when the last extension is removed, what would trigger that in rust?

No you're right. Yesterday I was sure that I had called `self.dispatchers.remove(&iface)` when the last program was removed, which as you've noted here would do the right thing.

bpfd/src/server/bpf.rs
Billy99: Using `#[inline(always)]` or meaning just get rid of the function and move the `self.dispatchers.remove(&iface);` call above?

bpfd/src/server/bpf.rs
Billy99: Done

bpfd/src/server/bpf.rs
Billy99: Done using `#[inline(always)]`.

PR #62, opened 2022-07-07
Prevent panic when invalid interface provided

Billy99:
In any of the `bpfctl` commands, if an invalid interface is provide,
`bpfd` panics. Add a check to make sure the interface exists before
proceeding.

Resolves: #63

### DIFF
--- a/bpfd/src/server/bpf.rs
+++ b/bpfd/src/server/bpf.rs
@@ -9,8 +9,11 @@ use aya::{
 };
 use bpfd_common::*;
 use log::info;
-use nix::sys::socket::{
-    sendmsg, socket, AddressFamily, ControlMessage, MsgFlags, SockFlag, SockType, UnixAddr,
+use nix::{
+    net::if_::if_nametoindex,
+    sys::socket::{
+        sendmsg, socket, AddressFamily, ControlMessage, MsgFlags, SockFlag, SockType, UnixAddr,
+    },
 };
 use uuid::Uuid;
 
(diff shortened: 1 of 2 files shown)


Billy99:
@dave-tucker If there is a better way to verify, or a better package to use (this was all I could find), just let me know and I'll rework.

dave-tucker:
@Billy99 you can use https://docs.rs/nix/latest/nix/net/if_/fn.if_nametoindex.html instead, which won't require another dependency.
e.g
```rust
use nix::net::if_::if_nametoindex;

fn main() -> anyhow::Result<()> {
    let index = if_nametoindex("wlp2s0")?;
    println!("ifindex of wlp2s0 is {}", index);
    match if_nametoindex("foo") {
        Ok(index) => panic!("foo should not have ifindex {}", index),
        Err(_) => Ok(()),
    }
}
```

Billy99:
@dave-tucker Just an FYI - I used `if_nameindex()` and looped through the list instead of `if_nametoindex()` as described above. I found on my system (by accident) that my main interface has an `altname`.
```
$ ip a
:
2: ens3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 52:54:00:42:48:36 brd ff:ff:ff:ff:ff:ff
    altname enp0s3
    inet 192.168.122.118/24 brd 192.168.122.255 scope global dynamic noprefixroute ens3
       valid_lft 2686sec preferred_lft 2686sec
    inet6 fe80::5054:ff:fe42:4836/64 scope link noprefixroute 
       valid_lft forever preferred_lft forever
```
Both `ens3` and `enp0s3` pass the `if_nametoindex()` test. But if I insert a program on `ens3` and then insert one on `enp0s3`, then the `bpfd` panics. Since the index to the map is the interface name, they are treated as different interfaces, so on the second call, there is a panic when the interface is attempted to be taken. The other workaround would be to switch the index to map to be the ifindex, but not sure that is worth it just to allow the altname to be used,

dave-tucker:
I think its totally worth switching to use ifindex as the map key as its guarantees to be unique and therefore less error prone... Maybe we should make an issue for it.

Issue #22, opened 2022-06-01
Create a test deployment


dave-tucker:
A simple script to standup a VM, install Fedora 36, install the RPMs from #5 and provide SSH access for testing.
For bonus points we could do with a kernel that's built on net-next with unprivileged BPF enabled.

astoycos:
Closed by #65

PR #65, opened 2022-07-12
Added automated VM build

nerdalert:
- the Vagrantfile deploys a rawhide VM with virtualbox
  for cross-platform support
- the Vagrantfile file calls runs provision.yaml ansible file
- the systemd unit file gets copied to the VM and deployed

### DIFF
--- a/packaging/vm-deployment/Vagrantfile
+++ b/packaging/vm-deployment/Vagrantfile
@@ -1,0 +1,15 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure("2") do |config|
+  config.vm.box = "generic/fedora36"
+  config.vm.provider "virtualbox" do |v|
+    v.memory = 4094
+    v.cpus = 4
+  end
+
+  config.vm.provision :ansible do |ansible|
+    ansible.limit = "all"
+    ansible.playbook = "provision.yaml"
+  end
+end
--- a/packaging/vm-deployment/bpfd.service
+++ b/packaging/vm-deployment/bpfd.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=bpfd - loading daemon
+After=network.target
+
+[Service]
+ExecStart=/usr/sbin/bpfd
+MemoryAccounting=true
+MemoryLow=infinity
+MemoryMax=infinity
+User=bpfd
+Group=bpfd
+AmbientCapabilities=CAP_BPF CAP_NET_ADMIN CAP_SYS_RESOURCE
+CapabilityBoundingSet=CAP_BPF CAP_NET_ADMIN CAP_SYS_RESOURCE
+
+[Install]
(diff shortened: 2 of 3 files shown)


nerdalert:
- This is a stab at #22. Will change out to an RPM once the crate is up (which may already be the case)
- Will do a readme in another PR once this is reviewed and what not. Took some liberties with the file locations that @dave-tucker may want elsewhere. Ty!

To run do the following:

```
dnf -y install vagrant
python3 -m pip install --user ansible
cd packaging/vm-deployment
vagrant up
```

packaging/vm-deployment/Vagrantfile
```diff
@@ -0,0 +1,15 @@
+# -*- mode: ruby -*-
```
dave-tucker, line 1: Did you try using [this module[(https://docs.ansible.com/ansible/latest/collections/community/libvirt/virt_module.html)
Example: https://www.redhat.com/sysadmin/build-VM-fast-ansible

That would mean we don't need Vagrant :tada: and users on macOS would be OK as I *hope* libvirt supports using the qemu hvf backend - see: [https://libvirt.org/macos.html].

packaging/vm-deployment/bpfd.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=bpfd - loading daemon
+After=network.target
+
+[Service]
+ExecStart=/usr/sbin/bpfd
+MemoryAccounting=true
+MemoryLow=infinity
+MemoryMax=infinity
+User=vagrant
```
dave-tucker, line 10: Should be `bpfd` user and group here. Need to have ansible create them if they don't exist.

packaging/vm-deployment/provision.yaml
```diff
@@ -0,0 +1,90 @@
...
+  - name: Enable IPv6
+    command: sysctl -w net.ipv6.conf.all.disable_ipv6=0
+
+  - name: check if cargo is installed
+    become: no
+    shell: command -v cargo
+    register: cargo_exists
+    ignore_errors: yes
+
+  - name: Download Rust Installer
```
dave-tucker, line 26: It would probably be better to build the binaries outside of the test machine IMHO.

packaging/vm-deployment/bpfd.service
nerdalert: Done!

packaging/vm-deployment/Vagrantfile
nerdalert: I've seen that post before but never tried it. It would require more files (roles, galaxy dir, ansible.cfg, libvirt template etc.) and require a bit more maintenance (OS versioning that vagrant does for you etc). Good either way! Might be worth doing it on another PR since it will take me a few to test it out across various platforms. Ty!

dave-tucker (approved):
This looks great, thanks @nerdalert

PR #66, opened 2022-07-12
Fix CI build panic

nerdalert:
- Added an explicit rustup default nightly to resolve build issue


nerdalert:
Well hrmph, that didnt fix the build but worked locally for me. Will investigate more tomorrow!

astoycos:
I would think this would fix it... :/

Issue #19, opened 2022-06-01
Extend API to allow passing of chain call actions


dave-tucker:
See: https://github.com/xdp-project/xdp-tools/blob/master/lib/libxdp/protocol.org#populating-the-dispatcher-configuration-map

From the command line, we should have something like  `--proceed-on pass --proceed-on redirect`.. where that argument can be repeated multiple times...
The default would be to assume `--proceed-on pass`. Should be a `Vec<XdpAction>`, where `XdpAction` is an `enum` that has a `TryFrom<String>` implementation - see `ProgramType` for reference.

The `bpfd.proto` file will need a new argument added to the `Load` RPC. I would suggest a `repeated` arg of type `String`.

On the server side, we will need some code to convert the `Vec<String>` into a bitmap (see: libxdp protocol linked above).
The bitmap would be cached in `ExtensionProgram` for re-use...
And it needs to be populated into the `XdpDispatcherConfig`

PR #69, opened 2022-07-15
Extend API to allow passing of chain call actions

Billy99:
This PR extends the API to allow passing of chain call actions. The
following changes were made:
* Update `bpfctl load` by adding a new repeating parameter `proceed-on`.
* Update `bpfctl list` to output the new parameter.
* Update the RPC API for Load and List to include a list of `proceed-on` values.
* Update `bpfd` to process the new parameter. The list is converted to a
bitmask, stored in the metadata in `bpfd`, and passed to the dispatcher in
the existing `chain_call_actions` field of the `config` map, which is shared
between the dispatcher and `bpfd`.
* Update the tutorial.md to show example of new field.

Resolves: #19


bpfd/src/server/bpf.rs
```diff
@@ -17,19 +17,24 @@ use nix::{
...
+use crate::{
+    proto::bpfd_api::ProceedOn,
+    server::{
+        config::{Config, XdpMode},
+        errors::BpfdError,
+    },
 };
 
-const DEFAULT_ACTIONS_MAP: u32 = 1 << 2;
+const DEFAULT_ACTIONS_MAP: u32 =
```
dave-tucker, line 28: When I refactored this last I was really careful to try and create some meaningful boundaries between the different modules in this crate.

It would be great if there were a way not to require `proto::bpfd_api::ProceedOn` here.
Maybe we could add the defaults in the `rpc` module?

bpfd/src/server/bpf.rs
```diff
@@ -195,12 +222,32 @@ impl<'a> BpfManager<'a> {
             .collect::<Vec<_>>();
         extensions.sort_by(|(_, a), (_, b)| a.current_position.cmp(&b.current_position));
         for (id, v) in extensions.iter() {
+            let mut proceed_on = Vec::new();
```
dave-tucker, line 225: I think we should store this as `proceed_on: Vec<i32>` in the `struct ExetensionProgram`  (or `Metadata` if we want to include this in the sort).

You can then do:

```rust
impl ExtensionProgram {
    fn proceed_on_mask(&self) -> u32 {
      ...
    }
}
```
Instead of having this block of code here.

bpfd/src/server/bpf.rs
```diff
@@ -117,13 +134,18 @@ impl<'a> BpfManager<'a> {
                 current_position: None,
                 metadata: Metadata {
                     priority,
+                    proceed_on_mask,
```
dave-tucker, line 137: Including this in the `Metadata` will affect the sort.
When priority, name, and attached are equal, the `proceed_on_mask` will be used as a tie-breaker.
This might not be intended.

bpfd/src/server/bpf.rs
```diff
@@ -265,7 +312,7 @@ impl<'a> BpfManager<'a> {
             .unwrap()
             .values_mut()
             .collect::<Vec<&mut ExtensionProgram>>();
-        extensions.sort_by(|a, b| a.metadata.cmp(&b.metadata));
+        extensions.sort_by(|a, b| a.current_position.cmp(&b.current_position));
```
dave-tucker, line 315: I don't think we should sort by current position, as new programs would always go to the end, which isn't what we want if priority is lower.

bpfd/src/server/bpf.rs
```diff
@@ -165,8 +187,13 @@ impl<'a> BpfManager<'a> {
                 return Ok(());
             }
 
+            // Cache program length so programs goes out of scope and
+            // sort_extensions() can generate its own list.
+            let program_len = programs.len() as u8;
+            self.sort_extensions(&if_index);
```
dave-tucker, line 193: Why do we sort here? It seems like it's less efficient - we sort right before attach at the moment so we only iterate over the list once...
In this code we iter at least once more to set `current_position` :thinking:

dave-tucker (changes requested):
Looking great, thanks @Billy99. Left a few comments.

Billy99 (commented):
The commit to follow will address all the comments.

bpfd/src/server/bpf.rs
Billy99: Still need to work on this one. Just hard-coded it to 2 and 31 for now.

bpfctl/src/main.rs
```diff
@@ -88,15 +103,38 @@ async fn main() -> anyhow::Result<()> {
             iface,
             section_name,
             priority,
+            proceed_on,
         } => {
             let path_str: String = path.to_string_lossy().to_string();
-            let prog_type = ProgramType::try_from(program_type.to_string()).unwrap();
+
+            let prog_type = match ProgramType::try_from(program_type.to_string()) {
```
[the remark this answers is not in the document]
Billy99: Done

Billy99:
@dave-tucker Didn't know if you liked code review fixes in a separate commit or squashed. I added as a separate commit and can squash at anytime.

dave-tucker (approved):
Thanks @Billy99, this is awesome!

Issue #74, opened 2022-07-22
bpfd won't run if "/etc/bpfd/programs.d" doesn't exist


Billy99:
With the addition of adding the ability to load programs at daemon startup, bpfd won't run if "/etc/bpfd/programs.d" doesn't exist:

```
$ sudo ./target/debug/bpfd
Error: No such file or directory (os error 2)
```

PR #75, opened 2022-07-22
bpfd won't run if "/etc/bpfd/programs.d" doesn't exist

Billy99:
Add code to handle the fact that a file or directory may not be there.

Resolved: #74


Billy99:
@astoycos I'm out of town all week. If this isn't the right fix, just abandon this PR and find a better way. I just think bpfd should run if "/etc/bpfd/programs.d" doesn't exist.

dave-tucker (approved):
LGTM

PR #78, opened 2022-07-25
Ensure BPFD can start if parsing programs fails

### LINKED ISSUE #74: bpfd won't run if "/etc/bpfd/programs.d" doesn't exist

astoycos:
Ensure we always start bpfd reguardless of wether
or not parsing the files in the static program
directory fails.

Fixes #74 

Closes #75

### DIFF
--- a/bpfd/src/main.rs
+++ b/bpfd/src/main.rs
@@ -3,6 +3,7 @@
 
 use aya::include_bytes_aligned;
 use bpfd::server::{config_from_file, programs_from_directory, serve};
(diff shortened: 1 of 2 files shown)


astoycos:
Closing in Favor of #75

Issue #88, opened 2022-08-04
Switch to using env_logger


dave-tucker:
Aya is making this change also

Issue #89, opened 2022-08-04
Logging to systemd journal


dave-tucker:
According to [this article](https://www.freedesktop.org/software/systemd/man/systemd-journald.service.html#Journal%20Namespaces) stdout/stderr is automatically captured.
We should use the correct format for logs to get correctly picked at the right levels in the journal...
So either syslog, if `/dev/log` is supported/available or perhaps using a log file.

UPDATE: looks like https://docs.rs/systemd-journal-logger/0.5.0/systemd_journal_logger/ could solve this for us.
And using a snippet like [this](https://codeberg.org/flausch/systemd-journal-logger.rs/src/branch/main/examples/systemd_service.rs) we can fall back to using `env_logger` if not connected to the journal

PR #93, opened 2022-08-05
Switch to using env_logger

Billy99:
Following Aya, switch bpfd to use env_logger in place of simplelog. By
default, logs are printed at `error` level instead `info`. This can be
overwritten using the `RUST_LOG` environment variable.
`sudo RUST_LOG=info ./target/debug/bpfd`

Resolves: #88

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,15 @@
 version = 3
 
 [[package]]
+name = "aho-corasick"
+version = "0.7.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


Billy99:
As mentioned above, this changes the default log level from `info` to `error`. We can back to `info` by changing the `init()` to something like:
```
env_logger::Builder::new().filter_level(log::LevelFilter::Info).init();
```
But since the default seems to be to run `bpfd` in the background, may be best to limit the number of logs dumped. There are other issues around logging to systemd journal (#89) that may make it more useful to go back to a default of `info` in the future.

dave-tucker:
Thanks @Billy99!

PR #100, opened 2022-08-08
Logging to systemd journal

Billy99:
If bpfd is running as a service, log to journald.

Resolves: #89


dave-tucker (approved):
LGTM

PR #103, opened 2022-08-10
Make protobuf compiler extensions installation instructions more clear

anfredette:
The commands are explicitly provided for installing the needed dependencies for all cases except for the go protobuf compiler extensions.  It is easy to miss the fact that in this case, you have to go to the link for the instructions instead of just using the commands provided.  This is a small edit to make that more clear.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,8 @@ cargo install bpf-linker
 sudo dnf install protobuf-compiler
 ```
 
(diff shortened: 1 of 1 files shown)


Billy99 (approved):
/lgtm

dave-tucker (approved):
LGTM

PR #110, opened 2022-08-19
Enable running bpfd as systemd service

Billy99:
As part of this PR, eased the path to allow bpfd to run as a systemd service.

The following changes were made:
* Added a new `setup.sh` script to manage all the installation.
* Added a new `user.sh` script to handle creating a bpfd and bpfctl user
  and creating the `/etc/bpfd/` and a new a `/etc/bpfctl/` directory. Also
  create the associated sub-folders and assign user and user group to
  the files and directories. Also copy over a default `bpfd.toml` and
  `bpfctl.toml` file.
* Added a new `install.sh` script to copy the binaries to `/usr/sbin/.`
  and assign user and user group to the files. Also copy a default
  `bpfd.service` file to `/usr/lib/systemd/system/`.
* Modified the `certificates.sh` script to be called by `setup.sh`.
* The `bpfd.service` limits the set of capabilities bpfd service has.

Changes in behavior:
* Default `bpfd.toml` and `bpfctl.toml` files are copied over.
* Moved `/etc/bpfd.toml` to `/etc/bpfd/bpfd.toml`.
* Mover `/etc/bpfd/bpfctl.toml` to `/etc/bpfct/bpfctl.toml`.
* `certificates.sh` is no longer called directly. All certs creation
  goes through `setup.sh`.

See `tutorial.md` and `./scripts/setup.sh help` for full description
of how everything now works.

Signed-off-by: Billy McFall <22157057+Billy99@users.noreply.github.com>

Issue #117, opened 2022-08-30
Fix bpfctl configuration


dave-tucker:
Current layout in `/etc`:

```
/etc/bpfctl/
├── bpfctl.toml
└── certs
    ├── bpfctl
    │   ├── bpfctl.key
    │   └── bpfctl.pem
    └── gocounter
        ├── gocounter.key
        └── gocounter.pem
```

Notes:

1. `bpfctl` should support local configuration with a higher precedence e.g in `~/.config/bpfctl` or somesuch.

2. We use the `SubjectCN` of certificates as the `username` as far as `bpfd` is concerned. This will be used to correctly set ownership of files in `/var/run/bpfd/bpffs` with `${username:bpfd}` and to set the ACL to be something like `0664` to allow r/w to owner and bpfd group and r/o to everyone else (see #18) as well as to determine API access. We have 2 choices:
  a.  Implement (1) and deprecate using `/etc` for bpfctl
  b.  Decouple the transport authentication (TLS certs) from the user authentication somehow. i.e have our own username/apitoken database that we use for authentication. Then you can bootstrap a `/etc/bpfd/certs/client/client.pem` certificate that any user in the right group can use... but you must also provide an API token.

astoycos:
fixed by #118

PR #118, opened 2022-08-30
Change user `bpfctl` to `bpfd-clients`

Billy99:
This is a continuation from #110. Cleanup the initialization scripts
by using variables for USER, USER_GROUP and BINARY. Move
`/etc/bpfd/bytecode/` to `/var/bpfd/bytecode/`.

### DIFF
--- a/docs/admin/configuration.md
+++ b/docs/admin/configuration.md
@@ -50,7 +50,7 @@ proceed_on = ["pass", "dispatcher_return"]
 
 ## bpfctl
 
-bpfctl expects a configuration file to be present at `/etc/bpfctl/bpfctl.toml`.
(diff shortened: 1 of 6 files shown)


Billy99:
I can change the name if something is preferred over `bpfd-clients`.

dave-tucker (changes requested):
I don't think we should do this. See: #117 for context.
We can discuss more on the call today.

Billy99:
> I don't think we should do this. See: #117 for context.

OK, I updated the PR. It no longer changes the name of the user. Now just tweaks the scripts to hopefully be more clear as to how the strings `bpfd` and `bpfctl` are used (are they the User, the User Group or the Binary).

dave-tucker (approved):
LGTM. Thanks @Billy99

PR #139, opened 2022-09-12
Updated the installer to start the bpfd service

nerdalert:


nerdalert:
cc/ @Billy99 I think it makes sense to start the pfd service by default. I could add an env to explicitly not run in case of override `START_BPFD=${START_BPFD:-true}` and do a double conditional check on `(reinstall=false && start_bpfd=false)` but that might be overthinking things. Either way by default I would advocate starting the service. Ty!

Billy99:
Need to update documentation to match.
- In setup.sh (https://github.com/redhat-et/bpfd/blob/main/scripts/setup.sh#L72) indicate that service is started automatically and can managed using stop and start.
- In tutorial.md (https://github.com/redhat-et/bpfd/blob/main/docs/admin/tutorial.md?plain=1#L263) indicate that service is started automatically and can managed using stop and start.

nerdalert:
> Need to update documentation to match.

@Billy99 done, ty for taking a look!

astoycos:
/lgtm 

Thanks!

Issue #140, opened 2022-09-12
Systemd svc start error: "Error: unable to create mountpoint"


nerdalert:
Running into this error on latest main. I tested with an older binary from 9/1 (commit id: `41c6aa17c56decdae314fe9ca69dd88b0f13ec0a`) and it installs ok so its something that went in probably last week:

Replicate the issue with:
1. from main branch run `sudo ./scripts/setup.sh install`
2. attempt to start the svc `sudo systemctl start bpfd.service`

```
$ uname -r
5.18.19-200.fc36.x86_64

[vagrant@fedora36 bpfd]$ sudo ./scripts/setup.sh install
Creating users/groups:
  Creating user:group "bpfd:bpfd"  group_param=""
  Ensure "/etc/bpfd" exists
  chown and chmod of "/etc/bpfd"
  Copying "bpfd.toml"
  Creating "bpfd" specific directories
  setfacl of "/etc/bpfd/sock/"
  chown and chmod of "/var/bpfd"
  Creating user:group "bpfctl:bpfd"  group_param="-g bpfd"
  Ensure "/etc/bpfctl" exists
  chown and chmod of "/etc/bpfctl"
  Copying "bpfctl.toml"
Creating certs:
mode of '/etc/bpfd/certs/ca/ca.key' changed from 0600 (rw-------) to 0440 (r--r-----)
mode of '/etc/bpfd/certs/ca/ca.pem' changed from 0644 (rw-r--r--) to 0444 (r--r--r--)
Certificate request self-signature ok
subject=CN = bpfd
mode of '/etc/bpfd/certs/bpfd/bpfd.pem' changed from 0644 (rw-r--r--) to 0440 (r--r-----)
mode of '/etc/bpfd/certs/bpfd/bpfd.key' changed from 0600 (rw-------) to 0440 (r--r-----)
Certificate request self-signature ok
subject=CN = bpfctl
mode of '/etc/bpfctl/certs/bpfctl/bpfctl.pem' changed from 0644 (rw-r--r--) to 0440 (r--r-----)
mode of '/etc/bpfctl/certs/bpfctl/bpfctl.key' changed from 0600 (rw-------) to 0440 (r--r-----)
Copy binaries:
  Copying "bpfd" and chown "bpfd:bpfd"
  Copying "bpfctl" and chown "bpfctl:bpfd"
Copy service file:
  Copying "bpfd.service" and chown "bpfd:bpfd"
[...]

PR #145, opened 2022-09-15
Fix Systemd svc start error: "Error: unable to create mountpoint"

Billy99:
Resolves: #140


nerdalert:
@Billy99 that did it, thanks 🎉

nerdalert (commented):
@Billy99 that did it, thanks 🎉

Billy99:
Which docs go in `docs/admin/` and which go in `docs/developer/`?

astoycos:
@Billy99 Shouldn't this all be admin docs? The general point of the doc is how to configure bpfd correctly, which in a normal Cloud environment an "admin" would always be doing. 

Developer docs are probably more pubilc API related things IMO, but If there's other ideas I'm happy to hear them

Billy99:
> Shouldn't this all be admin docs?
I couldn't tell what went where. Not sure we even need two docs directories for bpfd, but happy to move. I'll move in a follow-up PR.

astoycos:
It's a good point... Really the only docs I would see Admin's needing would be those related to the APIs bpfd exposes i.e using bpfctl, and the specs to which we align with.  Otherwise I think this is GTG for now we can address where the docs are living (as a whole) in a follow up PR :)

PR #148, opened 2022-09-16
Stop some BPFD Panics

astoycos:
Ensure we don't panic If bpfd fails to process one of the 
commands, Instead report errors to users (both client 
and server side) and continue.

### DIFF
--- a/bpfd/src/server/rpc.rs
+++ b/bpfd/src/server/rpc.rs
@@ -6,7 +6,7 @@ use bpfd_api::v1::{
     list_response::ListResult, loader_server::Loader, ListRequest, ListResponse, LoadRequest,
     LoadResponse, UnloadRequest, UnloadResponse,
 };
-use log::info;
+use log::warn;
 use tokio::sync::{mpsc, mpsc::Sender, oneshot};
 use tonic::{Request, Response, Status};
 use uuid::Uuid;
@@ -102,12 +102,19 @@ impl Loader for BpfdLoader {
 
         // Await the response
         match resp_rx.await {
-            Ok(id) => {
-                reply.id = id.unwrap().to_string();
-                Ok(Response::new(reply))
-            }
+            Ok(res) => match res {
+                Ok(id) => {
+                    reply.id = id.to_string();
+                    Ok(Response::new(reply))
+                }
+                Err(e) => {
+                    warn!("BPFD load error: {}", e);
+                    Err(Status::aborted(format!("{}", e)))
+                }
+            },
+
             Err(e) => {
-                info!("RPC load error: {}", e);
+                warn!("RPC load error: {}", e);
                 Err(Status::aborted(format!("{}", e)))
             }
         }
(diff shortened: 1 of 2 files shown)


astoycos:
https://github.com/rust-lang/rust/issues/100219 Is blocking CI I think

nerdalert (commented):
/lgtm Huzzah! I will add this to IT. Looks good, here was the output.

```
$ sudo ./target/debug/bpfctl list --iface eth0
eth0
xdp_mode:
```

Ironically I hacked up a similar fix earlier this week talking to @dave-tucker as follows (note your rust is better than mine lol) but Im pasting it because I couldn't come up with a way to return a message like `no programs loaded on the specified interface` so I just hacked into xdp_mode into an emtpy vec 😅:

```rust
pub(crate) fn list_programs(&mut self, iface: String) -> Result<InterfaceInfo, BpfdError> {
        let if_index = self.get_ifindex(&iface)?;
        if !self.dispatchers.contains_key(&if_index) {
            let mut empty_results = InterfaceInfo {
                xdp_mode: "no programs loaded on the specified interface".parse().unwrap(),
                programs: vec![],
            };
            return Ok(empty_results)
            // return Err(BpfdError::NoProgramsLoaded);
        } else {
            let xdp_mode = self.dispatchers.get(&if_index).unwrap().mode.to_string();
            let mut results = InterfaceInfo {
                xdp_mode,
                programs: vec![],
            };
            let mut extensions = self
                .programs
                .get(&if_index)
                .unwrap()
                .iter()
                .collect::<Vec<_>>();
            extensions.sort_by(|(_, a), (_, b)| a.current_position.cmp(&b.current_position));
            for (id, v) in extensions.iter() {
                results.programs.push(ProgramInfo {
                    id: id.to_string(),
                    name: v.metadata.name.clone(),
                    path: v.path.clone(),
                    position: v.current_position.unwrap(),
                    priority: v.metadata.priority,
                    proceed_on: v.proceed_on.clone(),
                })
            }
            Ok(results)
        }
    }

// $ sudo ./target/debug/bpfctl list --iface eth0
// eth0
// xdp_mode: no programs loaded on the specified interface
```

Maybe there doesn't need to be a message and the empty vec is good enough. Part of me wants a table output like(ish):

```
UUID                                    PRIORITY        IFACE            PATH         
5b0ced60-2592-4f30-8cb7-efdf1f4cbfc5    50              eth0            /var/bpfd/bytecode/pass.bpf.o
39fbd797-5ee6-4bab-96cd-5fadff263641.   100              eth0            /var/bpfd/bytecode/pass.bpf.o
```

Not to hijack your thread, just wanted to throw it out there!

astoycos:
@nerdalert Now we show, is that pretty much what you meant? 

```
[astoycos@localhost bpfd]$ sudo RUST_LOG=info ./target/debug/bpfctl list --iface eno3
eno3
no programs loaded on the specified interface
```

nerdalert:
> no programs loaded on the specified interface

Only if folks think that makes sense to you all. I personally like a message!

bpfd/src/server/rpc.rs
```diff
@@ -171,23 +184,34 @@ impl Loader for BpfdLoader {
...
+                            proceed_on: r.proceed_on,
+                        })
+                    }
+                    Ok(Response::new(reply))
                 }
-                Ok(Response::new(reply))
-            }
+                Err(e) => match e {
+                    BpfdError::NoProgramsLoaded => {
+                        warn!("{}", e);
```
dave-tucker, line 204: I don't think we need to `warn!` on this case as we're ignoring the error and returning a success response

dave-tucker (dismissed):
One tiny nit but this looks great! Thanks for cleaning this up.

bpfd/src/server/rpc.rs
astoycos: Good point, Just fixed this, thanks!

Issue #144, opened 2022-09-14
Bug: Programs not retrievable with `bpfctl list` between bpfd restarts


nerdalert:
Steps to reproduce:
1. Start bpfd `sudo ./target/debug/bpfd`
2. Add a program `sudo ./target/debug/bpfctl load --iface eth0 --priority 100 --section-name xdp --program-type xdp ../xdp-tutorial/basic01-xdp-pass/xdp_pass_kern.o`3. 
3. Kill and restart the bpfd process
4. Run a list `sudo ./target/debug/bpfctl list --iface eth0`
5. Observe bpfd panic
6. Start bpfd again
7. Add another rule `sudo ./target/debug/bpfctl load --iface eth0 --priority 101 --section-name xdp --program-type xdp ../xdp-tutorial/basic01-xdp-pass/xdp_pass_kern.o`
8. Run a list `sudo ./target/debug/bpfctl list --iface eth0` (This works and prints both rules)

Workaround - Restarting bpfd and adding another rule will make `bpfctl list` function again

Once we resolve this, I will add it to IT.

```
### Load a program ###
$ sudo ./target/debug/bpfctl load --iface eth0 --priority 100 --section-name xdp --program-type xdp ../xdp-tutorial/basic01-xdp-pass/xdp_pass_kern.o
e918d0c6-6bc0-459f-9e15-61de2ba3e84c

### List the rules ###
$ sudo ./target/debug/bpfctl list --iface eth0
eth0
xdp_mode: skb

0: e918d0c6-6bc0-459f-9e15-61de2ba3e84c
	section-name: "xdp"
	priority: 100
	path: ../xdp-tutorial/basic01-xdp-pass/xdp_pass_kern.o
	proceed-on:

### Kill and restart bpfd ###
sudo ./target/debug/bpfd

### Run bpfctl list again ###
[fedora@ip-172-31-93-116 bpfd]$ sudo ./target/debug/bpfctl list --iface eth0
Error: status: Unknown, message: "transport error", details: [], metadata: MetadataMap { headers: {} }

Caused by:
    0: transport error
    1: connection error: unexpected end of file
    2: unexpected end of file

### bpfd panic ###
$ sudo ./target/debug/bpfd
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', bpfd/src/server/bpf.rs:369:46
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

PR #149, opened 2022-09-16
Programs not retrievable with bpfctl list between bpfd restarts

Billy99:
When rebuilding the state of the previously loaded programs, the `current_position` was not being set. This is the slot in the dispatcher and is only set when the list of programs are sorted.

Resolves: #144


astoycos:
/lgtm 

Just waiting for CI to sort itself out

Billy99:
I want @dave-tucker to take a look. `current_position` is definitely not being set, but not sure if there is a better way to calculate the value.

dave-tucker (approved):
LGTM

PR #154, opened 2022-09-20
Adds a workflow to validate bpfd installer/systemd

nerdalert:


nerdalert:
FYI: If this fails, it throw an exit 3 which will fail the workflow as desired.
```
    - name: Verify the bpfd systemd service is active
      run: systemctl is-active bpfd
```

dave-tucker (approved):

PR #168, opened 2022-09-24
[do not merge] Validate bpfctl can connect to the daemon in CI

nerdalert:


nerdalert:
This is an issue I'm trying to unravel that bpfd doesn't seem to be sending a cert when bpfctl tries to connect to the daemon. So far I'm only replicating this in the GH runner 🤔

PR #172, opened 2022-09-26
Validate bpfctl can connect to the daemon in CI

nerdalert:


nerdalert:
@dave-tucker this is g2g now. It tests both the installer/systemd and IT as well. Ty!

Issue #109, opened 2022-08-15
Explore what can become optional in the BPFD GRPC API 


astoycos:
From a bpfd standpoint we should:

1. Build config from BTF (skip that for now)
2. Apply overrides from image labels (if container)
3. Apply overrides from the API request

That probably means that a lot of what's in the GRPC api now can actually become optional... and the CLI design would follow.
So --path and --image  are the only required items (mutually exclusive)
Everything else would be `Option<...>`

This issue will track the work to explore how GRPC handles optional fields, and how we can optimize the API and bpfd to reduce "data over the wire" as much as possible.

Issue #38, opened 2022-06-23
Support other program types?


dave-tucker:
While the API is built specifically to handle network (TC/XDP) program types, there is an argument to be made that supporting other programs could be useful.

In which case we'd need to:

1. Rename our current `Load` API to `LoadXDP` and maybe `LoadTC` depending on #9 
2. Add new APIs for the other supported program types

Expanding the API to support more program types would likely require #37 to reduce the scope of authorized APIs for each user.

PR #114, opened 2022-08-23
bpfd tc poc

anfredette:
I'm working on a poc for support of tc-bpf. It doesn't work yet, but I'm submitting for discussion purposes.

Signed-off-by: Andre Fredette <afredette@redhat.com>

PR #174, opened 2022-09-27
Support for tracepoint programs

dave-tucker:
TODO: 
- [x] Update the List and Delete RPCs to support different program types
- [x] Rebuild state handling
- [x] Implement the tracepoint attachment
- [x] Fix internal datastructure to handle multple different program types
- [x] Config file update to support tracepoints

Fixes: #109 (supports single/multi-attach, and uses grpc `one_of`)
Updates: #114 (should make it a little easier to plumb in tc-ingress/tc-egress as a networkmultiattach type)
Fixes: #38 (adds another program type)


bpfd/src/server/command.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// Load a program
+    Load {
+        path: String,
+        section_name: String,
+        program_type: ProgramType,
+        attach_type: AttachType,
+        username: String,
+        responder: Responder<Result<Uuid, BpfdError>>,
+    },
+    Unload {
```
anfredette, line 25: I think we're going to want a ProgramType in both the Unload and List commands.

dave-tucker: Good point. I also need to adjust the protobuf definition for ListResponse.
But first I'm going to work through the internal changes required as a result of adding tracepoints, which should (hopefully) inform the design of the ListReponse RPC.
It's likely that some sort of AttachInfo will need to be modelled per-program type but I'm finding things out as I go

bpfd/src/bpf.rs
```diff
@@ -105,27 +55,11 @@ impl DispatcherProgram {
...
-    pub(crate) priority: i32,
-    pub(crate) proceed_on: Vec<i32>,
-}
-
 pub(crate) struct BpfManager<'a> {
     config: &'a Config,
     dispatcher_bytes: &'a [u8],
     dispatchers: HashMap<u32, DispatcherProgram>,
-    programs: HashMap<u32, HashMap<Uuid, ExtensionProgram>>,
+    programs: HashMap<Uuid, Program>,
```
anfredette, line 62: So, it looks like you just have one list of programs indexed by UUID.

bpfd/src/bpf.rs
```diff
@@ -207,148 +135,205 @@ impl<'a> BpfManager<'a> {
...
+                .map_pin_path(map_pin_path.clone())
+                .load_file(&data.path)?;
+
+            ext_loader.program_mut(&data.section_name).ok_or_else(|| {
+                let _ = fs::remove_dir_all(map_pin_path);
+                BpfdError::SectionNameNotValid(data.section_name.clone())
+            })?;
+
+            // Calculate the next_available_id
+            let next_available_id = self
```
anfredette, line 163: To make sure I'm understanding this.  It looks like you are calculating the number of programs of a given type on a given interface (in this case) by iterating through the entire list of programs and selecting the ones of the type and interface you're looking for.  Is that correct?

bpfd/src/bpf.rs
```diff
@@ -207,148 +135,205 @@ impl<'a> BpfManager<'a> {
...
-                priority,
-                name: section_name,
-                attached: false,
-            },
-            owner,
-            proceed_on,
-        };
-        prog.save(id)
-            .map_err(|_| BpfdError::Error("unable to persist program data".to_string()))?;
+            // Calculate the dispatcher revision
```
anfredette, line 179: A question I've had is why we need to keep the dispatcher revision.  I may be missing it, but I'm not seeing where you're using it for anything.

proto/bpfd.proto
```diff
@@ -28,11 +29,22 @@ message LoadRequest {
     bool from_image = 2;
     string section_name = 3;
     ProgramType program_type = 4;
+    oneof attach_type {
+        NetworkMultiAttach network_multi_attach = 5;
```
anfredette, line 33: Can you explain your thoughts behind the multi and single attach terminology?  It seems like we can have multiple programs attached in both cases.  It's just in the NetworkMultiAttach case (e.g., XDP or TC) we are using a dispatcher.

anfredette (commented):
I'll probably have more, but here are a few questions.  Thanks, Andre

bpfd/src/bpf.rs
dave-tucker: Yes that's correct.

bpfd/src/bpf.rs
dave-tucker: This has been refactored slightly in the latest patchset but the concept remains the same...
We walk the entire programs list twice - filtering only the programs we want, like XDP for if_index 3.
First pass is to sort the programs and update the `current_position`, second time is to create a copy which gets passed to `Dispatcher::new()`.
That can probably be collapsed into a single iteration but my Rust skills are failing me there.

bpfd/src/bpf.rs
dave-tucker: The revision is used to allow us to keep 2 sets of extension programs attached at the same time to facilitate atomic replacement
i.e these paths can co-exist:
`dispatcher_1_1/prog_1_link`
`dispatcher_1_2/prog_1_link`

And when we've atomically replaced the dispatcher we can delete `dispatcher_1_1`

proto/bpfd.proto
dave-tucker: I was trying to differentiate between:

1. BPF programs where the kernel takes care of ordering, and we don't care
2. BPF programs where we take care of the ordering

I'm not 100% sold on the name though.

dave-tucker:
There's one bug here when I rebuilt the state handling.... We don't populate the `current_postition` of an XDP program.
I'll probably address that after merge since I don't want to keep this around too long as it's gotten rather big 😓 

At this point it's ready for test/review.

I will push an image to quay.io and add an integration test for Tracepoint, and one for TC to ensure that these don't break in future.

anfredette:
Thanks for answering my questions.  I'll give this a look tomorrow.

astoycos:
I'll try to give this a look today as well :)

dave-tucker:
I found bugs 😆 

1. bpfctl needed a rewrite - will push that commit shortly
2. TC isn't loading properly - programs are immediately detached. Will debug it...

bpfd/src/multiprog/xdp.rs
```diff
@@ -0,0 +1,221 @@
...
+        Ok(prog)
+    }
+
+    pub(crate) fn delete(&self, full: bool) -> Result<(), BpfdError> {
+        let path = format!("{RTDIR_XDP_DISPATCHER}/{}-{}", self.if_index, self.revision);
+        fs::remove_file(path)
+            .map_err(|e| BpfdError::Error(format!("unable to cleanup state: {}", e)))?;
+
+        let path = format!(
+            "{RTDIR_FS}/xdp/dispatcher_{}_{}",
```
Billy99, line 209:
```suggestion
            "{RTDIR_FS_XDP}/dispatcher_{}_{}",
```

bpfd/src/multiprog/xdp.rs
```diff
@@ -0,0 +1,221 @@
...
+                    .unwrap();
+                let new_link: FdLink = ext.take_link(new_link_id)?.into();
+                let path = format!(
+                    "{RTDIR_FS_XDP}/dispatcher_{if_index}_{}/link_{k}",
+                    self.revision
+                );
+                new_link.pin(path).map_err(|_| BpfdError::UnableToPin)?;
+            } else {
+                let mut bpf = BpfLoader::new()
+                    .extension(&v.data.section_name)
```
Billy99, line 160:
```suggestion
                    .extension(&v.data.section_name)
                    .map_pin_path(map_pin_path)
```
Except need to pull from v.data

Billy99 (commented):
Still going through, but here are a few changes.

bpfd/src/bpf.rs
```diff
@@ -0,0 +1,358 @@
...
+            .len();
+
+        let did = program
+            .dispatcher_id()
+            .ok_or(BpfdError::DispatcherNotRequired)?;
+
+        let old_dispatcher = self.dispatchers.remove(&did);
+
+        if let Some(ref old) = old_dispatcher {
+            if next_available_id == 0 {
```
anfredette, line 239: When the last tc program is deleted, the tc dispatcher doesn't get detached.  So, when we hit this condition, for a tc dispatcher, we need to explicitly call detach().  However, I haven't been able to figure out how to access the link so I can call link.detach().

bpfd/src/multiprog/tc.rs
```diff
@@ -0,0 +1,243 @@
...
+            .map_err(|e| BpfdError::Error(format!("unable to cleanup state: {}", e)))?;
+
+        let base = match self.direction {
+            Direction::Ingress => RTDIR_FS_TC_INGRESS,
+            Direction::Egress => RTDIR_FS_TC_EGRESS,
+        };
+        let path = format!("{base}/dispatcher_{}_{}", self.if_index, self.revision);
+        fs::remove_dir_all(path)
+            .map_err(|e| BpfdError::Error(format!("unable to cleanup state: {}", e)))?;
+        // Dispatcher will be detached when this object is dropped
```
anfredette, line 240: It's actually not dropped presumably because it's not "detach on drop" in Aya.

anfredette (changes requested):

examples/gocounter/main.go
```diff
@@ -9,7 +9,6 @@ import (
 	"crypto/x509"
 	"flag"
 	"fmt"
-	"io/ioutil"
```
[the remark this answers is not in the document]
dave-tucker: ReadFile was moved to "os" in Go 1.18 (IIRC) and it fails to compile on later go versions. I removed the import but forgot to update to `os.ReadFile` will push a fix.

bpfd/src/bpf.rs
dave-tucker: I've fixed this now - fix fixed in `TcDispatcher::delete` where we
```rust
if let Some(link) = self.link.take() {
    link.detach()?;
}
```

bpfd/src/multiprog/tc.rs
dave-tucker: Ack. Updated the comment and added a FIXME

bpfd/src/config.rs
```diff
@@ -0,0 +1,215 @@
+// SPDX-License-Identifier: (MIT OR Apache-2.0)
```
astoycos, line 1: Did this change? Not sure why it's not just coming up as a rename? Or maybe a rename and then some changes

anfredette (approved):
LGTM

Issue #157, opened 2022-09-21
`proceed_on` is always empty


dave-tucker:
Per title, using  `sudo ./target/debug/bpfctl load --iface eth0 --priority 100 --from-image quay.io/dave_tucker/xdp_pass` to install a program results in bpfctl showing the `proceed_on: ` to be empty :exploding_head:

PR #186, opened 2022-10-05
proceed_on is always empty

Billy99:
If proceed-on is not set in the gRPC message, it will not be displayed on the list (data is empty in the gRPC request). It is applied properly because when writing to the dispatcher, it is converted to a bitmask and a default mask is used for an empty vector.

Resolves: #157


bpfd/src/server/bpf.rs
```diff
@@ -22,7 +22,10 @@ use crate::server::{
 };
 
 // Default is Pass and DispatcherReturn
-const DEFAULT_ACTIONS_MAP: u32 = 1 << 2 | 1 << 31;
+const DEFAULT_XDP_PROCEED_ON_PASS: i32 = 2;
+const DEFAULT_XDP_PROCEED_ON_DISPATCHER_RETURN: i32 = 31;
```
astoycos, line 26: why do we need it to be an int here if we're just casting to u32 again on the next line?

astoycos: Oh I guess you did this to reuse it when adding to default below

bpfd/src/server/bpf.rs
```diff
@@ -253,6 +256,12 @@ impl<'a> BpfManager<'a> {
             (None, 0)
         };
 
+        let mut lcl_proceed_on = proceed_on;
```
astoycos, line 259: You _should_ just be able to do something like this to remove the lcl_proceed_on intermediate
```
let proceed_on = if proceed_on.is_empty() { 
    vec![DEFAULT_XDP_PROCEED_ON_PASS, DEFAULT_XDP_PROCEED_ON_DISPATCHER_RETURN]
} else {
    proceed_on
};
```

See playground example -> https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6694380c8c83082c909c2dd4998966bc

bpfd/src/server/bpf.rs
```diff
@@ -262,7 +274,7 @@ impl<'a> BpfManager<'a> {
                 attached: false,
             },
             owner,
-            proceed_on,
+            proceed_on: proceed_on,
```
astoycos, line 277:
```suggestion
            proceed_on,
```

astoycos:
/lgtm 

thanks!

Issue #116, opened 2022-08-26
Implement POC using dispatcher approach for adding tc bpf programs


anfredette:
The intent is to have the same user experience for managing multiple programs regardless of the program type.  In particular, programs should be prioritized in the same way and we should support the atomic addition and deletion of programs.

The goal for this issue is to prove out the following design.

Instead of adding tc programs directly to the tc subsystem, the intent is to use a dispatcher in the same way as is currently done for xdp programs.

Atomic replacement will be accomplished as follows:

1. If none of the added tc programs returns a terminal value, the dispatcher will return a `TC_ACT_OK` so that no lower priority tc programs get executed.
2. When a program is added or deleted, we will create a new dispatcher program with the new list of programs as eBPF extension programs.
3. The new dispatcher program will first be added at a higher priority than the current dispatcher program first, and then the old dispatcher will be removed.
4. A set of tc priority slots will be reserved for bpfd programs.  We will need to handle the wrap case in which the current dispatcher is at the highest priority of the reserved priorities.  We intend to handle this case by moving the current dispatcher to the lowest priority before the new dispatcher is added.

astoycos:
@anfredette I'd say you finished this :) 

Feel free to re-open

anfredette:
Yup.  This is done.  The full implementation is pretty much done - except for a few optimizations that need to be completed after my latest Aya patches are available.

PR #189, opened 2022-10-10
Support tc programs Using the dispatcher approach

anfredette:
- Both TC Ingress and Egress are supported.
- The list output is currently hacky because the message format currently only supports XDP programs.  This will be fixed with pr #174. In the meantime, this pr adds the TC programs to the end of the XDP list so they can still be displayed.
- Persistence is not supported.  TC programs will remain loaded and attached if bpfd is terminated without explicitly unloading the TC programs.  If this happens, the TC programs can be cleaned up manually by deleting the associated files in the /var/run/bpfd directory. For example, `sudo rm -rf /var/run/bpfd 2>/dev/null` will clean up all programs loaded by bpfd.

### DIFF
--- a/bpfd-common/src/lib.rs
+++ b/bpfd-common/src/lib.rs
@@ -3,6 +3,7 @@
 
 #![no_std]
 
+// XDP Defines
 pub const XDP_METADATA_SECTION: &str = "xdp_metadata";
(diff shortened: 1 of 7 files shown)


anfredette:
Implements https://github.com/redhat-et/bpfd/issues/116.

Issue #166, opened 2022-09-23
Move bpfd directory/file management out of setup.sh and into code


Billy99:
bpfd cannot guarantee the that the `setup.sh` script is run, so move all the directory creation and chown/chmod to bpfd user and user group out of the `setup.sh` script and into the code.

dave-tucker:
Just a thought, maybe we could also move certificate generation here too (in the case we have a default cert path and the certs don't exist)?
e.g https://github.com/djc/sign-cert-remote/blob/main/src/main.rs

dave-tucker:
See also: https://www.freedesktop.org/software/systemd/man/systemd.exec.html

So ultimately systemd can handle RuntimeDir, StateDir, CacheDir etc.. for us, making sure they are created and owned by the users specified in the unit file. This is awesome for working with systemd, but obviously doesn't help with the non-systemd usecase - i.e local development

Billy99:
@dave-tucker I have most of this work completed for bpfd, directories managed by systemd and creating certs within bpfd if it doesn't exist (not pushed yet). How do we want to handle `bpfctl`? Add a new command, `bpfctl certs` to create the certificate? What about creating `/etc/bpfctl/` and sub-directories as `bpfctl:bpfd` since there is no associated systemd service?

Billy99:
FYI: Added a `bpfctl certs` command that will create a certificate and key. Added a `bpfctl.service` files that is a oneshot service that calls `bpfctl certs` and creates the directories as needed. Still working on some details, like how to kick off automatically from `bpfd.service`.

Billy99:
As an FYI: I missed the link about certs generation above. I pulled the logic from https://github.com/sfackler/rust-openssl/blob/master/openssl/examples/mk_certs.rs.

Billy99:
PR #195 was a follow-up to #191 and completed this issue.

PR #191, opened 2022-10-12
Move bpfd directory/file management out of setup.sh and into code

Billy99:
This implements #166.

**First commit:**
From: https://www.freedesktop.org/software/systemd/man/systemd.exec.html systemd manages the following directories:
* ConfigurationDirectory: `/etc/`
* RuntimeDirectory: `/run/`  Note: RuntimeDirectory is deleted when service stops, unless overwritten by `RuntimeDirectoryPreserve`.
* CacheDirectory: `/var/cache/`
* StateDirectory: `/var/lib/`
* LogsDirectory:  `/var/log/`

Based on that, made the following changes:
 * Moved all directories from `/var/run/` to `/run/`. On most systems, `/var/run/` is a symlink to `/run/`, but there are systems where that is not the case, so made sure bpfd directories are always in `/run`.
 * Moved bytecode from `/var/bpfd/bytecode/` to the RuntimeDirectory `/run/bpfd/bytecode/` since `/var/bpfd` is not a directory managed by systemd.
 * Moved the socket used by gocounter from the ConfigurationDirectory `/etc/bpfd/sock/` to the StateDirectory `/var/lib/bpfd/sock`.
 * Added a new file, `bpfd-api/src/util.rs` and added all the directory strings. All hardcoded strings in the code were replaced the const strings.
 * Updated the `bpfd.service` file to create all the directories and set the mode such that files created under the directory inherit the user group. Also gave users within the same user group RO access to the directories, same as the script was doing. (mode=6755)
 * Updated the scripts to remove all the directory management (except certs, which will come in a follow-up PR).
 * Updated bpfd to create all the directories in one spot for the case where bpfd is not run as a systemd service.

**Second Commit:**
Create CA Certificate and Key, and bpfd Certificate and Key if they do not exist from within bpfd. Used https://github.com/sfackler/rust-openssl/blob/master/openssl/examples/mk_certs.rs to determine the logic to create certificates in rust.

**Third Commit:**
Add a oneshot bpfctl service that will create the needed directories (`/etc/bpfctl/certs/bpfctl`) as the user:group bpfctl:bpfd, then creates the required bpfctl certs.

This adds a new `bpfctl certs` command that creates the certs if they don't exist. If certs doesn't exist when other commands are entered (like `bpfctl list`), the commands fail.

The `./scripts/setup.sh install` still copies the binaries over. Then a command was added, `sudo systemctl enable bpfctl`. Then when the bpfd service is started, bpdctl service will be started by a `WantedBy=bpfd.service` line in the bpfctl.service file.

If certs need to be regenerated, remove the existing certs files and restart the serivces (bpfd must be restarted first):
```
sudo rm -rf /etc/bpfd/certs/
sudo rm -rf /etc/bpfctl/certs/
sudo systemctl restart bpfd
sudo systemctl restart bpfctl
```

The setup.sh script now only needs to support:
```
  ./scripts/setup.sh install
  ./scripts/setup.sh reinstall
  ./scripts/setup.sh uninstall
  ./scripts/setup.sh gocounter
```


bpfd-api/src/util.rs
```diff
@@ -0,0 +1,31 @@
...
+    //
+    // RuntimeDirectory: /run/bpfd/
+    pub const RTDIR:            &str = "/run/bpfd";
+    pub const RTDIR_BYTECODE:   &str = "/run/bpfd/bytecode";
+    pub const RTDIR_DISPATCHER: &str = "/run/bpfd/dispatchers";
+    pub const RTDIR_FS:         &str = "/run/bpfd/fs";
+    pub const RTDIR_FS_MAPS:    &str = "/run/bpfd/fs/maps";
+    pub const RTDIR_PROGRAMS:   &str = "/run/bpfd/programs";
+    //
+    // CacheDirectory: /var/cache/
```
astoycos, line 24: Is this an erroneous comment? 

Same with `LogsDirectory`

astoycos, line 24: Were the `CacheDirectory` and `LogsDirectory` Vars left here by accident?

bpfd/src/main.rs
```diff
@@ -53,23 +51,31 @@ fn main() -> anyhow::Result<()> {
             );
             setrlimit(Resource::RLIMIT_MEMLOCK, RLIM_INFINITY, RLIM_INFINITY).unwrap();
 
-            create_dir_all(BPFFS).context("unable to create mountpoint")?;
-            create_dir_all("/var/run/bpfd/programs")?;
-            create_dir_all("/var/run/bpfd/dispatchers")?;
+            create_dir_all(RTDIR)?;
+            create_dir_all(RTDIR_FS).context("unable to create mountpoint")?;
```
astoycos, line 55: Any reason for only adding context to one of these create calls?

bpfd-api/src/util.rs
```diff
@@ -0,0 +1,31 @@
...
+    // via the bpfd.service settings. They will be manually created below in case bpfd
+    // is not being run as a service.
+    //
+    // ConfigurationDirectory: /etc/bpfd/
+    pub const CFGDIR:                 &str = "/etc/bpfd";
+    pub const CFGDIR_CA_CERTS:        &str = "/etc/bpfd/certs/ca";
+    pub const CFGDIR_STATIC_PROGRAMS: &str = "/etc/bpfd/programs.d";
+    pub const CFGPATH_BPFD_CONFIG:    &str = "/etc/bpfd/bpfd.toml";
+    //
+    // RuntimeDirectory: /run/bpfd/
```
astoycos, line 16: What was the reason to move from `/var/run` again?

astoycos:
This generally looks good to me but should probably get a pass from @dave-tucker Since I'm not super experienced as to where things _should_ be living in root

bpfd-api/src/util.rs
Billy99: I left it in to spark conversation. These are the directories managed by systemd. Do some of the current runtime directories really belong in the CacheDirectory? By default, RunTime is deleted when daemon is stopped. Overriding with `RuntimeDirectoryPreserve=yes`.

Do we want logs in `/var/log/bpfd` or just with all the other logs?

Billy99: One comment wasn't enough?

bpfd/src/main.rs
Billy99: That's the way the existing code was, I just added more directories and moved them behind const strings.

bpfd-api/src/util.rs
Billy99: Systemd manages `/run/`, not `/var/run/`. On most systems, `/var/run/` is a symlink to `/run/` so it is a noop, but this is not guaranteed.

bpfd-api/src/util.rs
astoycos: GH glitch lol

bpfd/src/main.rs
astoycos: Wanna just go and add context to all of them it could be a nicer error message if ever encountered :)

dave-tucker (approved):
LGTM - this looks awesome! Thanks @Billy99

PR #195, opened 2022-11-04
Additional changes to "Move bpfd directory/file management out of setup.sh"

Billy99:
This is a follow-up to PR #191. It was decided to remove the `bpfctl` user, so the `bpfctl` service is no longer needed (which was used to make sure certs for `bpfctl` was created as the `bpfctl` user).

The following changes were made:
* Remove the bpfctl service
* Remove the bpfctl user from all the startup scripts
* Move the `bpfctl` certs creation to the start of bpfd (same as the `bpfd` certs creation)
* Move the certs generation out of bpfd-api (added in PR #191) and into bpfd
* Consolidate the configuration toml files to /etc/bpfd/bpfd.toml. So move the bpfd config.rs to bpfd-api, and remove the bpfctl version. Add new sections to the toml file for bpfctl and bpfd-agent.
* Get the k8s deployment working
* In conjunction with PR #191, reduce the modes bpfd runs in. Either as a systemd service, where the bpfd user is created by startup scripts and all the associated files are created by the service and owned by the bpfd user.  Or run manually as root and all the associated files are created by the service and owned by the root user.  Removed all the file and directory creation out of the init scripts.
* Updated all the associated documentation.


bpfd-api/src/config.rs
```diff
@@ -4,13 +4,18 @@
...
 use log::{error, warn};
 use serde::{Deserialize, Serialize};
 
+use crate::util::directories::*;
+
 #[derive(Debug, Deserialize, Default)]
 pub struct Config {
     pub tls: TlsConfig,
+    #[serde(default)]
+    pub bpfctl: BpfctlTlsConfig,
```
astoycos, line 16: Can we make this more generic and get rid of the dedicated `BpfdAgentTlsConfig`, `BpfctlTlsConfig` types? I'm assuming we generally just need a `TlsConfig` object for `bpfd` itself and then any and all `bpfd-clients`? and if we don't need a field (like `ca_cert`) then we can just make it an `Option` 

It seems like the main reason for this was for implementing different default functions but that feels clunky to include in the API

bpfd-api/src/util.rs
```diff
@@ -25,23 +33,15 @@ pub mod directories {
     pub const RTDIR_FS:         &str = "/run/bpfd/fs";
     pub const RTDIR_FS_MAPS:    &str = "/run/bpfd/fs/maps";
     pub const RTDIR_PROGRAMS:   &str = "/run/bpfd/programs";
+
```
astoycos, line 36: Nit: extra newline

bpfd-api/src/util.rs
```diff
@@ -25,23 +33,15 @@ pub mod directories {
...
     //
     // CacheDirectory: /var/cache/
+
```
astoycos, line 39: Nit: Extra newline

bpfd-api/src/util.rs
```diff
@@ -25,23 +33,15 @@ pub mod directories {
...
-    // via the bpfctl.service settings. They will be manually created in the case where bpfctl
-    // is not run as a service.
     //
-    // ConfigurationDirectory: /etc/bpfctl/
-    pub const CFGDIR_BPFCTL:            &str = "/etc/bpfctl";
-    pub const CFGDIR_BPFCTL_CERTS:      &str = "/etc/bpfctl/certs/bpfctl";
-    pub const CFGPATH_BPFCTL_CERTS_PEM: &str = "/etc/bpfctl/certs/bpfctl/bpfctl.pem";
-    pub const CFGPATH_BPFCTL_CERTS_KEY: &str = "/etc/bpfctl/certs/bpfctl/bpfctl.key";
-    pub const CFGPATH_BPFCTL_CONFIG:    &str = "/etc/bpfctl/bpfctl.toml";
+    // LogsDirectory:  /var/log/
```
astoycos, line 46: Does anything actually live in /var/log/ anymore?

bpfd-api/src/util.rs
```diff
@@ -8,15 +8,23 @@ pub mod directories {
...
-    pub const CFGDIR_STATIC_PROGRAMS: &str = "/etc/bpfd/programs.d";
-    pub const CFGPATH_BPFD_CONFIG:    &str = "/etc/bpfd/bpfd.toml";
-    pub const CFGPATH_CA_CERTS_PEM:   &str = "/etc/bpfd/certs/ca/ca.pem";
-    pub const CFGPATH_CA_CERTS_KEY:   &str = "/etc/bpfd/certs/ca/ca.key";
-    pub const CFGPATH_BPFD_CERTS_PEM: &str = "/etc/bpfd/certs/bpfd/bpfd.pem";
-    pub const CFGPATH_BPFD_CERTS_KEY: &str = "/etc/bpfd/certs/bpfd/bpfd.key";
+    pub const CFGDIR:                       &str = "/etc/bpfd";
+    pub const CFGDIR_BPFCTL_CERTS:          &str = "/etc/bpfd/certs/bpfctl";
+    pub const CFGDIR_BPFD_CERTS:            &str = "/etc/bpfd/certs/bpfd";
+    pub const CFGDIR_BPFD_AGENT_CERTS:      &str = "/etc/bpfd/certs/bpfd-agent";
```
astoycos, line 14: Again Can't we just do `bpfd-client` and then those can be used by bpfd-agent, bpfctl or both? 

(I'm not sure just asking for the sake of review :) )

Billy99: I can if that is the way we want to go. Before I started there was a bpfctl and a gocounter cert, so was just following that model. So one cert for all clients?

bpfd/src/server/certs.rs
```diff
@@ -32,72 +33,86 @@ pub enum CertsError {
...
+                &config.tls.cert,
+                CN_BPFD_NAME,
+            )
+            .await?;
+
+            Identity::from_pem(cert_pem, cert_key_pem)
+        }
+    };
+
+    // Read bpfctl Cert Key and Cert files and make sure they exist. If they don't, created them.
```
astoycos, line 88: /s/created/create

bpfd-api/src/util.rs
Billy99: No, do we want bpfd logs in something like /var/log/bpfd/?

bpfd/src/main.rs
```diff
@@ -118,10 +121,21 @@ fn has_cap(cset: caps::CapSet, cap: caps::Capability) {
 }
 
 fn drop_linux_capabilities() {
-    let res = User::from_uid(getuid()).unwrap().unwrap();
+    let uid = getuid();
+    let res = match User::from_uid(uid) {
+        Ok(res) => res.unwrap(),
+        Err(_e) => {
```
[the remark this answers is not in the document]
Billy99: I don't think we want to return error back to user. I could include it in the log. This was only hitting an error in a container. Kenjiro had a separate PR that suggested it was because the scratch image didn't have a /etc/passwd file. I'll revisit in the bpfd operator work and make sure there is a bpfd user.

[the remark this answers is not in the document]
Billy99: I think I had `Err(_)` and one of the link/format suggested the `_e`. I'll remove the `e` and see if I hit it again. Maybe not, can't remember.

bpfd-api/src/util.rs
astoycos: If that would work it actually would simplify all of this pretty dramatically... Like I don't want to have to add functionality to bpfd every time we add some sort of new client. 

And at the end of the day in any real production scenario I can't imagine BPFD would be managing it's own certs

bpfd-api/src/config.rs
Billy99: Have one `TlsConfig` type with bpfd and bpfd-client file paths. All fields are optional in the file, and if not provided will default to a value that makes sense for running on the host. K8s will need to set to filenames used by certManager or equivalent.

bpfd-api/src/util.rs
Billy99: I removed some extra new lines. I left newline between different sections. Let me know if you want those gone too, but I think is is more readable this way.

bpfd-api/src/util.rs
Billy99: I removed some extra new lines. I left newline between different sections. Let me know if you want those gone too, but I think is is more readable this way.

bpfd-api/src/util.rs
Billy99: Removed comments for directories we aren't using.

dave-tucker: Other way around 😆 
Systemd service - logs go to journal
Not systemd service - logs go to stdout (or log file)

bpfd-api/src/util.rs
Billy99: Now we have a ca, bpfd and bpfd-client cert created by bpfd if they don't exist. All clients (bpfctl, bpfd-agent, gocounter) share the single bpfd-client certs.

bpfd/src/main.rs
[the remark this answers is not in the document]
Billy99: Threw a debug log in to capture the error string.

bpfd/src/server/certs.rs
Billy99: Done

examples/gocounter/kubernetes-deployment/gocounter-bytecode.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: bpfd.io/v1alpha1
 kind: EbpfProgram
 metadata:
   name: gocounter
+  namespace: kube-system
```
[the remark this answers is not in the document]
Billy99: We are rewriting for new operator. Let's handle making this cluster scoped there?

dave-tucker (approved):
LGTM

astoycos (approved):

Issue #198, opened 2022-11-09
bpfd container produces `ENOENT` error


nak3:
### Description

When running bpfd contains, it alwasy produces `ENOENT` error.
This is because nix's `from_uid()` calls `getpwnam_r()` but scratch image does not have any name service like `/etc/passwd` so it returns error.

In fact, if I run the command with `-v /etc/passwd:/etc/passwd` the error is gone.

### Step to reproduce

Just follow https://github.com/redhat-et/bpfd/blob/main/docs/developer/image-build.md#bpfd-1
The following error is produced wen running the command:

```
$ sudo podman run --init --privileged --net=host -v /etc/bpfd/certs/:/etc/bpfd/certs/ -v /sys/fs/bpf:/sys/fs/bpf quay.io/bpfd/bpfd:mainain
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: ENOENT', bpfd/src/main.rs:121:40
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: ENOENT', bpfd/src/main.rs:121:40
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: ENOENT', bpfd/src/main.rs:121:40
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: ENOENT', bpfd/src/main.rs:121:40
thread 'tokio-runtime-worker' panicked at 'thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: ENOENT', bpfd/src/main.rs:121:40
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: ENOENT', bpfd/src/main.rs:121:40
called `Result::unwrap()` on an `Err` value: ENOENT', bpfd/src/main.rs:121:40
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: ENOENT', bpfd/src/main.rs:121:40
 ...
```

nak3:
Closing as https://github.com/redhat-et/bpfd/commit/77a660ab3d86e841a9a714254e79b8dddb7808d9 fixed this.

PR #199, opened 2022-11-09
Copy /etc/passwd file from the previous stage

nak3:
scratch image does not have a name service so
it fails to uid via `getpwnam_r()` and keep printing the error.

This patch copies `/etc/passwd` file from the previous stage.

Fix https://github.com/redhat-et/bpfd/issues/198


Billy99:
I have an alternate fix in PR #195 which unwinds the double `unwrap()` call `User::from_uid(getuid()).unwrap().unwrap()`.

See `fn drop_linux_capabilities()` in:
https://github.com/redhat-et/bpfd/pull/195/files#diff-11329bfce8dbaa78e496c0605048f3a11be2b6b9bd671126453afea141079f5b

PR #200, opened 2022-11-09
Fix path for the bpfd-agent-rbac.yaml

nak3:
`bpfd-agent-rbac.yaml` exists but `bpfd-rbac.yaml` does not exist in https://github.com/redhat-et/bpfd/tree/main/packaging/kubernetes-deployment/bpfd-core

This patch fixes it.


Billy99:
/lgtm

I have the same fix in PR #195, but we should go ahead a merge this one in case mine takes longer to get in.

Billy99 (approved):

dave-tucker:
LGTM

PR #201, opened 2022-11-10
Bump k8s-openapi and kube

nak3:
This patch just includes #194 and #162.

It needs some tweak for the update such as:
- Add `features = ["v1_23"]` to `bpfd-agent/Cargo.toml`
- Tweak `error_policy` argument

### DIFF
--- a/bpfd-agent/Cargo.toml
+++ b/bpfd-agent/Cargo.toml
@@ -20,7 +20,7 @@ validator = { version = "0.16.0", features = ["derive"] }
 anyhow = "1.0.65"
 futures = "0.3.17"
 jsonpath_lib = "0.3.0"
-k8s-openapi = { version = "0.15.0", default-features = false }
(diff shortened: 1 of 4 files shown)


bpfd-agent/Cargo.toml
```diff
@@ -20,7 +20,7 @@ validator = { version = "0.16.0", features = ["derive"] }
 anyhow = "1.0.65"
 futures = "0.3.17"
 jsonpath_lib = "0.3.0"
-k8s-openapi = { version = "0.15.0", default-features = false }
+k8s-openapi = { version = "0.16.0", features = ["v1_23"], default-features = false }
```
astoycos, line 23: Thanks for taking a look here! Hrm we should probably just bump this all the way to `v1_25` right?

nak3: Yes, changing this to `v1_25` is fine. Updated.

astoycos:
/lgtm 

Thanks @nak3!

PR #203, opened 2022-11-11
Update clap to v4 

nak3:
This patch just includes #202.

It needs to update the following annotations as per [clap's CHANGELOG.md](https://github.com/clap-rs/clap/blob/master/CHANGELOG.md).

* `parse(from_os_str)` to `value_parser`

> For #[clap(parse(from_os_str)] for PathBuf, replace it with #[clap(value_parser)] (as mentioned earlier this will call value_parser!(PathBuf) which will auto-select the right ValueParser automatically).

* `multiple = true` to `num_args(0..)`

> Replace Arg::multiple_values(true) with Arg::num_args(1..)

* Value in `conflicts_with_all` and `required_unless_present` to case sensitive.

> (derive) Leave Arg::id as verbatim casing, requiring updating of string references to other args like in conflicts_with or requires (#3282)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -433,26 +433,24 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "3.2.22"
+version = "4.0.22"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


dave-tucker:
This is awesome! Thanks @nak3

Issue #210, opened 2022-11-14
`bpfctl --help` needs cert file setup


nak3:
As per title, `bpfctl --help` does not work when certificate is not set up.

#### 1. Build bpfctl

```
$ cargo build
```

#### 2. Run with help option without cert files setup

```
$ ./target/debug/bpfctl --help
Error: Cert File does not exist

Caused by:
    No such file or directory (os error 2)
```

PR #211, opened 2022-11-14
Parse option before checking cert files

nak3:
This patch changes to parse option before checking cert files.
Otherwise, the command always fail even `--help` without cert files setup.

Fix https://github.com/redhat-et/bpfd/issues/210


dave-tucker (approved):
LGTM

PR #212, opened 2022-11-15
Refactor Bpfd-Agent and write Bpfd-Operator

astoycos:
See #224  for more details


astoycos:
@Billy99 I had to patch the API slightly see https://github.com/redhat-et/bpfd/pull/212/commits/4c1aa817dd7dd737f0b071ee7c74954360d572fb

Billy99:
> @Billy99 Heads up I had to patch the API slightly see [4c1aa81](https://github.com/redhat-et/bpfd/commit/4c1aa817dd7dd737f0b071ee7c74954360d572fb)

I had a similar change in my local branch to `EbpfProgramConfig`. Hadn't gotten to `EbpfProgram` yet :)

astoycos:
@Billy99 I know you're busy with other stuff but if you could take a quick pass here before I merge that'd be helpful :)

Billy99:
> @Billy99 I know you're busy with other stuff but if you could take a quick pass here before I merge that'd be helpful :)

Will do. I'll review in a few.

astoycos:
I also need to update the K8's Readme

bpfd-operator/controllers/bpfd-agent.go
```diff
@@ -0,0 +1,504 @@
...
+			return false, nil
+		}
+
+		// otherwise load it
+		return loadFunc(loadRequest)
+	}
+
+	// EbpfProgram exists but either EbpfProgramConfig is being deleted or node is no
+	// longer selected....unload program
+	if !ebpfProgramConfig.DeletionTimestamp.IsZero() || !isNodeSelected {
```
Billy99, line 362: Seems you are doing a bit of work if the ebpfProgram does not exist (i.e. "Always create the ebpfProgram Object if it doesn't exist" and building a loadRequest object). is it worth adding a check earlier to just bail if the ebpfProgram does not exist and node is not Selected or ebpfProgramConfig is being deleted?

bpfd-operator/cmd/bpfd-agent/main.go
```diff
@@ -0,0 +1,134 @@
...
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+
+	utilruntime.Must(bpfdiov1alpha1.AddToScheme(scheme))
+	//+kubebuilder:scaffold:scheme
+}
+
+func main() {
+	var metricsAddr string
+	var probeAddr string
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
```
Billy99, line 61: Do any of these ports need to be configurable (8080, 8081, 9443, 50051)?

bpfd-operator/controllers/bpfd-agent.go
```diff
@@ -0,0 +1,504 @@
...
+	ebpfProgramConfigs := &bpfdiov1alpha1.EbpfProgramConfigList{}
+
+	opts := []client.ListOption{}
+
+	if err := r.List(ctx, ebpfProgramConfigs, opts...); err != nil {
+		return ctrl.Result{Requeue: false}, fmt.Errorf("failed getting ebpfProgramConfigs for full reconcile %s : %v",
+			req.NamespacedName, err)
+	}
+
+	if len(ebpfProgramConfigs.Items) == 0 {
```
Billy99, line 127: Do we need a check to ensure no ebpfPrograms exist?

bpfd-operator/cmd/bpfd-agent/main.go
astoycos: They are all configurable in the deployment yaml (daemonset)

astoycos:
The clippy failures should be fixed in https://github.com/redhat-et/bpfd/pull/231

bpfd-operator/controllers/bpfd-agent.go
astoycos: So we actually always do want to create the ebpfProgram Object even if our node isn't selected, since then in the status we can report something like 

```
	// Write NodeNodeSelected status
	nodeNotSelectedCondition := metav1.Condition{
		Type:    string(EbpfProgCondNotSelected),
		Status:  metav1.ConditionTrue,
		Reason:  "nodeNotSelected",
		Message: "This node is not selected to run the ebpfProgram",
	}
```

> is it worth adding a check earlier to just bail if the ebpfProgram does not exist and node is not Selected or ebpfProgramConfig is being deleted?

Ehh I'm worried about races here.... If we rely on "the ebpfProgram Object existing" to ALWAYS mean that program is actually loaded on the node things get a little messy... The way I have things now we explicitly check if that program exists via BPFD (`nodeState` map) before doing anything which seems to cut down the number of corner cases

bpfd-operator/cmd/bpfd-agent/main.go
astoycos: But we can in the future :)

bpfd-operator/controllers/bpfd-agent.go
astoycos: No because there can/should not be any `ebpfProgram` Objects if there are no ebpfProgramConfig objects

astoycos:
OK @Billy99 this should be pretty much ready to go, At the very least I don't want it to get any bigger :)

bpfd-operator/README.md
```diff
@@ -1,80 +1,179 @@
...
+make run-on-kind
 ```
 
-2. Build and push your image to the location specified by `IMG`:
-	
-```sh
-make docker-build docker-push IMG=<some-registry>/bpfd-operator:tag
+**NOTE**
+The container images used for `bpfd`,`bpfd-agent`, and `bpfd-operator` can also be
+configured, by defualt local image builds will be used for the kind deployment.
```
Billy99, line 26:
```suggestion
configured, by default local image builds will be used for the kind deployment.
```

bpfd-operator/README.md
```diff
@@ -1,80 +1,179 @@
...
+
+```bash
+make run-on-kind
 ```
 
-2. Build and push your image to the location specified by `IMG`:
-	
-```sh
-make docker-build docker-push IMG=<some-registry>/bpfd-operator:tag
+**NOTE**
```
Billy99, line 24:
```suggestion
**NOTE**

```
Not sure if adding a new line does it, but **NOTE** is on the same line as the next sentence when the markdown is rendered. Otherwise, add a `:` to NOTE and leave it on the same line.

bpfd-operator/README.md
```diff
@@ -1,80 +1,179 @@
...
+Applications wishing to use bpfd to deploy/manage their bpf programs in kubernetes will make use of this
+object to find references to the bpfMap pin points (`spec.maps`) in order to configure their bpf programs.
+
+### Controllers
+
+The Bpfd-Operator performs a few major functions and houses two major controllers the `bpfd-agent` and `bpfd-operator`.
+
+#### bpf-agent
+
+The bpfd-agent controller is deployed alongside bpfd in a daemonset.  It's main purpose to to watch user intent (in ebpfProgramConfig Objects) and communicate with
```
Billy99, line 114:
```suggestion
The bpfd-agent controller is deployed alongside bpfd in a daemonset.  It's main purposeis to watch user intent (in ebpfProgramConfig Objects) and communicate with
```

bpfd-operator/README.md
```diff
@@ -1,80 +1,179 @@
...
+bpfd via an mTLS secured connection in order to translate the cluster-wide user-inetent to per node state.
```
Billy99, line 115:
```suggestion
bpfd via a mTLS secured connection in order to translate the cluster-wide user-intent to per node state.
```
Not sure if it's `a` or `an`, I think `a`.

bpfd-operator/config/bpfd-deployment/daemonset.yaml
```diff
@@ -53,23 +53,30 @@ spec:
...
+          readOnly: true
           # These certs are missing in the scratch image to keep it small.  Mount
           # and use the host's instead.
         - name: host-bytecode
           mountPath: /run/bpfd/bytecode
       - name: bpfd-agent 
-        image: quay.io/bpfd/bpfd-agent:main
+        command:
+        - /bpfd-agent      
+        image: bpfd-agent-image:latest
```
Billy99, line 67: Why is it `bpfd-agent-image:latest` but `bpfd-image:main`? Seems the default tags should be the same.

Billy99 (commented):
Superficial changes. Looks good. Lots of hard work here.

bpfd-operator/config/bpfd-deployment/daemonset.yaml
astoycos: Good catch!! This is actually just a tag used by `kustomize` to configure things.... Using a command like 

```
	cd config/bpfd-operator-deployment && $(KUSTOMIZE) edit set image bpfd-operator-image=${BPFD_OPERATOR_IMG} 
	cd config/bpfd-deployment && $(KUSTOMIZE) edit set image bpfd-image=${BPFD_IMG} && $(KUSTOMIZE) edit set image bpfd-agent-image=${BPFD_AGENT_IMG}
``` 

Is actually what gets these deployments get their default images (as specified in the Makefile)

bpfd-operator/README.md
astoycos: Fixed

bpfd-operator/README.md
astoycos: Removed it :)

bpfd-operator/README.md
astoycos: Done

Billy99 (approved):
/lgtm

PR #216, opened 2022-11-17
Skip rebuilding state for TC programs.

anfredette:
Rebuiding state for TC programs isn't fully supported yet. The current code loads the programs and adds them to the programs list, but doesn't rebuild the TC dispatchers.  As a result, list programs crashes bpfd if there were any TC programs saved.

### DIFF
--- a/bpfd/src/bpf.rs
+++ b/bpfd/src/bpf.rs
@@ -38,6 +38,11 @@ impl<'a> BpfManager<'a> {
                 let uuid = entry.file_name().to_string_lossy().parse().unwrap();
                 let mut program = Program::load(uuid)
                     .map_err(|e| BpfdError::Error(format!("cant read program state {}", e)))?;
+                // TODO: State reconstruction not supported for TC programs, so don't add them to the list.
+                if program.kind() == ProgramType::Tc {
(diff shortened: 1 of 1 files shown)


bpfd/src/bpf.rs
```diff
@@ -38,6 +38,11 @@ impl<'a> BpfManager<'a> {
                 let uuid = entry.file_name().to_string_lossy().parse().unwrap();
                 let mut program = Program::load(uuid)
                     .map_err(|e| BpfdError::Error(format!("cant read program state {}", e)))?;
+                // TODO: State reconstruction not supported for TC programs, so don't add them to the list.
```
Billy99, line 41: Do we not need to set the attached state on line 47?

anfredette: I don't think so.  This change avoids inserting the struct into the `programs` list, so I don't think there's any point in setting it.

anfredette:
@dave-tucker, can you take a look at this and merge it unless you see a problem?  It's a temporary fix to a potential bpfd crash introduced by your big pr.  This will get removed when I implement rebuild_state() for TC.  Thanks, Andre

dave-tucker (approved):

PR #217, opened 2022-11-25
Rebuild TC dispatcher state

anfredette:
Rebuild TC dispatcher state.

Fixes https://github.com/redhat-et/bpfd/issues/235

A couple of key points to note:
- I reduced the number of priority levels used by TC to two. That's all we really need -- one for the current dispatcher and a second one for the new dispatcher.
- I reload the TC dispatcher 3 times when rebuilding state so that it clears out the old dispatcher.

Both of these workarounds can be removed when https://github.com/aya-rs/aya/pull/445 lands upstream.

TODO:
- [x] Remove duplication in dispatcher load code for Xdp and Tc

### DIFF
--- a/bpfd/src/bpf.rs
+++ b/bpfd/src/bpf.rs
@@ -3,6 +3,7 @@
 
 use std::{collections::HashMap, convert::TryInto, fs};
 
+use anyhow::anyhow;
 use aya::{programs::TracePoint, BpfLoader};
 use bpfd_api::{config::Config, util::directories::*};
 use log::{debug, info};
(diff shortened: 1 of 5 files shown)


anfredette:
@dave-tucker, if you're okay with this approach, with some tweaks, rebuild_multiattach_dispatcher() could be called from add_single_attach_program() and remove_multi_attach_program() as well to de-duplicate some more code.

dave-tucker:
@anfredette just needs a rebase and then this is good to go

anfredette:
> LGTM. I have to say I'm not a fan of rebuilding the dispatcher 3 times, but let's merge this and fix it up when the Aya PR lands.

I'm not either, but it works, and as I said, it's just a workaround until my last Aya patch lands and we pull it in.

anfredette:
Thanks for the review, @dave-tucker.  This has been rebased.

PR #237, opened 2022-12-09
Fix image builds

astoycos:
Fix bpfd-agent and bpfd-operator
automated image builds.


astoycos:
Testing this quickly here https://github.com/astoycos/bpfd/actions/runs/3660658022 before merging

PR #238, opened 2022-12-09
Trigger image builds on workflow yaml change

astoycos:
Finish fixing operator image builds and run the image build action if we change the Action's yaml


astoycos:
Ok this causes us to trigger the image build action on merging to main OR if we changed the image-build.yaml 

I tested it on my repo and the trigger seemed to work as expected

Billy99 (approved):
/lgtm

PR #239, opened 2022-12-09
Go tc counter support

anfredette:
Support loading tc bpf program

Currently, only the basic option of loading an ingress tc program on a local host is supported.

TODO:
- Add configurable direction
- Support running in a container
- Support deploying on Kubernetes


anfredette:
I still need to do some more, but it works!

examples/go-tc-counter/go.mod
```diff
@@ -0,0 +1,21 @@
+module github.com/redhat-et/bpfd/examples/go-tc-counter
+
+go 1.18
```
Billy99, line 3: Nit: Andrew's Operator changes are using go 1.19, gobpfd and go-counter are still using 1.18. Is there a way to keep all of these in sync?

astoycos: Hrm There might be some Github action that allows us to do this

examples/go-tc-counter/bpf/tc_counter.c
```diff
@@ -0,0 +1,57 @@
...
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__type(key, __u32);
+	__type(value, datarec);
+	__uint(max_entries, TC_ACT_VALUE_MAX);
+	__uint(pinning, LIBBPF_PIN_BY_NAME);
+} tc_stats_map SEC(".maps");
+
+static __always_inline __u32 tc_stats_record_action(struct __sk_buff *skb,
```
dave-tucker, line 24: I don't recall what we did in the XDP variant, but `__always_inline` isn't required.

examples/go-tc-counter/go.mod
dave-tucker: Go workspaces?

dave-tucker (approved):
One nit, but otherwise LGTM

examples/go-tc-counter/bpf/tc_counter.c
anfredette: It's the same in the xdp variant, but I can remove it.

anfredette: Done

Billy99 (approved):
/lgtm

PR #248, opened 2022-12-13
Fix bpfd-operator module

astoycos:
Woops the name for the operator module wasn't
nested correctly fix it here


Billy99 (approved):
/lgtm
I'll merge once it passes

PR #256, opened 2022-12-20
Command error handling option

anfredette:

### DIFF
--- a/bpfd/src/lib.rs
+++ b/bpfd/src/lib.rs
@@ -12,7 +12,6 @@ mod static_program;
 mod utils;
 
 use anyhow::{bail, Context};
-//use aya::programs::ProgramError;
 use bpf::BpfManager;
 use bpfd_api::{
     config::Config,
@@ -150,11 +149,9 @@ pub async fn serve(config: Config, static_programs: Vec<StaticPrograms>) -> anyh
                             .await;
 
                     match prog_data_result {
-                        // TODO: Return better error for Err case?
-                        Err(e) => Err(BpfdError::Error(format!("{e}"))),
                         Ok(prog_data) => {
-                            let prog = match program_type {
-                                command::ProgramType::Xdp => Program::Xdp(XdpProgram {
+                            let prog_result: Result<Program, BpfdError> = match program_type {
+                                command::ProgramType::Xdp => Ok(Program::Xdp(XdpProgram {
                                     data: prog_data.clone(),
                                     info: NetworkMultiAttachInfo {
                                         if_index,
@@ -168,8 +165,8 @@ pub async fn serve(config: Config, static_programs: Vec<StaticPrograms>) -> anyh
                                         proceed_on: proc_on,
                                         if_name: iface,
                                     },
-                                }),
-                                command::ProgramType::Tc => Program::Tc(TcProgram {
+                                })),
+                                command::ProgramType::Tc => Ok(Program::Tc(TcProgram {
                                     data: prog_data.clone(),
                                     info: NetworkMultiAttachInfo {
                                         if_index,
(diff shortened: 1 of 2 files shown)


bpfd/src/lib.rs
```diff
@@ -144,43 +145,50 @@ pub async fn serve(config: Config, static_programs: Vec<StaticPrograms>) -> anyh
...
-                            direction: direction.unwrap(),
-                        }),
-                        _ => panic!("unsupported prog type"),
-                    };
-                    bpf_manager.add_program(prog)
+                            .await;
+
+                    match prog_data_result {
+                        // TODO: Return better error for Err case?
+                        Err(e) => Err(BpfdError::Error(format!("{e}"))),
```
astoycos, line 154: I think this will work alright maybe just add an error like 

```
    #[error("Failed to get BpfBytecode {0}")]
    BpfBytecodeError(#[from] ParseError)
 ```
 
 then `Err(e) => Err(BpfdError::BpfBytecodeError(e))` Should work (If we can pass the compiler :)

File: bpfd/src/lib.rs [hunk 2]
```diff
@@ -12,6 +12,7 @@ mod static_program;
 mod utils;
 
 use anyhow::{bail, Context};
+//use aya::programs::ProgramError;
```
astoycos, line 15: Nit remove this comment

bpfd/src/lib.rs
```diff
@@ -144,43 +145,50 @@ pub async fn serve(config: Config, static_programs: Vec<StaticPrograms>) -> anyh
...
+                                            priority,
+                                            name: prog_data.section_name,
+                                            attached: false,
+                                        },
+                                        proceed_on: proc_on,
+                                        if_name: iface,
+                                    },
+                                    direction: direction.unwrap(),
+                                }),
+                                _ => panic!("unsupported prog type"),
```
Billy99, line 187: Replace panic with error while you are in here?

Billy99:
Thanks @anfredette , this does the trick!

bpfd/src/lib.rs
anfredette: Will do.  I wanted to go back and figure out why that wasn't needed anymore first.

bpfd/src/lib.rs
anfredette: I'll look at that now.

Billy99:
@anfredette Single Attach has the same issue (looking at code haven't tested).  It has the `.await?` and panic.

anfredette:
> @anfredette Single Attach has the same issue (looking at code haven't tested). It has the `.await?` and panic.

Ok, I'll make the same change there.

anfredette:
> > @anfredette Single Attach has the same issue (looking at code haven't tested). It has the `.await?` and panic.
> 
> Ok, I'll make the same change there.

Done.

bpfd/src/lib.rs
```diff
@@ -195,15 +205,22 @@ pub async fn serve(config: Config, static_programs: Vec<StaticPrograms>) -> anyh
...
-                        data: ProgramData::new_from_location(location, section_name, username)
-                            .await?,
-                        info: attach,
-                    }),
-                    _ => panic!("unsupported prog type"),
+                let prog_result: Result<Program, BpfdError> = match program_type {
+                    command::ProgramType::Tracepoint => {
+                        Ok(Program::Tracepoint(TracepointProgram {
+                            data: ProgramData::new_from_location(location, section_name, username)
+                                .await?,
```
Billy99, line 212: In Multi-Attach, `new_from_location()` is one level up and result stored in `prog_data_result`. Just reading code, the `?` will return if an error is found, right?

anfredette: Are you talking about the code that loads static programs?  There are at least 3 unhandled errors there.  It also only handles Xdp programs.  Did I hear someone say something about removing static program support?

In any case, that only happens at start time.  Should we skip and log an error for static program entries with problems, or just prevent bpfd from running until they fix the config?

anfredette: Never mind.  I see it now.  I'll take care of that one too.

anfredette: Done

Billy99 (approved):
/lgtm

PR #257, opened 2022-12-20
Update the go-counter K8s example

Billy99:
Update the gocounter example to be inline with the recent bpfd-operator changes.


bpfd/src/bpf.rs
```diff
@@ -271,6 +271,7 @@ impl<'a> BpfManager<'a> {
                 return Err(BpfdError::NotAuthorized);
             }
         } else {
+            info!("InvalidID: {id}");
```
astoycos, line 274: Nit: I'd bump this to `debug`

examples/gocounter/go.mod
```diff
@@ -10,15 +10,40 @@ require (
...
+	k8s.io/client-go v0.26.0 // indirect
+	k8s.io/klog/v2 v2.80.1 // indirect
+	k8s.io/utils v0.0.0-20221107191617-1a15be271d1d // indirect
+	sigs.k8s.io/controller-runtime v0.13.0 // indirect
+	sigs.k8s.io/json v0.0.0-20220713155537-f223a00ba0e2 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
+	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace github.com/redhat-et/bpfd/examples/pkg/dyn-client => ../pkg/dyn-client
```
astoycos: So we aren't planning for external consumers to use this package right? if not I would just have a single go.mod/ go.sum for the entire examples directory so then we don't need any funny hacks like this to pickup local changes Ultimately I think it should look like 

```
examples/
    go.mod
    go.sum
    internal/ 
        - Contains shared libaries for examples/ 
    go-tc-counter/ 
         - main.go 
         - etc...
    go-xdp-counter/
         - main.go
         - etc
    go-tracepoint/
    etc...
 ```

Comment (astoycos) [reply examples/gocounter/go.mod]: Then a user can simply cd examples and do `go run --exec <EXAMPLE` similar to [how cilium organizes things](https://github.com/cilium/ebpf/tree/master/examples)

File: examples/gocounter/kubernetes-deployment/gocounter.yaml [hunk 3]
```diff
@@ -1,87 +1,48 @@
----
-apiVersion: v1
-kind: ConfigMap
+apiVersion: apps/v1
```
astoycos: Discussed in meeting , just condense this + all other kube stuff into a single yaml

examples/pkg/dyn-client/dyn-client.go
```diff
@@ -0,0 +1,99 @@
...
+
+	bpfProgramName := BpfProgramConfigName + "-" + nodeName
+	bpfProgramBlob, err := clientSet.Resource(bpfProgramResource).
+		Get(ctx, fmt.Sprintf("%s", bpfProgramName), metav1.GetOptions{})
+	if err != nil {
+		log.Printf("Error reading BpfProgram %s: %v", bpfProgramName, err)
+		return mapPath, err
+	}
+
+	programsMap, found, err := unstructured.NestedMap(bpfProgramBlob.UnstructuredContent(), "spec", "programs")
```
astoycos: Wasn't the plan here to just cast to a `bpfProgram` Object here rather than just doing fancy blob parsing?

examples/pkg/bpfd-app-client/client.go
```diff
@@ -54,44 +64,21 @@ func GetMapPath() (string, error){
...
 	}
 
-	programsMap, found, err := unstructured.NestedMap(bpfProgramBlob.UnstructuredContent(), "spec", "programs")
-	if err == nil {
-		if !found {
-			return mapPath, fmt.Errorf("programs NOT found in BpfProgram %s", bpfProgramName)
-		}
-	} else {
-		return mapPath, fmt.Errorf("Error unpacking BpfProgram %s", bpfProgramName)
+	var bpfProgram bpfdiov1alpha1.BpfProgram
```
astoycos, line 74: Ahh looks like you did it all here :)

astoycos (changes requested):
Some small things but otherwise this is good one small nit would just be to squash the middle two commits

examples/gocounter/go.mod
astoycos: Lol you did this in a later commit :)

astoycos: I think just squashing 5a671e7955538d92e36d63a826ce4633f1092954 bd0fb164bf47558c84c8fe9c099c5c4c93348ab1 and 394d05cf1c558d8854208f47548f82aa9434ab84 together will help :)

examples/pkg/dyn-client/dyn-client.go
astoycos: You did this in a later commit :)

examples/gocounter/go.mod
Billy99: Done

bpfd/src/bpf.rs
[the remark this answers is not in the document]
Billy99: Done

astoycos (approved):
/lgtm

PR #268, opened 2023-01-05
Update Go TC Counter to work with bpfd-operator

Billy99:

### DIFF
--- a/examples/go-tc-counter/main.go
+++ b/examples/go-tc-counter/main.go
@@ -14,7 +14,7 @@ import (
 
 	"github.com/cilium/ebpf"
 	gobpfd "github.com/redhat-et/bpfd/clients/gobpfd/v1"
-	"github.com/redhat-et/bpfd/examples/pkg/config-mgmt"
+	configMgmt "github.com/redhat-et/bpfd/examples/pkg/config-mgmt"
 	"github.com/redhat-et/bpfd/examples/pkg/bpfd-app-client"
 	"google.golang.org/grpc"
 )
@@ -43,7 +43,7 @@ func main() {
 	signal.Notify(stop, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
 
 	// Parse Input Parameters (CmdLine and Config File)
-	paramData, configFileData, err := configMgmt.ParseParamData(configMgmt.ProgTypeTc, DefaultConfigPath, DefaultByteCodeFile)
+	paramData, err := configMgmt.ParseParamData(configMgmt.ProgTypeTc, DefaultConfigPath, DefaultByteCodeFile)
 	if err != nil {
 		log.Printf("error processing parameters: %v\n", err)
 		return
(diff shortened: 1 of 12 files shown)


examples/go-xdp-counter/kubernetes-deployment/go-xdp-counter.yaml
```diff
@@ -60,7 +60,7 @@ spec:
         effect: NoSchedule
       containers:
       - name: go-xdp-counter
-        image: quay.io/billy99/gocounter:latest
+        image: quay.io/billy99/go-xdp-counter:latest
```
Billy99, line 63: @astoycos Do we want a copy of go-xdp-counter and go-tc-counter (userspace containers) in quay.io/bpfd/?

examples/go-tc-counter/main.go
```diff
@@ -71,196 +42,89 @@ func main() {
...
+				if err != nil {
+					conn.Close()
+					log.Print(err)
+					return
+				}
+				conn.Close()
+			}(paramData.Uuid)
+		}
+
+		// 3. Get access to our map
```
anfredette, line 124: This is a real nit, and I copied the comment here from the xdp example, so it's the same there, but we're really not getting access to our map here.  I would change it to something like "Set the path to our map".

examples/go-tc-counter/main.go
```diff
@@ -5,61 +5,32 @@ package main
...
-	BytecodeLocation string `toml:"bytecode_location"`
-}
-
-type ConfigData struct {
-	Tls    Tls
-	Config Config
-}
-
 const (
 	DefaultConfigPath     = "/etc/bpfd/gocounter.toml"
```
anfredette, line 28: Do we want to use the same `DefaultConfigPath` for all examples?  It seems like it might be better to have a file per example type.

anfredette (approved):
lgtm

examples/go-tc-counter/main.go
Billy99: @anfredette @astoycos Not sure we need to pass "config" through a TOML file any more. I used it with the original K8s deployment to pass config to the userspace container, and that container used either the UUID in the config to find it's bpf program, or used the rest of the config to send a request to bpfd. Now, in k8s deployment, something outside the userspace program creates a bpfProgram CRD Object, The userspace container just uses a fixed bpfProgram name. We could pass a bpfProgram name in if we wanted to, but probably overkill.

We should leave the TOML file reading in place to allow the certs file location to be overwritten and strip out the config portion, but it's not currently used in the k8s deployment, so would only be for running on the host.

astoycos: I was going to say the same when I was reviewing but I thought it may have been serving some explicit purpose for the non-k8s setup, I'd be happy to get rid of it!

Billy99: In progress

Billy99: Done

anfredette (approved):
lgtm

PR #274, opened 2023-01-09
Bulk rename of bytecode images

astoycos:
This commit renames all of the existing
bytecode container images from `quay.io/bpfd/bytecode:<PROG>` to `quay.io/bpfd-bytecode/<PROG>:latest`


Billy99 (approved):
/lgtm

PR #287, opened 2023-01-17
containers: Use latest stable rust

dave-tucker:
Changes 1.63.0-bullseye tag to 1-slim, which is currently Rust 1.66 on Bullseye. This fixes the image build error introduced in the latest dependency bump and hopefully reduces likelihood of it happening again while also reducing image size by moving to a slim variant

Signed-off-by: Dave Tucker <dave@dtucker.co.uk>


anfredette (approved):
Thanks for doing this.  I notice the error yesterday after I merged the dependabot prs, and was going to look into it today.

/lgtm

Issue #299, opened 2023-01-25
Support the configuration of proceed-on for TC programs


anfredette:

astoycos:
Done in #342

anfredette:
This isn't fully supported yet.

anfredette:
Completed in https://github.com/bpfd-dev/bpfd/issues/299.

PR #300, opened 2023-01-25
Restore default proceed-on behavior for TC

anfredette:
The configuration of `proceed-on` is not supported yet for TC programs, and issue #299 has been opened to track this.  However, the default configuration was working. 

This pr reverts two commits that broke the default configuration of `proceed-on` for TC programs.  Additionally, most of the code that was removed will be needed when support is added for Issue #299.

It turns out that the configuration of `proceed-on` was actually mostly working, but it was wrong in that TC would get configured with the XDP return values, and this would result in unexpected behavior.  So, in addition to reverting the two commits, this pr logs a warning message if the user configures `proceed-on` for TC.

### DIFF
--- a/bpfctl/src/main.rs
+++ b/bpfctl/src/main.rs
@@ -71,8 +71,7 @@ enum LoadCommands {
         priority: i32,
         /// Optional: Proceed to call other programs in chain on this exit code.
         /// Multiple values supported by repeating the parameter.
-        /// Possible values: [aborted, drop, pass, tx, redirect, dispatcher_return]
-        /// Default values: pass and dispatcher_return
+        /// Not yet supported for TC programs. May have unintended behavior if used.
         #[clap(long, num_args(1..))]
         proceed_on: Vec<String>,
     },
--- a/bpfctl/src/main.rs
+++ b/bpfctl/src/main.rs
@@ -69,6 +69,12 @@ enum LoadCommands {
         /// Required: Priority to run program in chain. Lower value runs first.
         #[clap(long)]
         priority: i32,
+        /// Optional: Proceed to call other programs in chain on this exit code.
+        /// Multiple values supported by repeating the parameter.
+        /// Possible values: [aborted, drop, pass, tx, redirect, dispatcher_return]
+        /// Default values: pass and dispatcher_return
+        #[clap(long, num_args(1..))]
+        proceed_on: Vec<String>,
     },
     Tracepoint {
         /// Required: The tracepoint to attach to. E.g sched/sched_switch
@@ -140,13 +146,20 @@ async fn main() -> anyhow::Result<()> {
                     direction,
                     iface,
                     priority,
+                    proceed_on,
                 } => {
                     let attach_direction = match direction.as_str() {
                         "ingress" => Direction::Ingress,
                         "egress" => Direction::Egress,
                         other => bail!("{} is not a valid direction", other),
                     };
-                    let proc_on = Vec::new();
+                    let mut proc_on = Vec::new();
+                    if !proceed_on.is_empty() {
+                        for i in proceed_on.iter() {
+                            let action = ProceedOn::try_from(i.to_string())?;
+                            proc_on.push(action as i32);
+                        }
+                    }
                     Some(load_request::AttachType::NetworkMultiAttach(
                         NetworkMultiAttach {
                             priority: *priority,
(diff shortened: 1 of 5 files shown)


bpfctl/src/main.rs
```diff
@@ -69,6 +69,12 @@ enum LoadCommands {
         /// Required: Priority to run program in chain. Lower value runs first.
         #[clap(long)]
         priority: i32,
+        /// Optional: Proceed to call other programs in chain on this exit code.
+        /// Multiple values supported by repeating the parameter.
+        /// Possible values: [aborted, drop, pass, tx, redirect, dispatcher_return]
```
Billy99, line 74: Replace the last two lines of description with "Not currently support on TC programs."

bpfd/src/lib.rs
```diff
@@ -141,7 +141,15 @@ pub async fn serve(config: Config, static_programs: Vec<StaticPrograms>) -> anyh
                             _ => proceed_on,
                         }
                     } else {
-                        proceed_on
+                        // FIXME: when proceed-on is supported for TC programs just return: proceed_on
+                        match program_type {
+                            command::ProgramType::Xdp => proceed_on,
+                            command::ProgramType::Tc => {
+                                warn!("proceed-on config not supported yet for TC and my have unintended behavior");
```
astoycos, line 148:
```suggestion
                                warn!("proceed-on config not supported yet for TC and may have unintended behavior");
```

astoycos:
@anfredette This will still be an issue right? https://github.com/redhat-et/bpfd/blob/main/bpfd-api/src/lib.rs#L58

astoycos:
i.e ProceedOn::TX == 3 and TC_ACT_PIPE == 3 so the printing of proceed on get's all wacky

bpfctl/src/main.rs
anfredette: Ok.

bpfd/src/lib.rs
anfredette: > i.e ProceedOn::TX == 3 and TC_ACT_PIPE == 3 so the printing of proceed on get's all wacky

I don't think it's an issue right now because we don't seem to use it for tc programs, but that will all need to be sorted out when I handle issue #299.

E.g., we don't print out the ProceedOn config for TC now:
```
$ bpfctl list
 UUID                                  Type  Name       Location                                                                       Metadata                                                                                            
 c600bb0a-7879-4224-8c26-39c1cee177f0  tc    drop_icmp  file:///home/afredette/bpf/drop-icmp.o                                         { "priority": 50, "iface": "vboxnet1", "position": 2, direction: INGRESS }                          
 fd537f1a-d127-4f15-b1f8-4e39757cd2f4  tc    accept     file:///home/afredette/bpf/accept-all.o                                        { "priority": 40, "iface": "vboxnet1", "position": 0, direction: INGRESS }                          
 2da6b733-fef5-4078-9ed6-9f6dbecdbf12  tc    drop_icmp  file:///home/afredette/bpf/drop-icmp.o                                         { "priority": 50, "iface": "vboxnet1", "position": 3, direction: INGRESS }                          
 687dd0b3-fb5d-4175-8dc2-47c33f54168d  tc    accept     file:///home/afredette/bpf/accept-all.o                                        { "priority": 40, "iface": "vboxnet1", "position": 1, direction: INGRESS }                          
 f20a93d2-cc2f-414d-b0f8-35ae025fb937  xdp   xdp        file:///home/afredette/projects/xdp-tutorial/basic01-xdp-pass/xdp_pass_kern.o  { "priority": 30, "iface": "vboxnet1", "position": 1, "proceed_on": ["dispatcher_return"] }         
 a4a923aa-2000-4459-b374-a50cb13d1f4e  xdp   xdp        file:///home/afredette/projects/xdp-tutorial/basic01-xdp-pass/xdp_pass_kern.o  { "priority": 30, "iface": "vboxnet1", "position": 0, "proceed_on": ["pass", "dispatcher_return"] } 
```

anfredette: This was a response to your comments below.  

I'll fix the type you pointed out above.

anfredette:
I think I've addressed your comments, but don't worry about this until after the demo -- I wouldn't want to break anything inadvertently.

Billy99 (approved):
/lgtm

PR #310, opened 2023-01-30
Additional proceedOn fix

Billy99:
PR #298 moved around where the loadRequest message was being built. The issue is that the loadRequest message was being built after some logic to clear out the proceedOn fields (TC and XDP have different defaults and it is difficult to do a proper deepEqual with the value set). So the loadRequest was being built with a cleared out ProceedOn value. Moved the building of the loadRequest to before the logic to clear the fields.

### DIFF
--- a/bpfd-operator/pkg/bpfd-agent/bpfd-agent.go
+++ b/bpfd-operator/pkg/bpfd-agent/bpfd-agent.go
@@ -268,7 +268,21 @@ func (r *BpfProgramReconciler) reconcileBpfProgramConfig(ctx context.Context,
 		}
 
 		// otherwise load it
-		return r.loadBpfdProgram(ctx, BpfProgramConfig, bpfProgram)
+
+		// TODO(astoycos) This will need to end up being a list of loadRequests
+		// if a given BpfProgramConfig selects more than one attach point
+		// (i.e if we support a pod LabelSelector for interfaces) For now
+		// we only support specifying a single BpfProgramAttachPoint in the API so
+		// there will only be a single loadRequest per BpfProgramConfig Object.
+		loadRequest, err := internal.BuildBpfdLoadRequest(BpfProgramConfig)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


astoycos:
@Billy99 IMO we should leave the load-request building where it is and instead actually fix the proceed-on hack we have currently for matching State. It should be a simple as updating the proceed-on field in the `bpfProgramConfig` in the first reconcile (If Not set) to default the XDP's/TC's BpfProgramConfig's proceed-on to whatever BPFD defaults it to 

The main issue is we don't have a good way of having conditional defaults in the K8s API (i.e if prog type TC default bpfProgramConfig to `TC_ACT_PIPE`) And the only other way to do that in K8s world is with a mutating admissions webhook which isn't something I want to deal with at the moment :) 

Ultimately then a user will be able to update the proceed_on field via K8s and everything will just work which is a win!

anfredette:
> The main issue is we don't have a good way of having conditional defaults in the K8s API

This may be a stupid question, but is it possible to have two attributes?  One configured proceed-on, which could have the value "default" (or just be empty for default) or the actual config, and another with the operational proceed-on, which will be updated to whatever it is.

astoycos:
>One configured proceed-on, which could have the value "default" (or just be empty for default) or the actual config, and another with the operational proceed-on, which will be updated to whatever it is.

Hrm do you mean both in the K8s API? not sure if I follow 

> but is it possible to have two attributes?

Hrm we might be thinking different things here but you gave me an idea :) .... we _could_ have an `xdpProceedOn` AND a `tcProceedOn` type since they are different types technically

Billy99:
Issue #313 will cover the correct implementation of ProceedOn. Merging this now to get ProceedOn for XDP working up to the same level of functionality before commit "[bpfProgramConfig delete hangs when error creating](https://github.com/redhat-et/bpfd/pull/298/commits/c87d3d9c1c167268e862572dabc26d44f0abad9d)" in PR #298.

PR #316, opened 2023-02-01
Move from scripts to kubectl plugins

Billy99:


astoycos:
Oh sweet so `kubectl` plugins were relatively  easy to get up and running with :)

astoycos (approved):

Issue #318, opened 2023-02-02
Fix TC State Reconstruction


anfredette:
Use APIs recently added to Aya to remove current hack

PR #326, opened 2023-02-03
Fix tc hack

anfredette:
This set of commits fixes Issue #318.

The code depends on some commits that are in upstream Aya, but not in our fork.  So we either need to pull those commits into our fork (https://github.com/redhat-et/aya/pull/3) or move to upstream Aya, and update the main Cargo.toml before it can be merged.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203910656683151

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,5 +2,6 @@
 members = ["bpfd", "bpfd-api", "bpfd-common", "bpfctl", "xtask", "tests/integration-test", "tests/integration-test-macros"]
 
 [patch.crates-io]
-aya = { git = "https://github.com/redhat-et/aya", branch="main" }
+#aya = { git = "https://github.com/redhat-et/aya", branch="main" }
+aya = { git = "https://github.com/anfredette/aya", branch="recon" }
 oci-distribution = {git = "https://github.com/krustlet/oci-distribution", branch="main"}
--- a/bpfd/src/multiprog/tc.rs
+++ b/bpfd/src/multiprog/tc.rs
@@ -7,7 +7,7 @@ use aya::{
     include_bytes_aligned,
     programs::{
         links::FdLink,
-        tc::{self, SchedClassifierLink},
+        tc::{self, SchedClassifierLink, TcOptions},
         Extension, Link, PinnedProgram, SchedClassifier, TcAttachType,
     },
     Bpf, BpfLoader,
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -2,5 +2,6 @@
 members = ["bpfd", "bpfd-api", "bpfd-common", "bpfctl", "xtask", "tests/integration-test", "tests/integration-test-macros"]
 
 [patch.crates-io]
-aya = { git = "https://github.com/redhat-et/aya", branch="main" }
+# aya = { git = "https://github.com/redhat-et/aya", branch="main" }
+aya = { path = "/home/afredette/projects/rh-aya/aya" }
```
anfredette, line 6: This will need to be updated or reverted depending on where the needed commits end up first.

bpfd/src/multiprog/tc.rs
```diff
@@ -28,9 +28,8 @@ use crate::{
     errors::BpfdError,
 };
 
-const DEFAULT_PRIORITY: u32 = 50;
-const MIN_TC_DISPATCHER_PRIORITY: u16 = 50;
-const MAX_TC_DISPATCHER_PRIORITY: u16 = 49;
+const DEFAULT_PRIORITY: u32 = 50; // Default priority for user programs in the dispatcher
+const TC_DISPATCHER_PRIORITY: u16 = 50; // Default TC priority for TC Dispatcher
```
astoycos, line 32: Is the `TC_DISPATCHER_PRIORITY` the priority of the actual dispatcher program itself? And if so shouldn't we be setting it at a lower one?

bpfd/src/multiprog/tc.rs
```diff
@@ -134,26 +135,26 @@ impl TcDispatcher {
...
+
         if let Some(Dispatcher::Tc(mut d)) = old_dispatcher {
-            self.current_pri = d.current_pri - 1;
-            if self.current_pri < MAX_TC_DISPATCHER_PRIORITY {
-                self.current_pri = MIN_TC_DISPATCHER_PRIORITY
-            }
-            let link_id = new_dispatcher.attach(&iface, attach_type, self.current_pri)?;
-            let link = new_dispatcher.take_link(link_id)?;
-            self.link = Some(link);
             // FIXME: TcLinks should detach on drop
```
astoycos, line 151: Wait with the new API is this fixed?

astoycos (commented):
Generally LGTM just a few questions

bpfd/src/multiprog/tc.rs
anfredette: Yes, `TC_DISPATCHER_PRIORITY` is the priority of the dispatcher.  The range is 1-64K, with lower number = higher priority, so 50 is pretty high, but I wanted to leave some higher priority slots as well.

bpfd/src/multiprog/tc.rs
anfredette: Not yet.  Drop on detach was fixed by Alessandro pretty recently.  When we go to aya/main it should work and I should be able to remove the explicit detach.

bpfd/src/multiprog/tc.rs
astoycos: I'm alright with it for now just curious

astoycos (approved):

PR #342, opened 2023-02-17
proto: Update API based on design review

dave-tucker:


bpfd-api/src/bpfd.v1.rs
```diff
@@ -1,41 +1,63 @@
...
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct NoAttachInfo {}
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct XdpAttachInfo {
+    #[prost(int32, tag = "1")]
+    pub priority: i32,
+    #[prost(string, tag = "2")]
+    pub iface: ::prost::alloc::string::String,
+    #[prost(int32, tag = "3")]
+    pub position: i32,
```
Billy99, line 21: `position` is read-only, correct (set by bpfd)? Are there any others? If so, should they be in a separate struct or is there some way to mark as read-only in the proto definition?

proto/bpfd.proto
```diff
@@ -3,52 +3,46 @@ package bpfd.v1;
 option go_package = "github.com/redhat-et/clients/gobpfd/v1;v1";
 
 service Loader {
-    rpc Load (LoadRequest) returns (LoadResponse);
+    rpc Load(LoadRequest) returns (LoadResponse);
```
astoycos, line 6:
```suggestion
    rpc Load (LoadRequest) returns (LoadResponse);
```

proto/bpfd.proto
```diff
@@ -3,52 +3,46 @@ package bpfd.v1;
...
 
-enum ProgramType {
-    XDP = 0;
-    TC = 1;
-    TRACEPOINT = 2;
+message LoadRequestCommon {
+    string location = 1;
+    string section_name = 2;
+    int32 program_type = 3;
+    optional string id = 4;
```
astoycos, line 15: nit :) 

```suggestion
    optional string uuid = 4;
```

astoycos:
Otherwise looks good!

bpfd-api/src/bpfd.v1.rs
dave-tucker: There isn't a way to make this RO.
We *could* use a separate struct, but I decided it wasn't worth it - we'll just ignore it if it's set in a request.

dave-tucker:
@astoycos Rust side is now complete and clippy is happy.
PTAL, and I might have to lean on you to do the Go changes - I took a look and decided to run for the hills 🏃

bpfd-common/src/lib.rs
```diff
@@ -23,14 +23,14 @@ unsafe impl aya::Pod for XdpDispatcherConfig {}
 // TC Defines
 pub const TC_METADATA_SECTION: &str = "tc_metadata";
 pub const TC_DISPATCHER_VERSION: u32 = 1;
-pub const TC_DISPATCHER_RETVAL: u32 = 0; //TC_ACT_OK
+pub const TC_DISPATCHER_RETVAL: u32 = 31;
```
astoycos, line 26: Is this a typo? Why 31? I don't see that as a valid return code https://elixir.bootlin.com/linux/v4.5/source/include/uapi/linux/pkt_cls.h#L82

bpfd/src/bpf.rs
```diff
@@ -64,10 +63,6 @@ impl<'a> BpfManager<'a> {
         direction: Option<Direction>,
         path: &str,
     ) -> Result<(), anyhow::Error> {
-        debug!(
```
astoycos, line 67: any reason for removing all this logging?

bpfd/src/bpf.rs
```diff
@@ -5,19 +5,18 @@ use std::{collections::HashMap, convert::TryInto, fs};
 
 use anyhow::anyhow;
 use aya::{programs::TracePoint, BpfLoader};
-use bpfd_api::{config::Config, util::directories::*};
-use log::{debug, info};
+use bpfd_api::{config::Config, util::directories::*, ProgramType, TcProceedOn};
```
astoycos, line 8: Wait a second... In bpf.rs we're internal right? Shouldn't we only be using internal types here (i.e those mirrored in command.rs)

bpfd/src/bpf.rs
```diff
@@ -400,38 +394,37 @@ impl<'a> BpfManager<'a> {
...
-                        crate::command::NetworkMultiAttach {
-                            iface: p.info.if_name.to_string(),
-                            priority: p.info.metadata.priority,
-                            proceed_on: p.info.proceed_on.clone(),
-                            direction: None,
-                            position: p.info.current_position.unwrap_or_default() as i32,
-                        },
-                    ),
+                    program_type: ProgramType::Xdp as i32,
+                    attach_info: crate::command::AttachInfo::Xdp(crate::command::XdpAttachInfo {
```
astoycos, line 398: This syntax is funky now... Is there any way we can just make it `ProgramType::Xdp::AttachInfo`

again in 397 we're using the external types defined in `bpfd-api` and then on 398 we're using internal types from `command.rs`

astoycos (commented):
Some initial comments! Overall this looks good I just want to make sure we're really explicit about where/when we use internal vs external API types 

i.e in the bpfd module we should really only be using the types defined in that module.... and all type conversion should take place in `rpc.rs`

Billy99:
I was testing with the PR. There are issues with the ProceedOn:

Load a program with no proceedon value, it succeeds but can't list:
```
$ bpfctl load --location file:///home/bmcfall/src/xdp-tutorial/basic01-xdp-pass/xdp_pass_kern.o -s "xdp" xdp -i vethdd340d3 --priority 35
9924559c-7a9e-4e67-bc76-cf2d9becd82c

$ bpfctl list
Error: error parsing proceed_on 5 is not a valid proceed-on value
```

Try to load with the XDP default values (pass and dispatcher_return) and it fails:
```
$ bpfctl load --location file:///home/bmcfall/src/xdp-tutorial/basic01-xdp-pass/xdp_pass_kern.o -s "xdp" xdp -i vethdd340d3 --priority 35 --proceed-on pass --proceed-on dispatcher_return
Error: status: Aborted, message: "failed to parse proceed_on", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Fri, 24 Mar 2023 15:26:49 GMT", "content-length": "0"} }
```

Trying to debug, but going into a series of meetings, but I will continue. But wondering of it is because proceedon is a i32, dispatcher_return value is 31, and there is a bit shift going on.

bpfd-common/src/lib.rs
dave-tucker: Not a typo - it's a special code used by the XDP dispatcher that gets returned if we've entered a stub function an extension program isn't loaded.

bpfd/src/bpf.rs
dave-tucker: yeah the values it was logging disappeared as I was updating the API so deleted it to remove compiler erros! can look at adding it back 😆

bpfd/src/bpf.rs
dave-tucker: > again in 397 we're using the external types defined in bpfd-api

Well actually it's an i32 😉 This used to be an enum in the protobuf spec...
The case for using this here is that while we *could* use the program types from aya (if they're accessible), we actually need TryFrom<i32> and TryFrom<String> for those types (for bpfctl cli and static-programs) which IIRC aren't implemented.

dave-tucker:
> Trying to debug, but going into a series of meetings, but I will continue. But wondering of it is because proceedon is a i32, dispatcher_return value is 31, and there is a bit shift going on.

ack. will look at that. will add some unit tests for proceed_on to try and figure out what's up.

anfredette (commented):
Otherwise, it looks good to me so far.

anfredette:
> > Trying to debug, but going into a series of meetings, but I will continue. But wondering of it is because proceedon is a i32, dispatcher_return value is 31, and there is a bit shift going on.
> 
> ack. will look at that. will add some unit tests for proceed_on to try and figure out what's up.

See my comments on the subject and latest commit on this subject.

anfredette:
FYI, I just force pushed a rebase to main.

astoycos:
@dave-tucker I went ahead and rebased this against #372 Since it was kind of a bear :) I'm sure I still have some cleanup to do, but I'll do that and fixup the operator tomorrow!

astoycos:
Ok so I got the new K8s API implemented but I'm still chugging through sorting out all updates to the reconciler logic locally.  It will take a bit more time unfortunately :(

bpfd-common/src/lib.rs
shaneutt: What about adding a link here to the special code, so that future readers can sort of find there way to where it originates without having to have tacit knowledge on the subject?

astoycos:
@nak3 I'd appreciate your review on the K8s API changes here as they're pretty major (specifically https://github.com/redhat-et/bpfd/pull/342/commits/af8d58a3a4c09cd5fcc8bcf2c9239f221125e60c)

bpfd-operator/apis/v1alpha1/xdpProgram_types.go
```diff
@@ -0,0 +1,84 @@
...
+
+	Spec XdpProgramSpec `json:"spec"`
+	// +optional
+	Status XdpProgramStatus `json:"status,omitempty"`
+}
+
+// +kubebuilder:validation:Enum=aborted;drop;pass;tx;redirect;dispatcher_return
+type XdpProceedOnValue string
+
+// BpfProgramConfigSpec defines the desired state of BpfProgramConfig
```
nak3, line 44:
```suggestion
// XdpProgramSpec defines the desired state of XdpProgram
```

bpfd-operator/apis/v1alpha1/xdpProgram_types.go
```diff
@@ -0,0 +1,84 @@
...
+	// is only applicable for XDP programs.
+	// NOTE: These values are not updatable following bpfProgramConfig creation.
+	// +optional
+	// +kubebuilder:default:={pass,dispatcher_return}
+	ProceedOn []XdpProceedOnValue `json:"proceedon"`
+}
+
+// XdpProgramStatus defines the observed state of XdpProgram
+type XdpProgramStatus struct {
+	// Conditions houses the global cluster state for the BpfProgram
```
nak3, line 69:
```suggestion
	// Conditions houses the global cluster state for the XdpProgram
```

bpfd-operator/apis/v1alpha1/tcProgram_types.go
```diff
@@ -0,0 +1,90 @@
...
+
+	// Direction specifies the direction of traffic the bpfprogram should
+	// attach to for a given network device, this field should only be
+	// set for programs of type TC.
+	// +kubebuilder:validation:Enum=ingress;egress
+	Direction string `json:"direction"`
+
+	// ProceedOn allows the user to call other programs in chain on this exit code.
+	// Multiple values are supported by repeating the parameter. This feature
+	// is only applicable for XDP programs.
```
nak3, line 66:
```suggestion
	// Multiple values are supported by repeating the parameter.
```
This is applicable for TC program but I think this comment should be removed.

bpfd-operator/apis/v1alpha1/xdpProgram_types.go
```diff
@@ -0,0 +1,84 @@
...
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=1000
+	Priority int32 `json:"priority"`
+
+	// ProceedOn allows the user to call other programs in chain on this exit code.
+	// Multiple values are supported by repeating the parameter. This feature
+	// is only applicable for XDP programs.
+	// NOTE: These values are not updatable following bpfProgramConfig creation.
+	// +optional
+	// +kubebuilder:default:={pass,dispatcher_return}
```
nak3, line 63:
```suggestion
	// +kubebuilder:default:={pass,dispatcher_return}
	// +kubebuilder:validation:MaxItems=6
```

nak3 (commented):
Great! Just some minor comments but LGTM overall!

bpfd-operator/apis/v1alpha1/tcProgram_types.go
```diff
@@ -0,0 +1,90 @@
...
+	// NOTE: These values are not updatable following bpfProgramConfig creation.
+	// +optional
+	// +kubebuilder:default:={ok,pipe,dispatcher_return}
+	ProceedOn []TcProceedOnValue `json:"proceedon"`
+}
+
+// TcProgramStatus defines the observed state of TcProgram
+type TcProgramStatus struct {
+	// Conditions houses the global cluster state for the BpfProgram
+	// Known .status.conditions.type are: "Available", "Progressing", and "Degraded"
```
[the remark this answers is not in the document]
astoycos: Yep you're right, I think our condition's still need some work granted this PR is large enough so I opened https://github.com/redhat-et/bpfd/issues/430

bpfd-operator/apis/v1alpha1/tcProgram_types.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +genclient
+// +genclient:nonNamespaced
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+//+kubebuilder:resource:scope=Cluster
+
```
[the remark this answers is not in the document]
astoycos: Good point! I went ahead and added some here

bpfd-operator/apis/v1alpha1/tcProgram_types.go
astoycos: Done!

bpfd-operator/apis/v1alpha1/tracepointProgram_types.go
```diff
@@ -0,0 +1,69 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec TracepointProgramSpec `json:"spec"`
+	// +optional
+	Status TracepointProgramStatus `json:"status,omitempty"`
+}
+
+// +kubebuilder:validation:Enum=aborted;drop;pass;tx;redirect;dispatcher_return
+type TracepointProceedOnValue string
```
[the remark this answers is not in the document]
astoycos: Nope :)

bpfd-operator/apis/v1alpha1/xdpProgram_types.go
astoycos: Done!

astoycos:
Thanks a bunch @nak3!!!

dave-tucker:
I'm satisfied that this works, that we have some test coverage and the code is ~ok.
Therefore I'm going to go ahead and merge this to unblock other work in bpfd that needs to build on top of this.
Thanks @astoycos for carrying this the rest of the way, I know it's been a chore!
Overall though the API is looking better, and while still not quite perfect, it's a lot better than it was before.

Issue #351, opened 2023-03-02
Update Linux Capabilities Doc


anfredette:
Update [Linux Capabilities Doc](https://bpfd.netlify.app/linux-capabilities/) with info learned during the investigation on accessing maps without CAP_BPF

Issue #349, opened 2023-03-02
Set bpfd group permissions appropriately on maps created by bpfd


anfredette:
When the bpfd service is launched, it is launched as a `bpfd` user in the `bpfd` group. A user can be added to the `bpfd` group, which allows it to access the TLS certs, so then it can send gRPC requests to bpfd to load the bpf bytecode and not need to be privileged (no CAP_BPF required). The same should be the same for accessing any maps shared between the Userspace and BPF program. 

When maps are create, they are placed in `/run/bpfd/fs/maps/<PROGRAM_UUID>/`. In bpfd.service, the `/run/bpfd/` directory is setup with mode 6770.
- 6xxx: Indicates that any files created under the directory inherit the same User and Group settings as the parent.
- x7xx: Provides Owner with Read/Write/Exec privileges.
- xx7x: Provides Group with Read/Write/Exec privileges.
- xxx0: Provides All users with No privileges.

The problem is that maps are being created 0600 (only owner access), so the Userspace programs need CAP_DAC_SEARCH to read or CAP_DAC_OVERIDE to read/write the maps. I believe this is because the maps are created, then moved into desired directory, and moved files do not inherit the parent privileges.

PR #352, opened 2023-03-03
Set bpfd group permissions appropriately on maps created by bpfd

Billy99:
Maps are being created 0500 (User RW but no Group access), so the Userspace program needs CAP_DAC_SEARCH to read or CAP_DAC_OVERIDE to read/write the maps. Update the map files after program is loaded to grant access the bpfd group.

Update the documents regarding the CAP_BPF and CAP_DAC_OVERRIDE for accessing maps.

Resolves: #349
Resolves: #351


anfredette (approved):
This works as intended, and I think this is at least a good short-term solution.  However, I wonder if it would be better to set the permissions closer to where the maps are getting pinned.  I believe this is in the Aya code, but I'm not sure of the details.  I think it would also be good to be able to set the file permissions based on the permissions on the map (e.g., don't allow writes if the map is read-only).

anfredette (changes requested):
I'm withdrawing my approval because of the integration test failure.

I think you may just need to rebase.

Billy99:
> However, I wonder if it would be better to set the permissions closer to where the maps are getting pinned.

I'm not sure if we want something like this in Aya, but maybe. Guess that's a Dave question.

I originally made the function call in `bpf.rs` in `add_multi_attach_program()` and `add_single_attach_program()`, but the `tokio::fs::set_permissions()` was an async call, which means all functions calling it need to be async, all the way up the chain. So I added the call to `set_map_permissions()` at the highest level that was already an async call. Not a rust expert, so I can rework if desired.

Billy99:
> I think it would also be good to be able to set the file permissions based on the permissions on the map (e.g., don't allow writes if the map is read-only).

That information is not in the bpfd gRPC API at the moment. We would need to add it.

anfredette (approved):
This looks good to me.

Issue #348, opened 2023-02-28
Unpin toolchain from nightly-2023-02-26


anfredette:
When https://github.com/clap-rs/clap/issues/4733 is fixed, https://github.com/redhat-et/bpfd/pull/347 can be reverted.

anfredette:
Done

PR #357, opened 2023-03-10
Unpin toolchain from nightly-2023-02-26

anfredette:
Fixes issue https://github.com/redhat-et/bpfd/issues/348.

Running `cargo xtask build-proto` also resulted in updates to `bpfd.v1.rs`.  I assume this is due to updates in the rust tools.

### DIFF
--- a/bpfd-api/src/bpfd.v1.rs
+++ b/bpfd-api/src/bpfd.v1.rs
@@ -1,3 +1,4 @@
+#[allow(clippy::derive_partial_eq_without_eq)]
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct LoadRequest {
     #[prost(string, tag = "1")]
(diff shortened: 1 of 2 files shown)


anfredette:
@dave-tucker @astoycos Do the updates to `bpfd.v1.rs` that happened after running `cargo xtask build-proto` look okay?  They are apparently unrelated to the rust toolchain updates because @Billy99 saw them too without updating rust.  Do you know what change caused them?

bpfd-api/src/bpfd.v1.rs
```diff
@@ -101,6 +112,15 @@ impl ProgramType {
             ProgramType::Tracepoint => "TRACEPOINT",
         }
     }
+    /// Creates an enum from field names used in the ProtoBuf definition.
```
astoycos, line 115: Hrm now we almost don't need this function :) https://github.com/redhat-et/bpfd/blob/main/bpfd/src/command.rs#L64

Anyways I'll leave for now but I think wherever we can use new autogenerated helpers the better

PR #362, opened 2023-03-16
Modify BpfProgramConfig to take interface selector instead of just an interface

Billy99:
In Kubernetes, it's common to not know what interface gets plugged into a pod, or which interface is the main interface of a node. This PR updates the BpfProgramConfig and BpfProgram CRD objects, as well as the code to process it to allow more generic selection of interfaces. These changes will now allow multiple interfaces on a node to be selected for a given BpfProgram.


astoycos:
Yay I is a GO @Billy99

bpfd-operator/controllers/bpfd-agent/bpfd-agent.go
```diff
@@ -138,6 +140,12 @@ func (r *BpfProgramReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			req.NamespacedName, err)
 	}
 
+	if r.NodeIface == "" {
+		if err := r.setNodeInterface(ourNode); err != nil {
```
astoycos, line 144: As mentioned earlier not sure we need this

bpfd-operator/controllers/bpfd-agent/bpfd-agent.go
```diff
@@ -241,147 +251,141 @@ func (r *BpfProgramReconciler) reconcileBpfProgramConfig(ctx context.Context,
...
-		}
-
-		// Make sure if we're not selected just exit
-		if !isNodeSelected {
-			r.Logger.V(1).Info("bpfProgramConfig does not select this node")
-			// Write NodeNodeSelected status
-			r.updateStatus(ctx, bpfProgram, BpfProgCondNotSelected)
+				// If K8s hasn't cleaned up here it means we're no longer selected
+				// write NodeNodeSelected status ignoring error (object may not exist)
+				r.updateStatus(ctx, bpfProgram, BpfProgCondNotSelected)
```
astoycos, line 332: Can you make this into a function that we return? i.e we should ALWAYS exit out of the reconcile loop if we end up changing the status

astoycos (commented):
Still have more to do, just a start :)

bpfd-operator/internal/bpfd-client-helpers.go
```diff
@@ -382,3 +400,31 @@ func PrintNodeState(state map[ProgramKey]ExistingReq) {
 		log.V(1).Info("Node State---->", "ProgramKey", k, "Value", v)
 	}
 }
+
+func ConvertToBpfProgramConfigList(BpfProgramConfig *bpfdiov1alpha1.BpfProgramConfig, NodeIface string) (nodeState []*bpfdiov1alpha1.BpfProgramConfig) {
```
astoycos, line 404: convert this to take in a bpfProgramConfig.Spec

bpfd-operator/apis/v1alpha1/shared_types.go
```diff
@@ -38,12 +38,27 @@ type BpfProgramAttachPoint struct {
...
+type InterfaceSelector struct {
+	// Interface refers to the name of a network interface to attach BPF program too.
+	// +optional
+	Interface *string `json:"interface"`
+
+	// Attach BPF program to the primary interface on the node. Only 'true' accepted.
+	// +optional
+	PrimaryNodeInterface *bool `json:"primarynodeinterface"`
+
+	//PodLabel *string `json:"podLabel"`
```
astoycos, line 53: Remove comment

bpfd-operator/bundle.Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM scratch
```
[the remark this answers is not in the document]
Billy99: Yes, I assume so as part of the `make bundle ` command.

bpfd-operator/apis/v1alpha1/bpfprogram_types.go
```diff
@@ -40,6 +40,12 @@ type BpfProgram struct {
 
 // BpfProgramSpec defines the desired state of BpfProgram
 type BpfProgramSpec struct {
+	// Node program is loaded on
+	Node string `json:"node"`
```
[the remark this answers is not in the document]
Billy99: Not intentionally non-optional. But the `BpfProgram` is programmatically created by bpfd-agent, so it is always set. Added it to help with debugging. The value is in the name, but there is not an easy way to obtain and use the node as a filter when displaying.

Will make optional on next commit.

bpfd-operator/apis/v1alpha1/shared_types.go
```diff
@@ -38,12 +38,22 @@ type BpfProgramAttachPoint struct {
...
 type ProceedOnValue string
 
+// InterfaceSelector defines interface to attach to.
+// +kubebuilder:validation:MaxProperties=1
+// +kubebuilder:validation:MinProperties=1
+type InterfaceSelector struct {
+	// Interface refers to the name of a network interface.
+	Interface *string `json:"interface"`
+	//PodLabel *string `json:"podLabel"`
+	//Primary *bool `json:"primaryInterface"`
```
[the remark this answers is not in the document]
Billy99: Leftover comment. Implemented in the next commit.

bpfd-operator/controllers/bpfd-agent/bpfd-agent.go
```diff
@@ -241,147 +243,142 @@ func (r *BpfProgramReconciler) reconcileBpfProgramConfig(ctx context.Context,
...
 	isNodeSelected = selector.Matches(nodeLabelSet)
 
-	programKey := internal.ProgramKey{
-		Name:        BpfProgramConfig.Spec.Name,
-		ProgType:    BpfProgramConfig.Spec.Type,
-		AttachPoint: internal.StringifyAttachType(&BpfProgramConfig.Spec.AttachPoint),
-	}
+	// Convert BpfProgramConfig object to a list of BpfProgramConfig, which maps any non-interface
+	// (like pod label or primary interface flag) into its actual interface or list of interfaces.
+	bpfProgramConfigList := internal.ConvertToBpfProgramConfigList(BpfProgramConfig)
```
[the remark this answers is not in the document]
Billy99: OK. After discussion, @astoycos and @Billy99 both agree, but will done in a follow-up PR because it is a lot of re-work. The follow-up PR, which will add support for specifying a pod label will include a new internal type that both the BpfProgramConfig Spec and bpfd list result structs map too.

bpfd-operator/controllers/bpfd-agent/bpfd-agent.go
Billy99: This is not the Node Name (which was mentioned above as not needed, but I think is helpful for debugging). This is where the interface that maps to the `PrimaryNodeInterface` is calculated.

bpfd-operator/internal/bpfd-client-helpers.go
Billy99: I think you meant for `BuildBpfdLoadRequest()` to take in a `bpfProgramConfig.Spec` struct and for `ConvertToBpfProgramConfigList()` to return a list of `bpfProgramConfig.Spec` instead of a list of ``bpfProgramConfig` structs.

Done on next commit.

bpfd-operator/controllers/bpfd-agent/bpfd-agent.go
```diff
@@ -241,147 +251,141 @@ func (r *BpfProgramReconciler) reconcileBpfProgramConfig(ctx context.Context,
...
 	isNodeSelected = selector.Matches(nodeLabelSet)
 
-	programKey := internal.ProgramKey{
-		Name:        BpfProgramConfig.Spec.Name,
-		ProgType:    BpfProgramConfig.Spec.Type,
-		AttachPoint: internal.StringifyAttachType(&BpfProgramConfig.Spec.AttachPoint),
-	}
+	// Convert BpfProgramConfig.Spec object to a list of BpfProgramConfig.Spec, which maps any non-interface
+	// (like pod label or primary interface flag) into its actual interface or list of interfaces.
+	bpfProgramConfigSpecList := internal.ConvertToBpfProgramConfigSpecList(&BpfProgramConfig.Spec, r.NodeIface)
```
astoycos, line 256: I'm still not stoked on this... It really explodes the complexity of this function... I get having to expand an interface selector to a list of interfaces (with a list of bpfProgramConfigs) but now we're reconciling essentially multiple config objects in `reconcileBpfProgramConfig` which is a function designed to explicitly reconcile a single one. And from my experience we're going to run into unforseen races here

bpfd-operator/controllers/bpfd-agent/bpfd-agent.go
```diff
@@ -241,147 +251,141 @@ func (r *BpfProgramReconciler) reconcileBpfProgramConfig(ctx context.Context,
...
+				// If K8s hasn't cleaned up here it means we're no longer selected
+				// write NodeNodeSelected status ignoring error (object may not exist)
+				r.updateStatus(ctx, bpfProgram, BpfProgCondNotSelected)
+			}
 
 			return false, nil
 		}
 
-		// otherwise load it
+		modBpfProgramConfigSpec.NodeSelector = metav1.LabelSelector{}
```
astoycos, line 338: Why are we zeroing this out now?

bpfd-operator/controllers/bpfd-agent/bpfd-agent.go
```diff
@@ -241,147 +251,141 @@ func (r *BpfProgramReconciler) reconcileBpfProgramConfig(ctx context.Context,
...
+					Maps:        maps,
+				}
 
-			bpfProgram.Spec.Programs[v.Uuid] = bpfdiov1alpha1.BpfProgramMeta{
-				AttachPoint: internal.AttachConversion(loadRequest),
-				Maps:        maps,
-			}
+				r.Logger.V(1).Info("Updating programs from nodestate", "Programs", bpfProgram.Spec.Programs)
+				// Update bpfProgram once successfully loaded
+				if err = r.Update(ctx, bpfProgram, &client.UpdateOptions{}); err != nil {
```
astoycos, line 379: Now in one reconcile loop we could end up updating a single bpfProgramConfig object multiple times right? That could definitely break many things

astoycos: Especially since currently we've just implemented the "default interface selector" so the race I'm worried about won't come out now since I  _think_ bpfProgramConfigSpecList should always be one (until we implement podSelector

bpfd-operator/controllers/bpfd-agent/bpfd-agent.go
```diff
@@ -506,3 +510,45 @@ func nodePredicate(nodeName string) predicate.Funcs {
...
+func (r *BpfProgramReconciler) setNodeInterface(ourNode *v1.Node) error {
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		r.Logger.Error(err, "failed to read node interfaces")
+		return err
+	}
+
+	for _, ipaddr := range ourNode.Status.Addresses {
+		r.Logger.V(1).Info("Node IP  - ", "Type", ipaddr.Type, "Address", ipaddr.Address)
+		if ipaddr.Type == v1.NodeInternalIP {
```
astoycos, line 523: It might make this logic a bit simpler to write this with `ForEachInterface` or `ForeachInterfaceAddress`

(https://github.com/tailscale/tailscale/blob/v1.38.3/net/interfaces/interfaces.go#L189) https://github.com/tailscale/tailscale/blob/v1.38.3/net/interfaces/interfaces.go#L231

astoycos (commented):
Mostly there...

I'm just really concerned about the core agent logic... Like it's already way too complicated (from my doing :) ) and this adds on to the complexity (not necessarily yet since we still are always in essence only reconciling for a single interface) but it will in the future.

bpfd-operator/controllers/bpfd-agent/bpfd-agent.go
Billy99: Wrote issue #385 come back and investigate.

bpfd-operator/apis/v1alpha1/bpfprogram_types.go
```diff
@@ -40,6 +40,14 @@ type BpfProgram struct {
 
 // BpfProgramSpec defines the desired state of BpfProgram
 type BpfProgramSpec struct {
+	// Node program is loaded on
+	// +optional
```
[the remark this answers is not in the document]
Billy99: No, not currently used by bpfd-agent. It was added to allow sorting and filtering from `kubectl` or other kube api server calls.

bpfd-operator/controllers/bpfd-agent/bpfd-agent.go
astoycos: Looks like I need to go back and remember why I did that :)

bpfd-operator/apis/v1alpha1/shared_types.go
```diff
@@ -38,12 +38,25 @@ type BpfProgramAttachPoint struct {
...
+// +kubebuilder:validation:MaxProperties=1
+// +kubebuilder:validation:MinProperties=1
+type InterfaceSelector struct {
+	// Interface refers to the name of a network interface to attach BPF program too.
+	// +optional
+	Interface *string `json:"interface"`
+
+	// Attach BPF program to the primary interface on the node. Only 'true' accepted.
+	// +optional
+	PrimaryNodeInterface *bool `json:"primarynodeinterface"`
```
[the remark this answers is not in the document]
Billy99: I also added "omitempty" to `Node` and `Type` in the `BpfProgramSpec` since they were optional as well.

astoycos (approved):

astoycos:
Some nice follow up issues to tackle here in #386 and #385 but I think this is GTG for now!

Issue #338, opened 2023-02-15
Support patching BPF Program Data


dave-tucker:
Aya's [Bpf::set_global](https://docs.aya-rs.dev/user/aya/struct.bpfloader#method.set_global) should be exposed in the bpfd API to support patching data in .rodata maps BEFORE they are frozen.

From a GRPC standpoint this could be represented as:
```proto3
map<string, bytes> global_data = 7;
```
This may not need to be optional, since an empty map will be fine.

astoycos:
Being used in netobserv -> https://github.com/netobserv/netobserv-ebpf-agent/blob/main/pkg/ebpf/tracer.go#L68

astoycos:
```
so when you issue a load request you’d include something like this in the load-request:
global_data: map[string]interface{}{"direction": 1},
not sure how easy it is from the k8s-api side to include misc k/v pairs, but it must be possible because that’s how - env works in pod specs iirc
```

astoycos:
finished with https://github.com/redhat-et/bpfd/pull/371

PR #371, opened 2023-03-23
Support setting global data in bpf programs

anfredette:
Adds support for setting global values in xdp, tc, and tracepoint programs

Implements https://github.com/redhat-et/bpfd/issues/338.

An example use is as follows:
bpfctl load -s "accept" -g GLOBAL_1=01020304 GLOBAL_2=0A0B0C0D -l file://accept-all.o tc -d ingress -i mynet1 --priority 40

This pr uses the new Aya set_global API from @alessandrod defined in https://github.com/aya-rs/aya/pull/555, which has been merged into aya/main.


anfredette:
This pr is now based on aya/main, so it can be merged after a review.

bpfd/src/command.rs
```diff
@@ -191,15 +192,17 @@ pub(crate) struct TracepointProgram {
 #[derive(Serialize, Deserialize, Clone)]
 pub(crate) struct ProgramData {
     pub(crate) location: String,
-    pub(crate) path: String,
     pub(crate) section_name: String,
+    pub(crate) global_data: HashMap<String, Vec<u8>>,
```
astoycos, line 196: Hrm should this be an `Option<HashMap<String, Vec<u8>>` ?

Also I think we should change the new_from_location `ProgramData` function to just be `new` now that we're parsing global data and a location

astoycos (commented):
Thanks! It's pretty much there just a few questions so far

bpfd/src/command.rs
anfredette: > Hrm should this be an `Option<HashMap<String, Vec<u8>>` ?

I tried defining it as `optional` in `bpfd.proto`, but I get the following error:
```
protoc failed: bpfd.proto:40:21: Field labels (required/optional/repeated) are not allowed on map fields.
```
So, I need to include an empty hashmap in `LoadRequest` if the user doesn't enter anything.  Also, @dave-tucker said in the issue:
> This may not need to be optional, since an empty map will be fine.

So, I just carried that through to bpfd.  Also, the code is simpler that way.  I can just do
```rust
for (name, value) in &program.data.global_data {
    loader.set_global(name, value.as_slice());
}
```
Instead of wrapping that same thing in an `if let`. 
However, it could be an `Option` in the `ProgramData` struct if we prefer.

anfredette: > Also I think we should change the new_from_location ProgramData function to just be new now that we're parsing global data and a location

I thought that was a weird name since it's the only `new` method we have and we're providing more than just the location.  I didn't think adding another parameter changed it, but I'm happy to change this to just `new`.

astoycos: > but I'm happy to change this to just new

I think that'd be good! 

Ok that all sounds good for now, that last thing I'm wondering is does the "save" functionality work correctly for a Hashmap i.e if bpfd restarts will it still correctly come back up? (I don't think we have any testing on this)

bpfctl/src/main.rs
```diff
@@ -74,6 +76,13 @@ enum LoadCommands {
         /// Not yet supported for TC programs. May have unintended behavior if used.
         #[clap(long, num_args(1..))]
         proceed_on: Vec<String>,
+        // /// Optional: Global variables to be set when program is loaded.
```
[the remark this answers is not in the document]
anfredette: Done

bpfd/src/command.rs
anfredette: Yes, save and load work correctly for hashmaps.  However, we don't reload the programs after a bpfd restart, so it doesn't get used.

anfredette:
This should be ready to go.

astoycos (approved):

Issue #119, opened 2022-08-30
Add config option to specify container registry credentials for pulling bytecode images


astoycos:
Originates from https://github.com/redhat-et/bpfd/pull/102#discussion_r958350941

In https://github.com/redhat-et/bpfd/pull/102 we introduced the ability to extract bytecode from container images. 

These images are pulled from remote repositories and currently only public repositories are supported.  

We need to add a config option to `bpfd.toml` to authenticate against private image repositories 

something like 

```toml 
[ registry ]

[ registry.docker-io ]
username = foo
password = bar
```

and then nfer the registry from the image path and login if required.

astoycos:
In progress here -> https://github.com/astoycos/bpfd/tree/add-registry-auth

astoycos:
Will be completed in https://github.com/redhat-et/bpfd/pull/372

astoycos:
Fixed!

PR #372, opened 2023-03-24
Image auth and pull policy

### LINKED ISSUE #223: Investigate bytecode container persistence
Today in BPFD when a user loads a program passing in a supported container image URL the container image is ALWAYS pulled. This may be painful for some users who only want to use local images or are running in air-gapped deployments. 

Since one of BPFD's major use cases will be for controlling ebpf Program usage in a kubernetes it makes sense to introduce some sort of dependency on major Container runtimes (much like [Kubelet does currently](https://kubernetes.io/docs/concepts/containers/images/)). 

IMO this functionality should be baked into core bpfd and may even require an API change to specify a "kubernetes like" [image pull policy](https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy)

astoycos:
Works to fix #223 and #119

- [x]  Fix bpfd api to add image pull policy
- [x]  Fix bpfd api to add plaintext username and password credentials for private remote repositories
- [x] Create new oci_utils module to support new api fields
- [x] Update the K8s bpfd API objects to take advantage of these new fields
- [x] Add new integration test making use of the feature


bpfd/src/command.rs
```diff
@@ -51,6 +50,50 @@ pub enum AttachType {
...
+        match bpf_api_type {
+            bpfd_api::v1::ImagePullPolicy::Always => Ok(Self::Always),
+            bpfd_api::v1::ImagePullPolicy::IfNotPresent => Ok(Self::IfNotPresent),
+            bpfd_api::v1::ImagePullPolicy::Never => Ok(Self::Never),
+        }
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct BytecodeImage {
```
astoycos, line 74: Move this and all nested types into oci-utils

astoycos: Make fields private and do everything internally 

Add constructors for the internal use case and the testing case, control all access through public functions

astoycos:
Almost done with this... Just need some fixes with the operator.

.github/workflows/build.yml
```diff
@@ -94,7 +94,7 @@ jobs:
       - uses: golangci/golangci-lint-action@v3
         with:
           # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
-          # Keep this version synced with /bpfd/bpfd-operator/hack/verify-golint.sh
+          # Keep this version synced with /bpfd/bpfd-operator/hack/verify-golint.sh and /bpfd/bpfd-operator/scripts/verify-golint.sh
```
astoycos, line 97: Fix Comment

.github/workflows/build.yml
```diff
@@ -122,7 +122,7 @@ jobs:
       - uses: golangci/golangci-lint-action@v3
         with:
           # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
-          # Keep this version synced with /bpfd/bpfd-operator/hack/verify-golint.sh
+          # Keep this version synced with /bpfd/bpfd-operator/scripts/verify-golint.sh
```
astoycos, line 125: Fix Comment

bpfd/src/command.rs
```diff
@@ -183,15 +202,36 @@ pub(crate) struct TcProgram {
...
+        Self {
+            data,
+            info,
+            direction,
+        }
+    }
+}
+
 #[derive(Serialize, Deserialize, Clone)]
 pub(crate) struct TracepointProgram {
```
astoycos, line 221: Maybe look at making this private now so we're forced to use the constructor?

bpfctl/src/main.rs
```diff
@@ -234,6 +269,115 @@ async fn main() -> anyhow::Result<()> {
...
+            let attach_type = match &l.command {
+                LoadCommands::Xdp {
+                    iface,
+                    priority,
+                    proceed_on,
+                } => {
+                    let mut proc_on = Vec::new();
+                    if !proceed_on.is_empty() {
+                        for i in proceed_on.iter() {
+                            let action = ProceedOn::try_from(i.to_string())?;
```
shaneutt, line 287: I looked around clap for a bit to see if there was higher level tooling to parse input and convert the string values to enums, e.t.c. but nothing stood out right away. Overall it would seem nice if we could make the parsing of string input into the explicit types and enum variants they correspond with happen in it's own step discreetly in time, but I don't think it needs to hold up this PR just maybe part of a later refactoring opportunity.

(feel free to consider this comment resolved at your discretion)

bpfctl/src/main.rs
```diff
@@ -234,6 +269,115 @@ async fn main() -> anyhow::Result<()> {
...
+                location: Some(location),
+                section_name: l.section_name.to_string(),
+                program_type: prog_type as i32,
+                attach_type,
+                global_data,
+            });
+            let response = client.load(request).await?.into_inner();
+            println!("{}", response.id);
+        }
+
```
shaneutt, line 380: This main is getting quite large. It might be helpful for future readers and for maintenance to compose some of the logic into discreet functions.

bpfd-operator/internal/bpfd-client-helpers.go
```diff
@@ -110,10 +119,167 @@ func LoadTLSCredentials(tlsFiles Tls) (credentials.TransportCredentials, error)
 	return credentials.NewTLS(config), nil
 }
 
-func BuildBpfdLoadRequest(bpf_program_config_spec *bpfdiov1alpha1.BpfProgramConfigSpec, name string) (*gobpfd.LoadRequest, error) {
+// Structure from https://github.com/kubernetes/kubernetes/blob/master/pkg/credentialprovider/config.go#L39
+// ContainerConfigJSON represents ~/.docker/config.json file info
```
shaneutt, line 123: Comment is not in `godoc` format:

```suggestion
// ContainerConfigJSON represents ~/.docker/config.json file info.
// Structure from https://github.com/kubernetes/kubernetes/blob/master/pkg/credentialprovider/config.go#L39
```

bpfd-operator/internal/bpfd-client-helpers.go
```diff
@@ -110,10 +119,167 @@ func LoadTLSCredentials(tlsFiles Tls) (credentials.TransportCredentials, error)
...
+			return nil, err
+		}
+
+		creds = dockercfg
+	}
+
+	return &creds, nil
+}
+
+func BuildBpfdLoadRequest(bpf_program_config_spec *bpfdiov1alpha1.BpfProgramConfigSpec, name string, c client.Client) (*gobpfd.LoadRequest, error) {
```
shaneutt, line 229: Public symbol should have a `godoc` comment

bpfd-operator/internal/bpfd-client-helpers.go
```diff
@@ -110,10 +119,167 @@ func LoadTLSCredentials(tlsFiles Tls) (credentials.TransportCredentials, error)
...
+			creds, err := ParseAuth(c, bytecodeImage.ImagePullSecret)
+			if err != nil {
+				return nil, err
+			}
+
+			domain := reference.Domain(ref)
+
+			if domain == "docker.io" || domain == "" {
+				domain = "https://index.docker.io/v1/"
+			}
```
shaneutt, line 254: For future code readers, potentially as years go on, this defaulting behavior could use some explaining comments.

shaneutt (commented):
I've taken a first pass at this with a few small comments. Given the size of this PR I I'll need to take more passes. My biggest takeaway given the large size of this PR it would seem valuable to add more [integration] tests which validate this change and keep us from making unexpected behavior changes in later iterations.

examples/go-xdp-counter/main.go
```diff
@@ -82,7 +82,7 @@ func main() {
 			c := gobpfd.NewLoaderClient(conn)
 
 			loadRequest := &gobpfd.LoadRequest{
-				Location:    paramData.BytecodeLocation,
+				Location:    &gobpfd.LoadRequest_File{File: paramData.BytecodeLocation},
```
Billy99, line 85: I think is assuming a file, but the syntax before took a file or image:
```
./go-xdp-counter -iface vethdd340d3 -location image://quay.io/bpfd-bytecode/go-xdp-counter:latest
```
OR
```
./go-xdp-counter -iface vethdd340d3 -location file:///home/bmcfall/src/bpfd/examples/go-xdp-counter/bpf_bpfel.o
```

Billy99 (commented):
Partial review, I'll finish up later.

dave-tucker:
I'm going to review this once @astoycos gives me the nod and then we'll get this one merged since it's large and blocking other work - namely #342.
Any comments that remain unaddressed will be added as GH issues so they can be addressed in follow up commits.

bpfctl/src/main.rs
astoycos: #412

examples/go-xdp-counter/main.go
astoycos: Ok I opened an issue #413 but ended up taking a stab at fixing this here @Billy99

bpfctl/src/main.rs
```diff
@@ -48,8 +51,39 @@ struct Load {
...
+    command: LoadCommands,
+}
+
+#[derive(Args)]
+struct LoadImageArgs {
+    /// Name of the ELF section from the object file.
+    #[clap(short, long, default_value = "")]
+    section_name: String,
+
+    /// Required: Container Image URL
```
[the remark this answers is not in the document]
astoycos: Added this to the examples.

dave-tucker (approved):
LGTM

Issue #379, opened 2023-03-28
Allow configuration of listening endpoint.


Molter73:
Hey Folks!

I was playing around with bpfd, trying to build and run it and was hit with the following error message:

```
Error = tonic::transport::Error(Transport, hyper::Error(Listen, Os { code: 99, kind: AddrNotAvailable, message: "Cannot assign requested address" }))
```

The address itself is not used in my system, but I found an [issue](https://github.com/hyperium/tonic/issues/485) with a similar error on tonic (which to my understanding is what tokio uses for grpc and is therefore used by bpfd), suggesting to change the address from `[::1]` to `0.0.0.0`. When I went to do that, I found the address is hardcoded [here](https://github.com/redhat-et/bpfd/blob/e592b47c32db591ef83aed693c86d58be51b5ef8/bpfd/src/lib.rs#L38). No big deal for me, I simply changed it in the code and re-run, but I do think it’d be nice to have a way to configure this, not just as a way to fix the error, but also limit the scope of addresses bpfd binds to without having to re-compile.

I know this would be super low priority, but would be up to create a PR for it if there's interest in this. In my mind we could add something like this to the existing configuration:
```toml
[endpoint]
address = "[::1]"
port = 50051
``` 

We could also do something simpler like merging address and port:
```toml
[endpoint]
address = "[::1]:50051"
``` 

I also apologize if this is something duplicated in some other issue or already addressed in an open PR, I went through them but couldn't find anything that seemed to match this.

astoycos:
Hiya @Molter73 Thanks for opening an issue! :) 

It had been on the list to make this configurable so if you want to open a PR for it that'd be great! I think 

```
[grpc.endpoint]
address = "[::1]:50051"
```
would be perfectly fine syntax. Just be sure to also fix the [hard-coded address in `bpfctl`](https://github.com/redhat-et/bpfd/blob/main/bpfctl/src/main.rs#L108) as well.

Molter73:
Well I can't find a way to tell serde to surface a nested field, think I'll just open the PR as is and maybe someone more knowledgeable than me will point me in the right direction.

Two more questions before I open the PR though:
- I see there's a (seemingly unused) `bpfd/src/config.rs` and a `bpfd-api/src/config.rs`. Updating the latter seems to be enough for the change to work, should I duplicate the change on the other one for consistency?
- I haven't looked at what documentation would need to be updated, I'll look into it, but any pointers are appreciated 🙂

Issue #317, opened 2023-02-02
Investigate using GRPC API over Unix Socket


dave-tucker:
This can avoid using the network for communication between bpfd-agent and bpfd just in case a user were to add a filter that blocked communication

Molter73:
Hey there!

I've looked into this briefly and found a great example on having tonic listen to unix sockets here: https://github.com/hyperium/tonic/tree/master/examples/src/uds

I have a few questions on implementation details though:
- I'd assume we want bpfd to always listen on a fixed path (something like `/run/bpfd/bpfd.sock`) and there should be no need to make this configurable, right?
- Do we want to listen both on the unix socket AND a network socket? If so, is it OK to have 2 separate instances of tonic servers? This we would imply we need 2 instances of `BpfLoader` as well, which I assume wouldn't be a major problem, but we could also have a single instance for the Unix socket and have messages from the network socket forwarded to it (I have no idea how we would do this though, I need to look into it if this is the implementation we want).
- What behavior do we want to implement for `bpfctl`? Should it try on the network socket first, then the unix socket if it fails or the other way around? Do we want it to always try the network socket and implement a new `--unix` argument or something of the sort for it to use the unix socket instead?

dave-tucker:
> Hey there!
> 
> I've looked into this briefly and found a great example on having tonic listen to unix sockets here: https://github.com/hyperium/tonic/tree/master/examples/src/uds
> 
> I have a few questions on implementation details though:
> 
> * I'd assume we want bpfd to always listen on a fixed path (something like `/run/bpfd/bpfd.sock`) and there should be no need to make this configurable, right?

I would opt for a configurable path.

> * Do we want to listen both on the unix socket AND a network socket? If so, is it OK to have 2 separate instances of tonic servers? This we would imply we need 2 instances of `BpfLoader` as well, which I assume wouldn't be a major problem, but we could also have a single instance for the Unix socket and have messages from the network socket forwarded to it (I have no idea how we would do this though, I need to look into it if this is the implementation we want).

I'd start with just one, and then look at enabling multiple later.
bpfd has an (undocumented) internal API that keeps RPC implementations away from the mechanics of bpf.
we would ideally spawn tasks for each instance of a listening socket, and share the same channel that sends work to the bpf thread.

We should have some config that choses the socket - ie. `unix:///run/bpfd/bpfd.sock` or `tcp://127.0.0.1:12345` etc.. either in config file or command line flags.

[...]

PR #384, opened 2023-03-29
Make listening endpoint configurable

Molter73:
The listening endpoint being used was hardcoded, which can lead to a number of problems in a production system. With this change, we allow users to configure what the listening address will be.

Fixes: #379

### DIFF
--- a/bpfctl/src/main.rs
+++ b/bpfctl/src/main.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: (MIT OR Apache-2.0)
 // Copyright Authors of bpfd
 
-use std::collections::HashMap;
+use std::{collections::HashMap, net::SocketAddr};
 
 use anyhow::{bail, Context};
 use bpfd_api::{
@@ -145,7 +145,17 @@ async fn main() -> anyhow::Result<()> {
         .domain_name("localhost")
         .ca_certificate(ca_cert)
         .identity(identity);
-    let channel = Channel::from_static("http://[::1]:50051")
+    let endpoint = &config.grpc.endpoint;
+    let address = SocketAddr::new(
+        endpoint
+            .address
+            .parse()
+            .unwrap_or_else(|_| panic!("failed to parse address '{}'", endpoint.address)),
+        endpoint.port,
+    );
+    // TODO: Use https (https://github.com/redhat-et/bpfd/issues/396)
+    let address = format!("http://{address}");
+    let channel = Channel::from_shared(address)?
         .tls_config(tls_config)?
         .connect()
         .await?;
--- a/bpfd-api/src/config.rs
+++ b/bpfd-api/src/config.rs
@@ -14,6 +14,8 @@ pub struct Config {
     #[serde(default)]
     pub tls: TlsConfig,
     pub interfaces: Option<HashMap<String, InterfaceConfig>>,
+    #[serde(default)]
+    pub grpc: Grpc,
 }
 
 #[derive(Debug, Deserialize)]
@@ -93,6 +95,37 @@ impl ToString for XdpMode {
             XdpMode::Hw => "hw".to_string(),
         }
     }
+}
+
+#[derive(Debug, Deserialize, Default)]
+pub struct Grpc {
+    #[serde(default)]
+    pub endpoint: Endpoint,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct Endpoint {
+    #[serde(default = "default_address")]
+    pub address: String,
+    #[serde(default = "default_port")]
+    pub port: u16,
+}
+
+impl Default for Endpoint {
+    fn default() -> Self {
+        Endpoint {
+            address: default_address(),
+            port: default_port(),
+        }
+    }
+}
+
+fn default_address() -> String {
+    String::from("::1")
+}
+
+fn default_port() -> u16 {
+    50051
 }
 
 pub fn config_from_file<P: AsRef<Path>>(path: P) -> Config {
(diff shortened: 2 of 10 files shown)


bpfctl/src/main.rs
```diff
@@ -105,7 +105,8 @@ async fn main() -> anyhow::Result<()> {
         .domain_name("localhost")
         .ca_certificate(ca_cert)
         .identity(identity);
-    let channel = Channel::from_static("http://[::1]:50051")
+    let address = String::from("http://") + config.grpc.endpoint.address.as_str();
```
shaneutt, line 108: I know that the protocol was already here previously, but do do we have an issue to follow up and switch this to HTTPS?

Might be good to link an issue as a TODO here?

bpfd/src/lib.rs
```diff
@@ -35,7 +35,7 @@ use crate::command::{
 
 pub async fn serve(config: Config, static_programs: Vec<StaticPrograms>) -> anyhow::Result<()> {
     let (tx, mut rx) = mpsc::channel(32);
-    let addr = "[::1]:50051".parse().unwrap();
+    let addr = config.grpc.endpoint.address.parse().unwrap();
```
shaneutt, line 38: I get that this `unwrap()` existed previously, but it also operated on an unchanging literal string before (and as such, would never fail). Now that we're taking input for the address parsing, we should remove the unwrap and handle the result here.

Molter73: This makes sense, my only question would be how do we want to handle this exactly. Do we want to fallback to the default address or stop execution? In case we'd like the latter, do we want to `expect` with a meaningful message or return the error via `?` (I'd assume we want this last one).

bpfctl/src/main.rs
Molter73: I can't see any issues for switching over to https, maybe more of a question for maintainers since I'm not that well versed in the code base and roadmap.

shaneutt: @astoycos you think we should get drop a reminder here while we're making changes so we don't forget to encrypt this before (maybe?) `v0.2.0`?

bpfd/src/lib.rs
shaneutt: @astoycos should probably chime in here but my vote would be to fail fast when user provided configuration is bad, and return the error here such as to force the user operating BPFD to fix the configuration before it will continue work.

bpfctl/src/main.rs
Billy99: I don't think we have an issue to move to https, just #317, so I suggest writing one.

bpfd/src/lib.rs
Billy99: I think we should stop execution and fail with a meaning error.

bpfd-api/src/config.rs
```diff
@@ -95,6 +97,26 @@ impl ToString for XdpMode {
...
+            endpoint: Endpoint {
+                address: String::from("[::1]:50051"),
+            },
+        }
+    }
+}
+
+#[derive(Debug, Deserialize)]
+pub struct Endpoint {
+    pub address: String,
```
Billy99, line 117: I think this should be two fields, an address and a port. If you search through the code, there are several places that use the hard coded port 50051 (XDP example, TC example and bpfd-agent). They all use:
`grpc.DialContext(ctx, "localhost:50051", grpc.WithTransportCredentials(creds))`
I think this code will want to dial "localhost:"{port}. All these instances currently read bpfd.toml to get the certs, so they can get the port as well.

bpfctl/src/main.rs
Molter73: Created a placeholder issue (#396) and added a TODO comment.

bpfd/src/lib.rs
Molter73: Changed `parse` to `expect`, hopefully that is enough.

bpfd-api/src/config.rs
Molter73: I split the fields, do you want me to update those other places that are using the hardcoded port as part of this PR? Or should that be addressed in a follow up?

Billy99: I'd say do it now if it's not too much trouble, because if the port is changed, all the hardcoded places break. But it you want to do it as a follow-up, it needs to come closely after.

bpfctl/src/main.rs
```diff
@@ -105,7 +105,10 @@ async fn main() -> anyhow::Result<()> {
         .domain_name("localhost")
         .ca_certificate(ca_cert)
         .identity(identity);
-    let channel = Channel::from_static("http://[::1]:50051")
+    let endpoint = &config.grpc.endpoint;
+    // TODO: Use https (https://github.com/redhat-et/bpfd/issues/396)
+    let address = format!("http://{}:{}", endpoint.address, endpoint.port);
```
Molter73, line 110: Just realized this will probably not work with and IPv6 address, I'll change it.

bpfd/src/lib.rs
Molter73: Clippy suggested using `unwrap_or_else(|_| panic!(...))`, I assume this is because with this approach the string formatting is only evaluated if the unwrap fails.

bpfd-api/src/config.rs
Molter73: Changed the examples and the agent to use the port from the configuration, I'm terrible at go, so I apologize in advance.

astoycos:
Small fix here @Molter73 `cargo +nightly fmt` Should be the remedy.

Molter73:
> Small fix here @Molter73 cargo +nightly fmt Should be the remedy.

Yeah, I need to get myself some hooks for checking before pushing, sorry!

shaneutt (approved):

bpfd/src/lib.rs
shaneutt: Yep sounds good!

Billy99:
So I was trying to figure out why `kubernetes-integration-tests` were failing. Something changed recently but I haven't pin-pointed why it worked before. There is an existing bug in the all the configfile processing logic. So this is what a bpfd,toml file looks like:
```
[tls] # REQUIRED
  ca_cert = "/etc/bpfd/certs/ca/ca.pem"
  cert = "/etc/bpfd/certs/bpfd/bpfd.pem"
  key = "/etc/bpfd/certs/bpfd/bpfd.key"
  client_cert = "/etc/bpfd/certs/bpfd-client/bpfd-client.pem"
  client_key = "/etc/bpfd/certs/bpfd-client/bpfd-client.key"

[grpc.endpoint]
  address = "::1"
  port = 50051
```
Fields `cert` and `key` are for bpfd certs and `client_cert` and `client_key` are for the clients (bpfctl, go counter examples, bpfd-agent). So in `bpfd-operator/internal/bpfd-client-helpers.go`, the default client paths are setup properly:
```
const (
	:
	DefaultClientCertPath  = "/etc/bpfd/certs/bpfd-client/tls.crt"
	DefaultClientKeyPath   = "/etc/bpfd/certs/bpfd-client/tls.key"
	DefaultPort            = 50051
)
```
But they are improperly assigned to  `cert` and `key` instead of `client_cert` and `client_key`:
```
func LoadConfig() ConfigFileData {
	config := ConfigFileData{
		Tls: Tls{
			CaCert: DefaultRootCaPath,
			Cert:   DefaultClientCertPath,
			Key:    DefaultClientKeyPath,
		},
		Grpc: Grpc{
			Endpoint: Endpoint{
				Port: DefaultPort,
			},
		},
	}
```
And in `LoadTLSCredentials()`, when pulling the Client Certs, it's using `cert` and `key` instead of `client_cert` and `client_key`:
```
	clientCert, err := tls.LoadX509KeyPair(tlsFiles.Cert, tlsFiles.Key)
```

So the only reason I can think it worked before is that bpfd-agent didn't have a valid bpfd.toml file, so when it failed to read it, it was using the defaults for `cert` and `key`. Now it has a valid bpfd.toml, and it trying to read client certs from the wrong directory.

So the fix is:
```
$ git diff bpfd-operator/internal/bpfd-client-helpers.go
diff --git a/bpfd-operator/internal/bpfd-client-helpers.go b/bpfd-operator/internal/bpfd-client-helpers.go
index c53c205..87c6137 100644
--- a/bpfd-operator/internal/bpfd-client-helpers.go
+++ b/bpfd-operator/internal/bpfd-client-helpers.go
@@ -45,6 +45,8 @@ const (
        bpfdMapFs              = "/run/bpfd/fs/maps"
        DefaultConfigPath      = "/etc/bpfd/bpfd.toml"
        DefaultRootCaPath      = "/etc/bpfd/certs/ca/ca.crt"
+       DefaultCertPath        = "/etc/bpfd/certs/bpfd/tls.crt"
+       DefaultKeyPath         = "/etc/bpfd/certs/bpfd/tls.key"
        DefaultClientCertPath  = "/etc/bpfd/certs/bpfd-client/tls.crt"
        DefaultClientKeyPath   = "/etc/bpfd/certs/bpfd-client/tls.key"
        DefaultPort            = 50051
@@ -53,9 +55,11 @@ const (
 var log = ctrl.Log.WithName("bpfd-internal-helpers")
 
 type Tls struct {
-       CaCert string `toml:"ca_cert"`
-       Cert   string `toml:"cert"`
-       Key    string `toml:"key"`
+       CaCert     string `toml:"ca_cert"`
+       Cert       string `toml:"cert"`
+       Key        string `toml:"key"`
+       ClientCert string `toml:"client_cert"`
+       ClientKey  string `toml:"client_key"`
 }
 
 type Endpoint struct {
@@ -74,9 +78,11 @@ type ConfigFileData struct {
 func LoadConfig() ConfigFileData {
        config := ConfigFileData{
                Tls: Tls{
-                       CaCert: DefaultRootCaPath,
-                       Cert:   DefaultClientCertPath,
-                       Key:    DefaultClientKeyPath,
+                       CaCert:     DefaultRootCaPath,
+                       Cert:       DefaultCertPath,
+                       Key:        DefaultKeyPath,
+                       ClientCert: DefaultClientCertPath,
+                       ClientKey:  DefaultClientKeyPath,
                },
                Grpc: Grpc{
                        Endpoint: Endpoint{
@@ -117,7 +124,44 @@ func LoadTLSCredentials(tlsFiles Tls) (credentials.TransportCredentials, error)
        }
 
        // Load client's certificate and private key
-       clientCert, err := tls.LoadX509KeyPair(tlsFiles.Cert, tlsFiles.Key)
+       clientCert, err := tls.LoadX509KeyPair(tlsFiles.ClientCert, tlsFiles.ClientKey)
        if err != nil {
                return nil, err
        }
```
Similar changes are needed in the go examples directory.

Molter73:
Thanks for looking into the errors @Billy99! I'll add a commit with the fix ASAP.

Billy99 (approved):
/lgtm

PR #405, opened 2023-04-13
Handle SIGTERM gracefully

### LINKED ISSUE #175: bpfd: Initiate graceful shutdown on SIGTERM

Molter73:
A new tokio task is spawned and set to wait for both SIGTERM and SIGINT. When either of these signals is received a message on a oneshot channel is issued to trigger the tonic service to shutdown, which in turn also makes the main loop exit and cleanly stops bpfd.

Fixes #175

### DIFF
--- a/bpfd/src/lib.rs
+++ b/bpfd/src/lib.rs
@@ -20,10 +20,16 @@ use bpfd_api::{config::Config, util::directories::RTDIR_FS_MAPS, v1::loader_serv
 pub use certs::get_tls_config;
 use command::{Command, TcProgram, TcProgramInfo, TracepointProgram, TracepointProgramInfo};
 use errors::BpfdError;
-use log::info;
+use futures::{channel::oneshot, FutureExt};
+use log::{debug, info};
 use rpc::{intercept, BpfdLoader};
 use static_program::get_static_programs;
-use tokio::{net::UnixListener, sync::mpsc};
+use tokio::{
+    net::UnixListener,
+    select,
+    signal::unix::{signal, SignalKind},
+    sync::mpsc,
+};
 use tokio_stream::wrappers::UnixListenerStream;
 use tonic::transport::{Server, ServerTlsConfig};
 use utils::{get_ifindex, set_dir_permissions, set_file_permissions};
@@ -35,6 +41,8 @@ const SOCK_MODE: u32 = 0o0770;
 
 pub async fn serve(config: Config, static_program_path: &str) -> anyhow::Result<()> {
     let (tx, mut rx) = mpsc::channel(32);
+    let mut shutdown_senders = Vec::new();
+    let mut task_handles = Vec::new();
     let endpoint = &config.grpc.endpoint;
 
     // Listen on Unix socket
@@ -50,16 +58,19 @@ pub async fn serve(config: Config, static_program_path: &str) -> anyhow::Result<
 
     let loader = BpfdLoader::new(tx.clone());
 
+    let (shutdown_tx, shutdown_rx) = oneshot::channel();
+    shutdown_senders.push(shutdown_tx);
+
     let serve = Server::builder()
         .add_service(LoaderServer::new(loader))
-        .serve_with_incoming(uds_stream);
+        .serve_with_incoming_shutdown(uds_stream, shutdown_rx.map(drop));
 
-    tokio::spawn(async move {
+    task_handles.push(tokio::spawn(async move {
         info!("Listening on {}", unix);
         if let Err(e) = serve.await {
-            eprintln!("Error = {e:?}");
+            panic!("Error = {e:?}");
         }
-    });
+    }));
 
     // Listen on TCP socket
     let addr = SocketAddr::new(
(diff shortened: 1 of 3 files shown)


bpfd/src/lib.rs
```diff
@@ -48,6 +54,17 @@ pub async fn serve(config: Config, static_programs: Vec<StaticPrograms>) -> anyh
...
+    let mut sigint = signal(SignalKind::interrupt())?;
+    let mut sigterm = signal(SignalKind::terminate())?;
+    let signal_handler = tokio::spawn(async move {
+        select! {
+            _ = sigint.recv() => {debug!("Received SIGINT")},
+            _ = sigterm.recv() => {debug!("Received SIGTERM")},
+        }
+
+        shutdown_tx.send(()).unwrap();
+    });
```
dave-tucker, line 66:
```suggestion
    tokio::spawn(async move {
        let mut sigint = signal(SignalKind::interrupt()).unwrap();
        let mut sigterm = signal(SignalKind::terminate()).unwrap();
        loop {
           select! {
                _ = sigint.recv() => {debug!("Received SIGINT")},
                _ = sigterm.recv() => {debug!("Received SIGTERM")},
            }
            shutdown_tx.send(()).unwrap();
        }
    });

```

dave-tucker: This moves sending the shutdown command based on received signal into its own task.

bpfd/src/lib.rs
```diff
@@ -59,12 +76,13 @@ pub async fn serve(config: Config, static_programs: Vec<StaticPrograms>) -> anyh
     let serve = Server::builder()
         .tls_config(tls_config)?
         .add_service(LoaderServer::with_interceptor(loader, intercept))
-        .serve(addr);
+        .serve_with_shutdown(addr, shutdown_rx.map(drop));
```
dave-tucker, line 79:
```suggestion
        .serve_with_shutdown(addr, shutdown_rx.has_changed().map(|_|()));
```

bpfd/src/lib.rs
```diff
@@ -59,12 +76,13 @@ pub async fn serve(config: Config, static_programs: Vec<StaticPrograms>) -> anyh
...
 
     tokio::spawn(async move {
         info!("Listening on {addr}");
         if let Err(e) = serve.await {
             eprintln!("Error = {e:?}");
+            signal_handler.abort();
```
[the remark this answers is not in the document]
dave-tucker: This is only going to terminate the the listener though, not the larger loop below that receives messages from the channel. We don't want to panic here as it will prevent anything else from gracefully cleaning up, but the error should be logged.

dave-tucker (changes requested):
Overall this is looking good, thanks for tackling this.

I think you need to take a look at https://github.com/redhat-et/bpfd/pull/405/files#diff-1fee7d66fc3bd317e847b5e924f63dcde292b9719676529b6d773460371c57f4L128 before we merge this.

That should probably become something like:

```
loop {
        select! {
            biased;

            // check shutdown branch first
            _ = shutdown_rx.changed() => break,
            // handle message from channel
            msg =  rx.recv().await => {
                if let Some(cmd) = msg {
                    match cmd {
                        // Command handling logic goes here
                    }
                } 
            i = some_operation(42) => {
                println!("Result is {i}");
                unsafe { libc::raise(libc::SIGTERM)};
            },
        }
    }
```

bpfd/src/lib.rs
Molter73: This is giving me an error that I'm not smart enough to solve, `serve_with_shutdown` [expects a future](https://docs.rs/tonic/latest/tonic/transport/server/struct.Router.html#method.serve_with_shutdown) and starts shutdown when the future is solved, with this change I get:

```
error[E0277]: `Result<(), tokio::sync::watch::error::RecvError>` is not a future
   --> bpfd/src/lib.rs:79:36
    |
79  |         .serve_with_shutdown(addr, shutdown_rx.has_changed().map(|_| ()));
    |          -------------------       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `Result<(), tokio::sync::watch::error::RecvError>` is not a future
    |          |
    |          required by a bound introduced by this call
    |
    = help: the trait `std::future::Future` is not implemented for `Result<(), tokio::sync::watch::error::RecvError>`
    = note: Result<(), tokio::sync::watch::error::RecvError> must be a future or must implement `IntoFuture` to be awaited
note: required by a bound in `Router::<L>::serve_with_shutdown`
   --> /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/tonic-0.8.3/src/transport/server/mod.rs:605:41
    |
605 |     pub async fn serve_with_shutdown<F: Future<Output = ()>, ResBody>(
    |                                         ^^^^^^^^^^^^^^^^^^^ required by this bound in `Router::<L>::serve_with_shutdown`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `bpfd` due to previous error
```

All examples on the tonic repo use either `futures` or `futures-util`, I'll look into it a bit further but any help would be appreciated 🙂

dave-tucker: :( Looks like we need futures for FutureExt, which allows this to compile since you can `.map` on a future.

bpfd/src/lib.rs
Molter73: The `loop` part of this change makes it so this task never exits and, in turn, prevents bpfd from stopping as well. Most examples I find in the tokio and tonic docs just let the task finish once the signal is received, alternatively, we would need the main task to be able to remotely stop this with a sync mechanism (but that just seems unnecessarily complicated).

Molter73:
Hey @dave-tucker, sorry I took so long to address your main comment. I've changed the logic around so the signal handling task stops the main loop and it stops the listening servers. `cargo fmt` freaked out a bit on me and stopped formatting altogether when I started indenting inside `select!` though, not sure what that was about.

I though of an alternative implementation, adding a `Stop` command that the signal handling task could send over the same channel the listener tasks use, the change would be smaller and more concise, but I'm not sure adding a command is a thing we want for this, let me know what you think about this alternative.

Molter73:
Also, I see codecov is complaining my code is not covered by tests, should I add a test or something?

dave-tucker:
> Also, I see codecov is complaining my code is not covered by tests, should I add a test or something?
No you're ok 😆 coverage isn't quite where we'd like it anyway and the code is almost impossible to test right now 😢 

I decided to carry your last iteration of this PR in #472 - I hope you don't mind.
I ended up removing the dep on `futures` to address your [comment](https://github.com/bpfd-dev/bpfd/pull/405#discussion_r1169760247) then got stuck in a rabbit hole 🐰 of refactoring and cleanup.

Molter73:
Since the changes have moved to #472 and that is much more complete solution, I think we can go ahead and close this PR.

PR #410, opened 2023-04-17
Use Unix Domain Sockets

### LINKED ISSUE #317: Investigate using GRPC API over Unix Socket

Molter73:
Allow bpfd to listen on uds. This listener does no TLS checking. The location for the socket is configurable through the grpc.endpoint.unix field in bpfd.toml.

bpfctl now attempts to communicate over the UDS before attempting the tcp socket with TLS.

Fixes #317

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -257,6 +257,7 @@ dependencies = [
  "tokio",
  "toml",
  "tonic",
+ "tower",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


dave-tucker (approved):
Thanks @Molter73! This LGTM

Issue #397, opened 2023-04-04
Move bpfd to a new GitHub Organization


anfredette:
bpfd is currently hosted in the Red Hat Emerging Technologies GitHub organization.  As bpfd matures and is gaining interest from outside of Red Hat, it seems appropriate to move the bpfd repo into its own organization.

There is already an org called "[bpfd](https://github.com/bpfd)" which is private, so we need to find another name.  I propose we move it to an org called "[bpfd-dev](https://github.com/bpfd-dev)" (which I have reserved pending our decision).

Please comment on this issue with your opinions on the move.

astoycos:
@shaneutt we'd like you opinion here :)

stflaherty:
I support the move to Org bpfd-dev

shaneutt:
If the intention is to make an org that's _solely_ focused on BPFD then `bpfd-dev` seems OK to me. If the intention is to make an org that is more generally focused on solving higher level and management problems in the ebpf space, maybe we would consider something like `ebpf-tools` (which is available at the time of writing)? Just double checking what our goals as an org are.

anfredette:
Good question.  I was thinking that this would be for bpfd only, but the broader umbrella is worth thinking about.  Does anyone have opinions on this alternative?

astoycos:
I think for now our main focus _is_ a bit narrower and focused on bpfd specifically.  I worry if we make more of an abstract repo it could lead to some confusion.  In the future if the work stored there becomes broader we could always think about moving it again or renaming the org if needed.

shaneutt:
Ok, +1 for `bpfd-dev` then :+1:

anfredette:
Done.

Moved to https://github.com/bpfd-dev/

PR #417, opened 2023-04-20
github: Assign new issues to project

dave-tucker:
This probably won't work because its and org project not a repo project

### DIFF
--- a/.github/workflows/project.yml
+++ b/.github/workflows/project.yml
@@ -1,0 +1,22 @@
+name: Auto Assign to Project(s)
+
+on:
+  issues:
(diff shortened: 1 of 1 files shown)


astoycos:
I can just make it a repo project (or make a new one linked to the repo) right @dave-tucker?

dave-tucker:
> I can just make it a repo project (or make a new one linked to the repo) right @dave-tucker?

Maybe :) But you'd need to re-do all those fancy views. Either way it would be good to do before #397

astoycos:
Closed by #439

PR #420, opened 2023-04-21
Dispatcher from image

dave-tucker:
Depends on #342


bpfd/src/multiprog/xdp.rs
```diff
@@ -69,13 +66,23 @@ impl XdpDispatcher {
         };
 
         debug!("xdp dispatcher config: {:?}", config);
-
+        let image = BytecodeImage::new(
+            "quay.io/bpfd/xdp-dispatcher:v1".to_string(),
```
astoycos, line 70: Should we be doing the same for the TC dispatcher as well?

anfredette: I was going to ask the same question :-).  It seems like it would be a good idea.

anfredette (approved):
I assume that a future step would be to have libxdp release an image that we both use.

LGTM.

PR #421, opened 2023-04-21
Fixes for accessing bpfd via unix socket

Billy99:
There were two issues found:
* If `bpfctl` failed to access the Unix Socket, the command bailed instead of trying to us TCP over TLS
* `bpfctl` was failing to use the Unix Socket because the permissions on the socket were not setup as RW for the bpfd User Group


bpfctl/src/main.rs
```diff
@@ -173,13 +174,18 @@ async fn main() -> anyhow::Result<()> {
...
-    match execute_request(&cli.command, channel).await {
-        Ok(_) => return Ok(()),
-        Err(e) => eprintln!("Error = {e:?}"),
+        .await
+    {
+        Ok(channel) => {
+            info!("Using UNIX socket as transport");
+            match execute_request(&cli.command, channel).await {
+                Ok(_) => return Ok(()),
+                Err(e) => eprintln!("Error = {e:?}"),
```
Molter73, line 185: If we are using `log`, maybe we want to change all instance of `eprintln!` to either `warn!` or `error!`, seems a little confusing when one would be used over the other to me, WDYT?

bpfd/src/utils.rs
```diff
@@ -23,21 +23,34 @@ pub(crate) fn get_ifindex(iface: &str) -> Result<u32, BpfdError> {
...
-                file.unwrap().path(),
-                fs::Permissions::from_mode(0o0660),
+                file.as_ref().unwrap().path(),
+                fs::Permissions::from_mode(mode),
             )
             .await)
                 .is_err()
             {
-                info!("Unable to set permissions on map. Continuing");
+                info!(
```
Molter73, line 39: I know this was set as info before, it feels more like a warning, since it could trigger errors in subsequent operations.

bpfd/src/utils.rs
```diff
@@ -23,21 +23,34 @@ pub(crate) fn get_ifindex(iface: &str) -> Result<u32, BpfdError> {
...
 }
 
-pub(crate) async fn set_map_permissions(directory: &str) {
+pub(crate) async fn set_dir_permissions(directory: &str, mode: u32) {
     // Determine if User Group exists, if not, do nothing
     if get_group_by_name(USRGRP_BPFD).is_some() {
         // Iterate through the files in the provided directory
         for file in fs::read_dir(directory).unwrap() {
-            // Set the permissions on the map such that members of the "bpfd" group can read/write
+            // Set the permissions on the file based on input
```
Molter73, line 31: I think we could change the content of this for loop to directly call `set_file_permissions`, it would re-check `USRGRP_BPFD`, but I don't think that's a big deal, WDYT?

```rs
        for file in fs::read_dir(directory).unwrap() {
            match file {
                Ok(file) => set_file_permissions(file.path().to_str().unwrap(), mode).await,
                Err(e) => warn!("Failed to read file: {e}"),
            }
        }
```

bpfd/src/utils.rs
```diff
@@ -23,21 +23,34 @@ pub(crate) fn get_ifindex(iface: &str) -> Result<u32, BpfdError> {
...
         }
     }
 }
+
+pub(crate) async fn set_file_permissions(path: &str, mode: u32) {
+    // Determine if User Group exists, if not, do nothing
+    if get_group_by_name(USRGRP_BPFD).is_some() {
+        // Set the permissions on the file based on input
+        if (tokio::fs::set_permissions(path, fs::Permissions::from_mode(mode)).await).is_err() {
+            info!("Unable to set permissions on file {}. Continuing", path);
```
Molter73, line 53: Same as the previous comment on info vs warn, this feels more like a warning to me.

Molter73 (approved):
It seems I can't set my previous comments to resolved and this probably has no weight on the PR being able to be merged, but it still LGTM!

astoycos (approved):
/lgtm thanks @Billy99

PR #438, opened 2023-04-27
Fix Stale module versions after repo move

astoycos:
Fixes stale versioning after repository move to bpfd-dev.


anfredette (approved):
lgtm

PR #446, opened 2023-04-28
bpfd: Fix UUID Validation

dave-tucker:
This reverts parts of the 624b991 that changed bpfd's internal ID from UUID to String, which therefore removed UUID validation.

UUIDs coming over gRPC are String, since using a byte-based encoding might not be the easiest thing to use in all potential client languages.

However, once received from gRPC, UUIDs should be stored in UUID format, and any parsing failures should be returned to the caller.


astoycos:
Good catch, thanks @dave-tucker! 

LGTM

astoycos (approved):
LGTM

Issue #398, opened 2023-04-05
Bumps openssl from 0.10.47 to 0.10.49


astoycos:
Dependabot tried to do this manually and failed We'll need to go back and do it manually making sure things aren't broken.

Basically re-open https://github.com/redhat-et/bpfd/pull/389

astoycos:
```
2023/04/18 12:32:08 Reading /etc/bpfd/gocounter.toml ...
2023/04/18 12:32:08 Read /etc/bpfd/gocounter.toml failed: err open /etc/bpfd/gocounter.toml: no such file or directory
2023/04/18 12:32:08 rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: x509: certificate is valid for localhost,IP:127.0.0.1, not localhost"
```

PR #448, opened 2023-04-28
With openssl bump to 0.10.52, tls broken

Billy99:
Connection to bpfd over TLS broken with the bump to openssl 0.10.52.

$ ./go-xdp-counter -iface vethdd340d3
2023/04/28 08:18:24 Using Input: Interface=vethdd340d3 Priority=50 Source=0xc000548670 2023/04/28 08:18:24 Reading /etc/bpfd/gocounter.toml ... 2023/04/28 08:18:24 Read /etc/bpfd/gocounter.toml failed: err open /etc/bpfd/gocounter.toml: no such file or directory 2023/04/28 08:18:24 rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: tls: failed to verify certificate: x509: certificate is valid for localhost,IP:127.0.0.1, not localhost"

NOTE: openssl was bumped in the Cargo.lock file, but wasn't bumped in bpfd/Cargo.toml,
      so bumped in toml file as well.

### DIFF
--- a/bpfd/src/certs.rs
+++ b/bpfd/src/certs.rs
@@ -262,6 +262,7 @@ async fn generate_cert(
 
     let subject_alt_name = SubjectAlternativeName::new()
         .dns("localhost,IP:127.0.0.1")
+        .dns("localhost")
         .build(&cert_builder.x509v3_context(Some(&ca_cert_x590), None))?;
     cert_builder.append_extension(subject_alt_name)?;
 
--- a/bpfd/Cargo.toml
+++ b/bpfd/Cargo.toml
@@ -34,7 +34,7 @@ oci-distribution = {version = "0.9", default-features = false, features = ["rust
 serde_json = "1.0"
 tar = "0.4"
 flate2 = "1.0"
-openssl = { version = "0.10.47", features = ["vendored"] }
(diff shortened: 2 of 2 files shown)


astoycos:
Fixes #398

bpfd/src/certs.rs
```diff
@@ -262,6 +262,7 @@ async fn generate_cert(
 
     let subject_alt_name = SubjectAlternativeName::new()
         .dns("localhost,IP:127.0.0.1")
```
astoycos, line 264: @Billy99 Does this need to be 
```suggestion
        .dns("IP:127.0.0.1")
```

Now?

Billy99: If we change it, I think it might need to be:
```suggestion
        .dns("localhost,IP:127.0.0.1")
        .dns("localhost")
        .dns("127.0.0.1")
        .dns("::1")
```
I think I saw something similar on StackOverflow, which led me to the original fix, but can't find the link now. But I need someone else to tell me how it should be.

astoycos: I'm alright with the minimal change for now

dave-tucker (approved):
LGTM

Issue #447, opened 2023-04-28
bpfctl is retrying failed operations over TCP


dave-tucker:
For example, if you attempt to create a program with an invalid UUID (with #446 applied):

```shell
$ sudo ./target/debug/bpfctl load-from-image --id notauuid --image-url quay.io/bpfd-bytecode/xdp_pass:latest xdp  --iface enp8s0 --priority 50
Error = status: Aborted, message: "invalid UUID", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Fri, 28 Apr 2023 11:21:06 GMT", "content-length": "0"} }
Error = status: Aborted, message: "invalid UUID", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Fri, 28 Apr 2023 11:21:06 GMT", "content-length": "0"} }
```

You can see you first get an Aborted status from the Unix socket, and then another one from the TCP socket.

PR #450, opened 2023-04-28
bpfctl is retrying failed operations over TCP

Billy99:
With the addition of Unix Sockets, there was a missing return statement when an error was encountered when trying over socket.

Resolves: #447


dave-tucker (approved):
LGTM

PR #452, opened 2023-04-28
Finish implementing support for TC proceed-on

anfredette:
Fixes  issue #299.

A key point to understand about this implementation the following:

Valid TC return values range from -1 to 8.  Since -1 is not a valid shift value, 1 is added to the value to determine the bit to set in the bitmask and, correspondingly, The TC dispatcher adds 1 to the return value from the BPF program before it compares it to the configured bit mask.

I also tweaked the display for `bpfctl list` a little.

### DIFF
--- a/bpfctl/src/main.rs
+++ b/bpfctl/src/main.rs
@@ -134,7 +134,9 @@ enum LoadCommands {
         priority: i32,
         /// Optional: Proceed to call other programs in chain on this exit code.
         /// Multiple values supported by repeating the parameter.
-        /// Not yet supported for TC programs. May have unintended behavior if used.
+        /// Possible values: [unspec, ok, reclassify, shot, pipe, stolen, queued,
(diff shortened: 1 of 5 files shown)


dave-tucker (approved):
LGTM

anfredette:
This change seems to be causing problems for the kubernetes-integration-tests for TC, so I'll have to look into that.

astoycos:
@anfredette https://github.com/bpfd-dev/bpfd/blob/main/bpfd-operator/controllers/bpfd-agent/tc-program.go#L93 Needs to be fixed as well

astoycos:
(i.e why I think the k8s int tests are failing )

anfredette:
> (i.e why I think the k8s int tests are failing )

Thanks, @astoycos.  That looks like a likely suspect.  I pushed an update.

anfredette:
That did it.

astoycos (approved):
LGTM

Issue #416, opened 2023-04-20
Get code coverage from unit tests


dave-tucker:
We should get code coverage from our unit tests to see where our gaps are.

Issue #356, opened 2023-03-09
Stop including our Go-work files in container builds, possibly stop tracking in GIT 


astoycos:
In order to avoid pushing build which have incorrect versions of certain components we should consider removing our go work setup from container builds for the bpfd-agent and bpfd-operator processes.

Currenlty this is hard since we have a mono-repo... i.e if we make a change to the bpfd-go bindings and want to still build those changes successfully (before pushing to git) we need to use either a local `go replace` directive or go workspaces. 

However this should be reevaluated in the future.

PR #432, opened 2023-04-26
Add BpfdOperator Unit test scaffolding and basic unit tests

astoycos:
Works to fix #431

PR #453, opened 2023-04-28
.github: Generate Code Coverage

dave-tucker:
Adds code coverage to the GitHub actions and uploads results to Codecov

Fixes: #416
Fixes: #356


astoycos:
Thanks @dave-tucker All the unit testing should be cleaned up in #432 So if that goes in this should become green

astoycos:
This also fixes https://github.com/bpfd-dev/bpfd/issues/356

astoycos (commented):
@dave-tucker 
```
RUN cd bpfd-operator && CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o bpfd-agent main.go
```
Will now need to be something like 
```
RUN cd bpfd-operator/cmd/bpfd-agent && CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o bpfd-agent main.go
```
If you're copying the entire bpfd src now :)

PR #460, opened 2023-05-01
Allow registry secret namespace to be specified

astoycos:
This allows the user to specify the namespace of the registry secret, previously the secret had to be created in the BPFD namespace.

Fallout from discussion on https://github.com/anfredette/enhancements/pull/1


astoycos:
PTAL @dave-tucker

dave-tucker (approved):
LGTM

PR #462, opened 2023-05-01
.github: Checkout code to upload coverage reports

dave-tucker:


astoycos:
I've used https://github.com/marketplace/actions/debugging-with-ssh for debugging actions with alot of success :)

dave-tucker:
I'm pretty sure it's that the coverage reports go to a `coverage` dir, and that the codecov action wasn't smart enough to tell us the files were missing 😆 but i'll keep that ssh action for next time!

dave-tucker:
Hmm, well maybe not - trying with a checkout of the code in the job that uploads coverage since it might need to see the paths? 🤔

PR #466, opened 2023-05-02
bpfd: Ensure we set O_NOCTTY when opening files

dave-tucker:
Since we're a daemon, we're run without a TTY and we want to ensure that one doesn't accidentally get allocated.
To do this easily, move all fs access to tokio::fs and use read and read_to_string helpers to add the custom_flag

Fixes: #465


bpfd/src/bpf.rs
```diff
@@ -159,23 +158,25 @@ impl<'a> BpfManager<'a> {
...
             .map_pin_path(map_pin_path.clone())
             .load(&program_bytes)?;
 
-        ext_loader
-            .program_mut(&program.data().section_name)
-            .ok_or_else(|| {
-                let _ = fs::remove_dir_all(map_pin_path);
-                BpfdError::SectionNameNotValid(program.data().section_name.clone())
-            })?;
+        match ext_loader.program_mut(&program.data().section_name) {
```
astoycos, line 171: Why the move from `.ok_or_else` here?

bpfd/src/bpf.rs
```diff
@@ -271,24 +273,24 @@ impl<'a> BpfManager<'a> {
...
                 .load(&program_bytes)?;
 
-            let tracepoint: &mut TracePoint = loader
-                .program_mut(&program.data.section_name)
-                .ok_or_else(|| {
-                    let _ = fs::remove_dir_all(map_pin_path);
-                    BpfdError::SectionNameNotValid(program.data.section_name.clone())
-                })?
-                .try_into()?;
+            let tracepoint: &mut TracePoint = match loader.program_mut(&program.data.section_name) {
```
astoycos, line 285: Same question as before

bpfd-api/src/config.rs
```diff
@@ -136,7 +136,7 @@ fn default_unix() -> String {
 }
 
 pub fn config_from_file<P: AsRef<Path>>(path: P) -> Config {
-    if let Ok(contents) = fs::read_to_string(path) {
+    if let Ok(contents) = read_to_string(path) {
```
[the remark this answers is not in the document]
dave-tucker: Oh man you totally got me here. Turns out that we had some skeletons hidden in the closet here 💀 
I've pushed more changes now to fix this but TL;DR:

1. That we have duplicate config.rs in bpfd and bpfd-api where only the
       latter is used
1. That bpfd can be a binary only crate since everything exposed lives in `bpfd_api`
1. That we need to remove fs::read_to_string from bpf-api since bpfctl and bpfd might want to use different versions of read_to_string

So I ended up pushing that all in this PR 😓

bpfd/src/bpf.rs
dave-tucker: Because you can't use async functions in `ok_or_else`

astoycos (approved):
LGTM

Thanks @dave-tucker!

Issue #451, opened 2023-04-28
TLS from bpfctl to bpfd fails with invalid peer certificate: BadEncoding


Billy99:
TLS from bpfctl to bpfd is broken. I think it broke with openssl bump from 0.10.47 to 0.10.52. 

By default, bpfctl is using Unix Sockets, so to reproduce, delete the socket to force it to use TCP over TLS:
```
$ sudo rm /var/lib/bpfd/sock/bpfd.sock

$ RUST_LOG=debug bpfctl list
[2023-04-28T19:20:44Z WARN  bpfd_api::config] No config file provided. Using default
[2023-04-28T19:20:44Z DEBUG bpfctl] Error getting UNIX socket channel. Err: transport error
[2023-04-28T19:20:44Z DEBUG rustls::anchors] add_parsable_certificates processed 1 valid and 0 invalid certs
[2023-04-28T19:20:44Z DEBUG hyper::client::connect::http] connecting to [::1]:50051
[2023-04-28T19:20:44Z DEBUG hyper::client::connect::http] connected to [::1]:50051
[2023-04-28T19:20:44Z DEBUG rustls::client::hs] No cached session for DnsName("localhost")
[2023-04-28T19:20:44Z DEBUG rustls::client::hs] Not resuming any session
[2023-04-28T19:20:44Z DEBUG rustls::client::hs] Using ciphersuite TLS13_AES_256_GCM_SHA384
[2023-04-28T19:20:44Z DEBUG rustls::client::tls13] Not resuming
[2023-04-28T19:20:44Z DEBUG rustls::client::tls13] TLS1.3 encrypted extensions: [Protocols([ProtocolName(6832)]), ServerNameAck]
[2023-04-28T19:20:44Z DEBUG rustls::client::hs] ALPN protocol is Some(b"h2")
[2023-04-28T19:20:44Z DEBUG rustls::client::tls13] Got CertificateRequest CertificateRequestPayloadTLS13 { context: , extensions: [SignatureAlgorithms([ECDSA_NISTP384_SHA384, ECDSA_NISTP256_SHA256, ED25519, RSA_PSS_SHA512, RSA_PSS_SHA384, RSA_PSS_SHA256, RSA_PKCS1_SHA512, RSA_PKCS1_SHA384, RSA_PKCS1_SHA256]), AuthorityNames([DistinguishedName(30123110300e06035504030c07627066642d6361)])] }
[...]

PR #467, opened 2023-05-02
Bump tonic to 0.9.2 and rustls to 0.21.1

Billy99:
The tonic bump to 0.9.2 brought in rustls 0.20.0. With this bump, as well as the bump of OpenSSL to 0.10.52, mTLS between client and server were broken.

There was a bug in rustls (see https://github.com/rustls/rustls/issues/1288) where the certificate created by the rust server could not be parsed by the GO client (but an identical OpenSSL based certificate could be parsed). This was fixed in rustls 0.21.1.

rust-openssl was also updated in this timeframe, which cause an error parsing the rust generated certificate in the rust client (bpfctl). See https://github.com/rustls/rustls/issues/1292. This is fixed by this PR by changing how the certicate was created. The `.dns()` line below was causing the problem. A partial fix was implemented in #448, which fixed the GO clients, but the rust clients were still broken.
```
    let subject_alt_name = SubjectAlternativeName::new()
        .dns("localhost, IP:127.0.0.1")
        .build(&cert_builder.x509v3_context(Some(&ca_cert_x590), None))?;
```
As part of this fix, fixed up the `scripts/certificate.sh` file to generate the certificates properly using OpenSSL. This helped a lot in debugging this issue. Also rearranged the `bpfd/src/cert.rs` file to match the certificate created in the script so they lined up side-by-side.

Resolves: #398
Resolves: #411
Resolves: #424
Resolves: #451

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -293,6 +293,7 @@ dependencies = [
  "nix",
  "oci-distribution",
  "openssl",
+ "rustls 0.21.1",
  "serde",
  "serde_json",
  "sha2",
@@ -1036,9 +1037,9 @@ checksum = "1788965e61b367cd03a62950836d5cd41560c3577d90e40e0819373194d1661c"
 dependencies = [
  "http",
  "hyper",
- "rustls",
+ "rustls 0.20.8",
  "tokio",
- "tokio-rustls",
+ "tokio-rustls 0.23.4",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


astoycos:
I'll make an issue to bump the codedev patch target back to `auto` once we've stabilized a bit more

astoycos (approved):
LGTM 

Thanks @Billy99!!

PR #472, opened 2023-05-03
Graceful shutdown

dave-tucker:
This carries the last iteration of #405 which removes the use of `futures` by using a function.
I'd originally intended that we should use a tokio::sync::watch channel that was populated by a single signal handler - but rustc said nope because `rx.changed()` requires `rx` to have a static lifetime 😕 anyway, maybe we can fix that in future 😆 

While working on this, I took the opportunity to refactor the command handling a little:

1. Each `Command` that has > 1 arg has a `CommandArgs` struct, which tidies up the code a lot
1. Command processing has been moved from `serve` to `BpfManager::process_commands` as this is both easier to read, and easier to test
1. Since the processing of these commands has also got rather large, they've been factored out into their own functions `load_xdp_command`

### DIFF
--- a/bpfd/src/bpf.rs
+++ b/bpfd/src/bpf.rs
@@ -9,36 +9,42 @@ use aya::{
     BpfLoader,
 };
 use bpfd_api::{config::Config, util::directories::*, ProgramType};
-use log::debug;
-use tokio::fs;
+use log::{debug, info};
+use tokio::{fs, select, sync::mpsc};
 use uuid::Uuid;
 
 use crate::{
     command::{
-        Direction,
+        self, Command, Direction,
         Direction::{Egress, Ingress},
-        Program, ProgramInfo,
+        LoadTCArgs, LoadTracepointArgs, LoadXDPArgs, Program, ProgramData, ProgramInfo, TcProgram,
+        TcProgramInfo, TracepointProgram, TracepointProgramInfo, UnloadArgs, XdpProgram,
+        XdpProgramInfo,
     },
     errors::BpfdError,
     multiprog::{Dispatcher, DispatcherId, DispatcherInfo, TcDispatcher, XdpDispatcher},
     oci_utils::image_manager::get_bytecode_from_image_store,
-    utils::read,
+    serve::shutdown_handler,
+    utils::{get_ifindex, read, set_dir_permissions},
 };
 
 const SUPERUSER: &str = "bpfctl";
+const MAPS_MODE: u32 = 0o0660;
 
-pub(crate) struct BpfManager<'a> {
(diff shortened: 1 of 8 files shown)


bpfd/src/bpf.rs
```diff
@@ -553,4 +561,163 @@ impl<'a> BpfManager<'a> {
         }
         results
     }
+
+    pub(crate) async fn process_commands(&mut self) {
+        loop {
+            // Start receiving messages
+            select! {
+                biased;
```
astoycos, line 569: What is this doing?

bpfd/src/bpf.rs
```diff
@@ -553,4 +561,163 @@ impl<'a> BpfManager<'a> {
...
+                            let _ = responder.send(progs);
+                        }
+                    }
+                }
+            }
+            info!("Stopping processing commands");
+        }
+    }
+
+    async fn load_xdp_command<'a>(&mut self, args: LoadXDPArgs) -> anyhow::Result<()> {
```
astoycos, line 592: Do you need the lifetime identifier here?

bpfd/src/bpf.rs
```diff
@@ -553,4 +561,163 @@ impl<'a> BpfManager<'a> {
...
+                        attached: false,
+                    },
+                    proceed_on: args.proceed_on,
+                    if_name: args.iface,
+                },
+            }));
+
+            match prog_result {
+                Ok(prog) => self.add_program(prog, args.id).await,
+                Err(e) => Err(e),
```
astoycos, line 620:
```suggestion
            let prog = Program::Xdp(XdpProgram {
                data: prog_data.clone(),
                info: XdpProgramInfo {
                    if_index,
                    current_position: None,
                    metadata: command::Metadata {
                        priority: args.priority,
                        // This could have been overridden by image tags
                        name: prog_data.section_name,
                        attached: false,
                    },
                    proceed_on: args.proceed_on,
                    if_name: args.iface,
                },
            });
            
            self.add_program(prog, args.id).await,
```

This should do it since prog_result can never be ERR as it is now, same for other functions.

astoycos (commented):
Some comments and a suggestion, it looks good though @dave-tucker Thanks!

bpfd/src/bpf.rs
dave-tucker: Ensures that the shutdown branch gets priority.

bpfd/src/bpf.rs
dave-tucker: Good catch!

astoycos (approved):
LGTM

PR #473, opened 2023-05-03
Bpfctl refactoring

### LINKED ISSUE #412: Decompose bpfctl main.go
This main is getting quite large. It might be helpful for future readers and for maintenance to compose some of the logic into discreet functions.

See `bpfctl/src/main.rs`

ykaliuta:
Factor out functionality to own functions

Fixes #412

### DIFF
--- a/bpfctl/src/main.rs
+++ b/bpfctl/src/main.rs
@@ -343,7 +343,7 @@ impl Commands {
                         .as_str()
                         .try_into()
                         .expect("invalid image pull policy");
-                    match l.registry_auth.clone() {
+                    let (username, password) = match l.registry_auth.clone() {
                         Some(a) => {
                             let auth_raw = general_purpose::STANDARD_NO_PAD.decode(a)?;
 
@@ -351,20 +351,17 @@ impl Commands {
 
                             let (username, password) = auth_string.split(':').next_tuple().unwrap();
 
-                            Some(load_request_common::Location::Image(BytecodeImage {
-                                url: l.image_url.clone(),
-                                image_pull_policy: image_pull_policy as i32,
-                                username: username.to_owned(),
-                                password: password.to_owned(),
-                            }))
+                            (username.to_owned(), password.to_owned())
                         }
-                        None => Some(load_request_common::Location::Image(BytecodeImage {
-                            url: l.image_url.clone(),
-                            image_pull_policy: image_pull_policy as i32,
-                            username: "".to_owned(),
-                            password: "".to_owned(),
-                        })),
-                    }
+                        None => ("".to_owned(), "".to_owned()),
+                    };
+
+                    Some(load_request_common::Location::Image(BytecodeImage {
+                        url: l.image_url.clone(),
+                        image_pull_policy: image_pull_policy as i32,
+                        username,
+                        password,
+                    }))
                 };
             }
             _ => bail!("Unknown command"),
--- a/bpfctl/src/main.rs
+++ b/bpfctl/src/main.rs
@@ -158,6 +158,99 @@ struct GlobalArg {
     value: Vec<u8>,
 }
 
+struct ProgTable(Table);
+
+impl ProgTable {
+    fn new() -> Self {
+        let mut table = Table::new();
+
+        table.load_preset(comfy_table::presets::NOTHING);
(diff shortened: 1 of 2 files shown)


astoycos (commented):
Pretty much there!! I like what you did with the ListTable and loadCommands implementation but I think we could could eliminate that extra trait to make things simplier!

astoycos:
also `cargo fmt` should fix the immediate failure I'm seeing in CI

bpfctl/src/main.rs
```diff
@@ -248,264 +396,155 @@ async fn main() -> anyhow::Result<()> {
...
-                ),
-            };
-
-            let mut global_data: HashMap<String, Vec<u8>> = HashMap::new();
-
-            if let Some(global) = &l.global {
-                for g in global.iter() {
-                    global_data.insert(g.name.to_string(), g.value.clone());
-                }
+trait LoadArgsOps {
```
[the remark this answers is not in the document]
ykaliuta: Thanks! Next try :)

bpfctl/src/main.rs
```diff
@@ -158,6 +160,235 @@ struct GlobalArg {
...
+                        url: l.image_url.clone(),
+                        image_pull_policy: image_pull_policy as i32,
+                        username,
+                        password,
+                    }))
+                };
+            }
+            _ => bail!("Unknown command"),
+        };
+
```
astoycos, line 367:
```suggestion
        // Common fields
        let common = match self {
            Commands::LoadFromFile(l) => {
                LoadRequestCommon{
                    id: &l.id;
                    section_name: &l.section_name;
                    global: &l.global;
                    command: &l.command;
                    location: Some(load_request_common::Location::File(l.path.clone()));
                }
            }
            Commands::LoadFromImage(l) => {
                LoadRequestCommon{
                    id: &l.id;
                    section_name: &l.section_name;
                    global: &l.global;
                    command: &l.command;
                    location: {
                        let image_pull_policy: ImagePullPolicy = l
                            .image_pull_policy
                            .as_str()
                            .try_into()
                            .expect("invalid image pull policy");
                        let (username, password) = match l.registry_auth.clone() {
                            Some(a) => {
                                let auth_raw = general_purpose::STANDARD_NO_PAD.decode(a)?;
                                let auth_string = String::from_utf8(auth_raw)?;
                                let (username, password) = auth_string.split(':').next_tuple().unwrap();

                                (username.to_owned(), password.to_owned())
                            }
                            None => ("".to_owned(), "".to_owned()),
                        };

                        Some(load_request_common::Location::Image(BytecodeImage {
                            url: l.image_url.clone(),
                            image_pull_policy: image_pull_policy as i32,
                            username,
                            password,
                        }))
                    };
                }
            }
            _ => bail!("Unknown command"),
        };

```

bpfctl/src/main.rs
```diff
@@ -158,6 +160,235 @@ struct GlobalArg {
...
+        }
+
+        Ok(Some(LoadRequestCommon {
+            id: id.clone(),
+            location,
+            section_name: section_name.to_string(),
+            program_type: command.prog_type() as i32,
+            global_data,
+        }))
+    }
```
astoycos, line 381:
```suggestion
        if let Some(global) = global {
            for g in global.iter() {
                common.global_data.insert(g.name.to_string(), g.value.clone());
            }
        }

        Ok(Some(common))
    }
```

astoycos (commented):
Looking good, thanks for all the hard work just one small var nit :) 

Also if you could make an follow up issue for adding unit tests to all this functionality that would be really great, thanks!

bpfctl/src/main.rs
ykaliuta: Hmmm, I would not say I'm fan of the proposal (if I understand it correctly). I pushed an updated version. There is now code duplication for common fields (.to_string() ...) and still fetching `global` from different types require a variable.

anfredette:
> I just want @anfredette To take a final pass since he was just in this code

Looking now...

anfredette (commented):
LGTM.  Thanks for doing this @ykaliuta!

I'll let @astoycos approve it after the signing is done.

ykaliuta:
Thanks a lot for the review. Hopefully the final version

astoycos (approved):
LGTM

PR #477, opened 2023-05-04
Improve endpoint configuration

### LINKED ISSUE #418: Allow disabling endpoints
Since #410 was merged, bpfd is now listening on both a Unix and a TCP socket and the only configuration possible is where they actually listen on. It'd be great if we could have more control over what endpoints bpfd listens on. There are a few ways we could achieve this, off the top of my head these are some of them:

#### Make the default behavior to not listen on the TCP socket
This would be the simplest solution, make the configuration for TCP socket required for bpfd to listen on it. We could apply the same logic to the unix socket, but since bpfd needs to listen on at least one endpoint, I think it makes sense to always listen on the default unix socket.

#### Add an `enabled` field for each endpoint
We could add a `tcp_enabled` and a `unix_enabled` field to the `grpc.endpoint` configuration, they would default to `true` and setting them to `false` would cause bpfd to not open the specified endpoint. Though this achieves the objective of being able to disable bpfd from listening on certain types of endpoints, the naming is not the most friendly in my opinion.

Another option in this same idea is to split the `grpc.endpoint` section into `grpc.endpoint.unix` and `grpc.endpoint.tcp`, each of them having its own fields, looking something like this:
```toml
[grpc.endpoint.tcp]
address = "::1"
port = 50051
enabled = false

[grpc.endpoint.unix]
path = "/run/bpfd/bpfd.sock"
enabled = true
```

#### Unify endpoint configuration and allow more of them
Building a little on top of the previous example, we could also unify endpoint configuration and turn it from a simple table to an array of tables, looking something like this:
```toml
[[grpc.endpoint]]
type = "tcp"
enabled = true
address = "::1"
port = 50051

[[grpc.endpoint]]
type = "tcp"
enabled = false
address = "127.0.0.1"
port = 50051

[[grpc.endpoint]]
type = "unix"
enabled = true
address = "/run/bpfd/bpfd.sock"
port = 1  # This could issue a warning and ignore it when creating the server.
```
As stated on a previous point, not adding any endpoints in this example would cause bpfd to only listen on the default unix socket.

I admit this last option could be a bit overkill, but I still wanted to put it out there. There also might be a way to not add the `type` field and instead have serde desarialize an enum based on whether we have `address` and `port` or `path` in the element, but I'm not sure that's the case and don't want to delve deeper into implementation details right now.

P.S: Sorry for the wall of text, I got a little carried away.

Molter73:
Endpoint configuration now allows multiple unix and TCP sockets to be used for use as both server in bpfd and client in bpfctl. Individual endpoints can also be enabled and disabled via configuration. For client connections, priority is determined by the order the endpoints are listed in the TOML configuration file.

Fixes #418 

TODO:
- [x] Change go examples to use the new configuration.
- [x] Revisit config unit tests.
- [x] Change existing documentation.
- [x] Change existing bpfd.toml.

### DIFF
--- a/bpfctl/src/main.rs
+++ b/bpfctl/src/main.rs
@@ -6,7 +6,7 @@ use std::{collections::HashMap, fs, net::SocketAddr, str};
 use anyhow::{bail, Context};
 use base64::{engine::general_purpose, Engine as _};
 use bpfd_api::{
-    config::Config,
+    config::{self, Config},
     util::directories::*,
     v1::{
         list_response, load_request, load_request_common, loader_client::LoaderClient,
@@ -19,7 +19,7 @@ use clap::{Args, Parser, Subcommand};
 use comfy_table::Table;
 use hex::FromHex;
 use itertools::Itertools;
-use log::{debug, info, warn};
+use log::{info, warn};
 use tokio::net::UnixStream;
 use tonic::transport::{Certificate, Channel, ClientTlsConfig, Endpoint, Identity, Uri};
 use tower::service_fn;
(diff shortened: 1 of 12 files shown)


Molter73:
This PR will have conflicts with #473, we can hold on merging this until the bpfctl refactor is done, but I think this PR is far along for an initial review.

Billy99 (commented):
Thanks for this PR. Looks good. Just need to update some of the associated documentation.

Billy99 (approved):
LGTM

Issue #429, opened 2023-04-25
Allow for pre-pulling of bytecode images


astoycos:
Now that we have support for Image Pull Policy we should allow users to pre-pull images (before a program load request) 

The way I see it is we can 

1. Keep core privileged functionality limited to bpfd and add a `pull_bytecode` command to the API
2. Allow bpfctl to pull and deposit bytecode images into the bpfd image content store (this would unfortunately be a privileged operation since bpfctl would need to access /var/lib/bpfd) 

I'm leaning towards 1 but would like to hear other's opinions as well. 

cc @dave-tucker

dave-tucker:
1. is fine and would result in a `bpfctl pull` command anyway.

PR #481, opened 2023-05-04
Update documentation around examples

Billy99:
Update documentation around examples and fix some issues encountered in code while testing out the examples.


Billy99:
@astoycos In example-bpf.md, there are two places that reference pushing images to remote repository, one has a blurb about it being required because bpfd doesn't support local builds. Let me know how this should be reworded with regards to your latest image changes. Not sure it will show up in diff, so search for `push` in the file.

astoycos:
@Billy99 It's basically still the same until we get #429 fixed

Then you could load a bytecode image with bpfctl and set imagePullPolicy to "Never"

docs/example-bpf.md
```diff
@@ -300,6 +323,21 @@ and
...
+      --build-arg SECTION_NAME=tracepoint_kill_recorder \
+      --build-arg PROGRAM_TYPE=tracepoint \
+      --build-arg BYTECODE_FILENAME=bpf_bpfel.o \
+      --build-arg KERNEL_COMPILE_VER=$(uname -r) \
+      -f ../../packaging/container-deployment/Containerfile.bytecode . -t quay.io/$USER/go-tracepoint-counter-bytecode:latest
+```
+
 `bpfd` currently only supports pulling a remote container image, so push the image to a remote
 repository.
 For example:
```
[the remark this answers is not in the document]
Billy99: Done, thanks

docs/example-bpf.md
```diff
@@ -34,13 +39,22 @@ This section describes running bpfd and the example bpf programs on a local host
 When running bpfd, it can be run as a process or run as a systemd service.
 Examples run the same, independent of how bpfd is deployed, other than requiring `sudo` or not.
 
+> **_NOTE:_** When running bpfd as a systemd service, bpfd can only access files owned by the `bpfd`
```
astoycos, line 42: Wait is only a problem when running with Systemd? I feel like we'll have this issue when just running bpfd locally on a system without `sudo`

docs/example-bpf.md
```diff
@@ -116,25 +134,33 @@ The most basic example, just use `sudo` to start the `go-xdp-counter` program.
...
 The output should show the count and total bytes of packets as they pass through the
 interface as shown below:
 
 ```console
     sudo ./go-xdp-counter -iface ens3
-    2022/12/02 15:59:34 Reading /etc/bpfd/gocounter.toml ...
-    2022/12/02 15:59:34 Read /etc/bpfd/gocounter.toml failed: err open /etc/bpfd/gocounter.toml: no such file or directory
-    2022/12/02 15:59:34 Using Input: Interface=ens3 Priority=50 Source=bpf_bpfel.o
+    2022/12/02 15:59:34 Reading /etc/bpfd/bpfd.toml ...
+    2022/12/02 15:59:34 Read /etc/bpfd/bpfd.toml failed: err open /etc/bpfd/bpfd.toml: no such file or directory
```
astoycos, line 161: Do we wan't to remove printing this error? It always makes me think the examples are broken and I'm sure a naive user will feel the same

docs/example-bpf.md
```diff
@@ -1,30 +1,35 @@
 # Example BPF Programs
 
-Example applications that use the `bpfd-go` bindings can be found in
-[examples/go-xdp-counter/](https://github.com/bpfd-dev/bpfd/tree/main/examples/go-xdp-counter) and
-[examples/go-tc-counter/](https://github.com/bpfd-dev/bpfd/tree/main/examples/go-tc-counter).
+Example applications that use the `bpfd-go` bindings can be found in the
```
astoycos, line 3: So a high level comment... I would break these up into different docs under a docs/examples dir since this one is so long. I know it renders with a table of contents on our website but most users are gonna go "I just want to run this thing FAST how do I do it" So we should make that easy to find.... 

Additionally I'd probably have 3 docs 1: Building From Source 2: Running Locally 3: Running on K8s 
+
maybe even a "Quick Start" that just shows the explicit commands need to run it using prebuilt images and `sudo` all together.

WDYT?

examples/go-tracepoint-counter/kubernetes-deployment/go-tracepoint-counter.yaml
```diff
@@ -60,7 +60,7 @@ spec:
         effect: NoSchedule
       containers:
       - name: go-tracepoint-counter
-        image: quay.io/bpfd-userspace/go-tracepoint-counter:latest
+        image: quay.io/bpfd-userspace/go-tracepoint-counter:test
```
astoycos, line 63: Yep I assume this will get dropped

examples/go-tracepoint-counter/main.go
```diff
@@ -136,7 +148,7 @@ func loadProgram(paramData *configMgmt.ParameterData) (func(string), error) {
 	c := gobpfd.NewLoaderClient(conn)
 	loadRequestCommon := &gobpfd.LoadRequestCommon{
 		Location:    paramData.BytecodeSource.Location,
-		SectionName: "stats",
+		SectionName: "tracepoint_kill_recorder",
```
astoycos, line 151: Wow this _shouldn't_ have been working since passed SectionName didn't match the one on the image... Hrm 

https://github.com/bpfd-dev/bpfd/blob/main/bpfd/src/command.rs#L244

We definitely need and integration test for this :) 

I wonder how it was

examples/go-tracepoint-counter/main.go
```diff
@@ -102,21 +108,27 @@ func main() {
 	// retrieve and report on the number of kill -SIGUSR1 calls
 	index := uint32(0)
 	ticker := time.NewTicker(1 * time.Second)
-	defer ticker.Stop()
-	for range ticker.C {
-		var stats []Stats
-		var totalCalls uint64
+	go func() {
```
astoycos, line 111: What was the reason for putting this in a go-routine?

examples/pkg/config-mgmt/param.go
```diff
@@ -33,44 +33,64 @@ const (
...
 )
 
+type ProgType int
 const (
-	ProgTypeXdp = iota
+	ProgTypeXdp ProgType = iota
 	ProgTypeTc
 	ProgTypeTracepoint
 )
+func (s ProgType) String() string {
```
astoycos, line 42: Yuck lets try to standardize all this type handling in a follow up

astoycos (commented):
Pretty much there just some general questions and comments

docs/example-bpf.md
Billy99: I guess I didn't know we could do that. I also run as a systemd service (or test with `sudo ./target/debug/bpfd`). Don't you need sudo in the latter for CAP_BPF?

docs/example-bpf.md
Billy99: Or rewrite the log to something like:
`2022/12/02 15:59:34 Unable to read /etc/bpfd/bpfd.toml, using default values.`
What do you think?

examples/go-tracepoint-counter/kubernetes-deployment/go-tracepoint-counter.yaml
Billy99: Yes, this commit will get dropped.

examples/go-tracepoint-counter/main.go
Billy99: I think it was only really tested on K8s. Don't think it ever worked when run locally.

examples/go-tracepoint-counter/main.go
Billy99: The tracepoint programs weren't being unloaded. The defer statements were being skipped on a <CTRL-C>. I just ported over the XDP and TC pattern.

docs/example-bpf.md
Billy99: Yes, just tried it, you need sudo:
```
$ ./target/debug/bpfd 
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: EPERM', bpfd/src/main.rs:63:79
```

docs/example-bpf.md
Billy99: So I'll take a look. I think the "Quick Start" is the tutorial.md. I am working on a PR to update that file. That may still need to be simpler. The examples are for the next level, i.e. writing your own program. Not saying they don't need to be simplified, but just a point to consider.

docs/example-bpf.md
astoycos: Yeah that's fine too!

docs/example-bpf.md
astoycos: Ahh nw I've never actually run bpfd using systemd :( 

If that's not something we support anyways you can ignore my comment

examples/go-tracepoint-counter/main.go
astoycos: Ok I'm alright with that... I feel like that pattern also _could_ result in a break when `ctrl-c`ed at the right time :)

examples/go-tracepoint-counter/main.go
astoycos: Ahhh duh you're right, thanks!

docs/example-bpf.md
astoycos: Yeah but even in the existing quick start we only show how to load a program.... I think the real value add is much better shown when loading a program that's actually doing something IMO 

I hear you though we could leave these out of the quick start for now.

astoycos (approved):
LGTM

PR #476, opened 2023-05-03
Update bpfd README

anfredette:
Better describe what bpfd is and the problems it solves.

Addresses issue #474.

PR #483, opened 2023-05-04
Re-organize website and add more bpfd-operator docs

astoycos:
Works on https://github.com/bpfd-dev/bpfd/pull/383


astoycos:
I would at the very least like #481 and #476 To go in before this so they don't have any rebase confusions.

anfredette:
Thanks for doing this @astoycos.  I think it’s time for more organization.  However, as I was looking through the changes, I was thinking of another alternative that’s probably closer to what we had originally.  What do you think about the following:

- Home
  - Pull in the info from the README. (It would be nice to be able to document it one time, and then #include it both places, but I’d need to look into how to do this with md files).
- Community
  - The Project Governance docs from Dave
- Users
  (Info needed by application developers who want to use bpfd for their application)
  - K8S Quick Start
  - API Reference
  - EBPF Bytecode Image Specifications
  - Examples
- Developers
  (Info needed by people who want to write bpfd code)
  - Developing the bpfd-operator
  - TODO: Other bpfd design docs
  - Info on contributing

astoycos:
@Billy99 Thanks for the help here some nits

- Move the bpfd-operator Quick Start to the "Getting Started" section
- Link the "getting Started" section to the Home page
- Move "Community" to the last in the line so Home, Getting Started, Developers, Community
- Maybe change Developers to Developing? 
- Add a Note to the home page that we're listed as an ebpf.io "emerging project"

astoycos:
Hrm now for some reason the landing page isn't "home" it's 
image

anfredette (approved):
This is a big improvement.  However, I still think what is in the "Developer Guide" should be split into two separate sections.  1. Info needed by application developers who want to use bpfd for their applications (I called this "Users" in my earlier comment, but there could be a better name), and 2. Info useful for people who want to develop bpfd code ("Developers" in my earlier comment.  So, I'm okay with merging this, but I recommend we follow up with that change.

PR #486, opened 2023-05-05
Updates to API files

anfredette:
I noticed recently that `bpfd-api/src/bpfd.v1.rs` was getting changed when I ran `cargo xtask build-proto`.  I'm not sure what caused the change, but it may have been the [tonic bump](https://github.com/bpfd-dev/bpfd/pull/467) because they seem to be Tonic-related.  In any case, when the change happened, the new file didn't get checked in.

Just to check, I also updated Rust, did a `dnf update`, and tried again. The `dnf update` pulled in a new version of `protoc`.  This resulted in updates to the Go API files as well, but the only change is to a comment listing the new version of `protoc`.


astoycos:
So your local protoc was newer, specifically here it was bumped from `v3.19.4` to `v3.19.6` which is where these changes came from

Billy99 (approved):

PR #487, opened 2023-05-05
Remove Ok from TCProceedOn defaults

anfredette:
This makes our default TC dispatcher behavior the same as standard TC:

`TC_ACT_OK`, will terminate the packet processing pipeline and allows the packet to proceed
`TC_ACT_PIPE`, will iterate to the next action, if available


astoycos (approved):
LGTM

PR #492, opened 2023-05-10
Fix small bug in statematch

astoycos:
Fix a small bug where an AND should have been an OR :/

This broke the agent's ability to apply updates properly.


Billy99 (approved):
LGTM

PR #499, opened 2023-05-16
Fixup OCP deployment + Operator Bundle

astoycos:
This PR does a few things 

- Remove stale references to bpfProgramConfig APIs
- Adds role perms for the bpfd-agent to update finalizers on all *Program objects so we don't run into the following on OCP 
```
Program object: bpfprograms.bpfd.io \"xdp-pass-all-nodes-ip-10-0-132-235.ec2.internal\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on
```
- Fixup our operator bundle
- Fixup the deletion case, allowing us to safetly delete the bpfd configmap so that the bpfd-daemon is destroyed before operator cleanup once the bpfd-configmap is deleted.
- Remove bpfd.namespace config from the bpfd-configmap.  Instead we just deploy the bpfd-daemon to whatever namespace the configmap is in


bpfd-operator/bundle/manifests/bpfd.io_tcprograms.yaml
```diff
@@ -0,0 +1,315 @@
...
+            - sectionname
+            type: object
+          status:
+            description: TcProgramStatus defines the observed state of TcProgram
+            properties:
+              conditions:
+                description: Conditions houses the global cluster state for the TcProgram.
+                  The explicit condition types are defined internally.
+                items:
+                  description: "Condition contains details for one aspect of the current
```
Billy99, line 233: When is this printed pretty?

bpfd-operator/config/crd/patches/webhook_in_xdpprograms copy.yaml
```diff
@@ -0,0 +1,16 @@
+# The following patch enables a conversion webhook for the CRD
```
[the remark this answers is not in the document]
Billy99: Actually, I see `#- patches/webhook_in_bpfprograms.yaml` in `bpfd-operator/config/crd/kustomization.yaml`, but don't see any of the program specific (`webhook_in_tcprograms.yaml`, ` webhook_in_tracepointprograms.yaml` and `'webhook_in_xdpprograms copy.yaml'`) being used at all.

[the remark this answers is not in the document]
astoycos: So they're not being used today, but I imagine we may need to setup conversion webhooks sometime in the future which is why I didn't want to delete those manifests 

I'll add a comment to make that a bit more clear.

bpfd-operator/bundle/manifests/bpfd.io_tcprograms.yaml
astoycos: What do you mean? This is just part of the bundle that will be deployed whenever we're using OLM 

This CRD is just copied from the autogenerated one.

astoycos: so the only way you'd see it at the EOD is with a `kubectl get crd ...`

Billy99: Had `\n` in the descriptions. When I dumped:

```
$ kubectl describe crd bpfprograms.bpfd.io
Name:         bpfprograms.bpfd.io
Namespace:    
Labels:       <none>
Annotations:  controller-gen.kubebuilder.io/version: v0.11.3
:
                  Description:  Condition contains details for one aspect of the current state of this API Resource. --- This struct is intended for direct use as an array at the field path .status.conditions.  For example, 
 type FooStatus struct{ // Represents the observations of a foo's current state. // Known .status.conditions.type are: "Available", "Progressing", and "Degraded" // +patchMergeKey=type // +patchStrategy=merge // +listType=map // +listMapKey=type Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"` 
 // other fields }
:
```
The `\n` didn't make a difference.

astoycos: It is ugly let me try and fix this

astoycos: So actually I can't @Billy99 This is from https://github.com/kubernetes/apimachinery/blob/v0.26.2/pkg/apis/meta/v1/types.go#L1464 which is part of the core API

Billy99: Just ignore the comment then. Just thought it looked weird, but not a big deal.

Billy99 (approved):
LGTM

Issue #500, opened 2023-05-17
global data is not getting set


anfredette:
I tested both TC and XDP.  Both are broken.  I assume that trace points are broken as well.

PR #502, opened 2023-05-18
Fix setting of global data in extension programs

anfredette:
With changes in pr #372, the BpfLoader was getting re-initialized after the global values were set.

Fixes: #500


astoycos (approved):
Good catch, thanks! 

LGTM

Issue #501, opened 2023-05-18
TC dispatcher gets detached when bpfd exits and doesn't get re-attached after bpfd is restarted


anfredette:
Also, after a restart, the TC dispatcher doesn't get attached when new bpf programs are added.

PR #504, opened 2023-05-18
Change the way we save TC links

anfredette:
The behavior for TC links changed when detaching links on drop was
fixed in Aya. As a result, TC links started getting dropped when bpfd
would exit.

With this change, we don't save the link in TcDispatcher anymore.
Instead, we mem::forget the link and then recreate it when it needs
to be detached.  As a result, the link doesn't get dropped when bpfd
exits. This also fixes the reload issue.

Fixes: #501


bpfd/src/multiprog/tc.rs
```diff
@@ -144,15 +141,12 @@ impl TcDispatcher {
                 ..Default::default()
             },
         )?;
+
         let link = new_dispatcher.take_link(link_id)?;
         self.handle = Some(link.handle());
-        self.link = Some(link);
+        mem::forget(link);
```
astoycos, line 147: Just for my understanding... What's the difference between link and `link.handle` here? Would the link not get automatically dropped because we're keeping that reference to `link.handle`

bpfd/src/multiprog/tc.rs
```diff
@@ -294,31 +288,21 @@ impl TcDispatcher {
...
-        }
+        if let Some(old_handle) = self.handle {
+            let attach_type = match self.direction {
+                Direction::Ingress => TcAttachType::Ingress,
+                Direction::Egress => TcAttachType::Egress,
+            };
+            if let Ok(old_link) =
+                SchedClassifierLink::attached(&self.if_name, attach_type, self.priority, old_handle)
+            {
+                old_link.detach()?;
```
astoycos, line 299: Does the `old_link` var get created from `if let Ok(as)`?

bpfd/src/multiprog/tc.rs
anfredette: `link` is a `SchedClassifierLink` object, while `link.handle()` is just the value of the handle (a `u32`) of the tc dispatcher.  For example, in the following:
```
nfvsdn-04:~$ tc filter show dev mynet1 ingress
filter protocol all pref 50 bpf chain 0 
filter protocol all pref 50 bpf chain 0 handle 0x1 dispatcher direct-action not_in_hw id 681 tag 69ab5e4b570e07c0 
```
The handle is 0x1.  The way I'm attaching the dispatcher, the handle gets automatically assigned by the system, and I need to get it and save it so I can re-instantiate the link later when I need to detach it when it's being replaced by a new dispatcher.  Getting the handle doesn't cause the link to get dropped or prevent the link from being dropped.  When I used to save the link with `self.link = Some(link);`, that prevented the link from getting dropped when link went out of scope.  It also prevented it from getting dropped when bpfd exited because the destructor for TC links (and others) wasn't working right.  However, that got fixed, so then the link started getting dropped when bpfd exited.

bpfd/src/multiprog/tc.rs
anfredette: I'm not sure what you mean by `if let Ok(as)`.  `old_link` gets re-instantiated with the `SchedClassifierLink::attached()`.  `attached()` is effectively a `new()` call, but Alessandro thought calling it `new()` would be confusing.

astoycos: Ahh cool makes sense I was confused because my cod diff was rendering incorrectly

astoycos:
I'm mucking with release stuff don't merge yet

bpfd/src/multiprog/tc.rs
anfredette: > When you do this.... Has the link stayed up and running on the system in the meantime because it's pinned and you just need to "re-instantiate a reference to it in memory in order to actually do things with it?

Yes.

anfredette:
A. This needs to go in before we do the release, and
B. Please hold off on merging for a little bit.  I'm currently handling an error path for the case when the old link was actually detached.  This shouldn't happen now unless someone explicitly does it, but if it does, the TC dispatcher never gets attached again.

astoycos:
> A. This needs to go in before we do the release, and

Yep I'll make sure this is in before we release, just testing out all the tooling 

> B. Please hold off on merging for a little bit. I'm currently handling an error path for the case when the old link was actually detached. This shouldn't happen now unless someone explicitly does it, but if it does, the TC dispatcher never gets attached again.

I'm actually in a force push cycle to main right now to get everything working so you'll need to rebase before pushing any fixes to ^^ but there shouldn't be any conflicts

anfredette:
> I'm currently handling an error path for the case when the old link was actually detached. This shouldn't happen now unless someone explicitly does it, but if it does, the TC dispatcher never gets attached again.

I'm done, and this should be ready to go.

astoycos (approved):
LGTM

Thanks @anfredette

Issue #514, opened 2023-05-22
Run examples based on version tag


Billy99:
The goal is to be able to run a tagged version of the examples using kustomize:
`TAG=v0.2.0 make deploy-go-xdp-counter`

Kustomize is not setup to do the above (based on a variable). Kustomize is more static. Below are 3 examples of how it could be done, each with pros and cons.

**The Kustomize way (tracepoint):**
Most of the Kustomize examples have 2-3 static deployments, like `prod`, `staging`, `test`. There is a base directory with the default yaml, then each deployment has it's own directory with overrides. This is modeled in the tracepoint:
```
.
├── go-tracepoint-counter-bytecode
│   ├── base
│   │   ├── go-tracepoint-counter-bytecode.yaml
│   │   └── kustomization.yaml
│   ├── v0.2.0
│   │   └── kustomization.yaml
│   ├── v0.2.1
│   │   └── kustomization.yaml
│   ├── v0.2.2
│   │   └── kustomization.yaml
│   ├── v0.2.3
│   │   └── kustomization.yaml
│   ├── v0.3.0
│   │   └── kustomization.yaml
│   └── v0.3.1
│       └── kustomization.yaml
├── go-tracepoint-counter-userspace
│   ├── base
│   │   ├── go-tracepoint-counter-userspace.yaml
│   │   └── kustomization.yaml
│   ├── v0.2.0
│   │   └── kustomization.yaml
│   ├── v0.2.1
│   │   └── kustomization.yaml
│   ├── v0.2.2
│   │   └── kustomization.yaml
│   ├── v0.2.3
│   │   └── kustomization.yaml
│   ├── v0.3.0
│   │   └── kustomization.yaml
│   └── v0.3.1
│       └── kustomization.yaml
```

The `go-tracepoint-counter-bytecode/base/go-tracepoint-counter-bytecode.yaml` and `go-tracepoint-counter-userspace/base/go-tracepoint-counter-userspace.yaml` are the exact same yaml files as today. The `base/kustomization.yaml` files just point to the resource:
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
  - go-tracepoint-counter-bytecode.yaml
```

The versions specific `kustomization.yaml` files are cut and paste of the previous with the version tag changed:
```
$ cat go-tracepoint-counter-bytecode/v0.2.0/kustomize.yaml`
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
[...]

PR #515, opened 2023-05-22
Use Kustomize to deploy GO examples

Billy99:
See issue #514 for details


astoycos (commented):
So a few general comments

- Can we collapse all of the -bytecode/-userspace dirs into a single one with the directory name just the example name?
- kustomization.yaml.overlay should be either just kustomization.yaml or two files kustomization.yaml AND patch.yaml
- I think we can get rid of `base`, `image` parent dirs and just collapse it into "default" with three dirs `go-xdp-example`, `go-tc-example` and `go-tracepoint-example` those three dirs will have the base bytecode and deployment yamls and a basic kustomize yaml.  something like 

```
/default..
   kustomization.yaml
   patch.yaml(has 3 patches one for each programtype default image will come from makefile 
   VAR)
   /go-xdp-counter/ ( bytecode.yaml, deployment.yaml)
   /go-tc-counter/
   /go-tracepoint-counter/
/0.2.0
  pretty much just like you have it but pointing back to ../../default/go-xdp-counter/
```

The defaults can just be derived from Make defines see https://github.com/bpfd-dev/bpfd/blob/main/bpfd-operator/Makefile#L50.  All of the make files should try and use kustomize for consistency, NOT install by straight yamls.  This should simplify the Makefile as well.

Billy99:
> * Can we collapse all of the -bytecode/-userspace dirs into a single one with the directory name just the example name?

Done. Good catch. For some reason I had in my head they needed to be separated. MUCH cleaner with the userspace and bytecode combined.

> * kustomization.yaml.overlay should be either just kustomization.yaml or two files kustomization.yaml AND patch.yaml

Per discussion, this won't work, because the variable `URL_BC` would be replaced inline in the file. So it would work the first time, but not on subsequent calls. Moved the variable in to a patch.yaml.env file, some all the funniness is out of the kustomization.yaml file. Then `sed` call will substitute the variable in patch.yaml.env and pipe the output to patch.yaml, which is used by Kustomize.

> * I think we can get rid of base, image parent dirs and just collapse it into "default" with three dirs go-xdp-example, go-tc-example and go-tracepoint-example those three dirs will have the base bytecode and deployment yamls and a basic kustomize yaml. something like ...

I'm probably wrong, but I don't think this will work. In your example, `default/kustomization.yaml` would have patches for all 3 examples (tc, tracepoint, and xdp). But the user may only want to load one of the examples, not all 3, which would be possible if the kustomization.yaml were a directory above the example deployment yaml files. There is already a kustomization.yaml in the directory with the example deployment yaml files with the `resources` to support the `v0.2.0` directory referring to `bases: ../../base/go-*-counter` 

>  All of the make files should try and use kustomize for consistency, NOT install by straight yamls. This should simplify the Makefile as well.

Done

examples/config/default/go-tc-counter/kustomization.yaml
```diff
@@ -0,0 +1,21 @@
...
+kind: Kustomization
+bases:
+- ../../base/go-tc-counter
+
+# Patch the deployment.yaml to change container image in Daemonset
+# to new tag on the image.
+images:
+- name: quay.io/bpfd-userspace/go-tc-counter
+  newName: quay.io/bpfd-userspace/go-tc-counter
+  newTag: latest
```
astoycos, line 11: For consistency here should we default to the "main" tag?  (Now we automatically push these to :main in gh actions`

examples/Makefile
```diff
@@ -0,0 +1,193 @@
...
+build: fmt ## Build all the userspace example code.
+	go build ./...
+
+.PHONY: generate
+generate: ## Run `go generate` to build the bytecode for each of the examples.
+	go generate ./...
+
+##@ Deployment Variables (not commands)
+TAG: ## Used to set all images to a fixed tag. Example: make deploy TAG=v0.2.0
+IMAGE_TC_BC: ## TC Bytecode image. Example: make deploy-tc IMAGE_TC_BC=quay.io/user1/go-tc-counter-bytecode:test
```
astoycos, line 109: Do you need all these now? I think the below `IMAGE_TC_BC ?=`'s do a define if not defined operation right?

bpfd-operator/test/integration/tc_test.go
```diff
@@ -18,8 +18,8 @@ import (
 )
 
 const (
-	tcCounterBytecodeYaml      = "../../../examples/go-tc-counter/kubernetes-deployment/go-tc-counter-bytecode.yaml"
-	tcGoCounterUserspaceYaml   = "../../../examples/go-tc-counter/kubernetes-deployment/go-tc-counter.yaml"
+	tcCounterBytecodeYaml      = "../../../examples/config/base/go-tc-counter/bytecode.yaml"
```
astoycos, line 21: You can point this to a kustomize file rather than raw yaml now 

And then use `clusters.KustomizeDeployForCluster(ctx, env.Cluster(), tcCounterKustomize)` now I think :)

astoycos: Then we're fully consistent for bpfd core and the examples (always using Kustomize)

examples/config/default/go-tc-counter/kustomization.yaml
astoycos: That's just the default, and it means "main" branch! I'm happy to change everything to latest we'd just need to do it here -> https://github.com/bpfd-dev/bpfd/blob/main/bpfd-operator/Makefile#L50 too

examples/Makefile
astoycos: WFM

examples/config/default/go-tc-counter/kustomization.yaml
Billy99: I don't care, but there are 52 matches across 14 files of go-*-counter:latest in my branch. Then there a several instances of `xdp_pass:latest`, `tc_pass:latest` and `tracepoint:latest` which we aren't updating, so now some are using `main` and some are using `latest`.

astoycos: Lol then lets just update all the core bpfd stuff to latest :)

astoycos: There should be a lot less using `:main`

bpfd-operator/test/integration/tc_test.go
Billy99: Done

astoycos (approved):
Thanks for this!! I want to add a `build-images` tag to the makefile but we can do that in a follow up.

astoycos:
LGTM

PR #518, opened 2023-05-23
Add examples for tc, and setting globals and proceed-on

anfredette:


Billy99 (commented):
Thanks Andre. I like the content. Thinking we should move it down under individual headings. I'll add more sections over time as functionality grows.

anfredette:
@Billy99 thanks for the comments.  I think I made the updates you suggested, but I'm planning to add a little more description to the proceed-on.  So, I put a hold on this while I'm doing that.

anfredette:
@Billy99, I made a few more tweaks.  Please take another look when you have a chance.

Billy99 (approved):
LGTM

Issue #470, opened 2023-05-02
Move path handling into Location


dave-tucker:
We should remove `path` from the `ProgramData` struct and have the path logic encapsulated in [`Location`](https://github.com/bpfd-dev/bpfd/blob/60f7088883479b1f5c0ad00f7c60d5f848065644/bpfd/src/command.rs#L74)

i.e

```rust

impl Location {
    fn path(self) -> PathBuf {
       // Returns the filesystem path
      // Or pulls the image and returns the filesystem path.
    }

}
```

PR #520, opened 2023-05-24
Implement path() as Location method

ykaliuta:
Closes: #470

### DIFF
--- a/bpfd/src/bpf.rs
+++ b/bpfd/src/bpf.rs
@@ -179,25 +179,25 @@ impl BpfManager {
 
     pub(crate) async fn add_multi_attach_program(
         &mut self,
-        program: Program,
+        mut program: Program,
         id: Uuid,
         map_pin_path: String,
     ) -> Result<Uuid, BpfdError> {
         debug!("BpfManager::add_multi_attach_program()");
 
-        let program_bytes = program.data().program_bytes().await?;
+        let program_bytes = program.data_mut().program_bytes().await?;
 
         // This load is just to verify the Section Name is valid.
         // The actual load is performed in the XDP or TC logic.
         let mut ext_loader = BpfLoader::new()
-            .extension(&program.data().section_name.borrow())
+            .extension(&program.data().section_name)
             .map_pin_path(map_pin_path.clone())
             .load(&program_bytes)?;
 
-        match ext_loader.program_mut(&program.data().section_name.borrow()) {
+        match ext_loader.program_mut(&program.data().section_name) {
             Some(_) => Ok(()), // BILLY: Do we need an unload here?
             None => Err(BpfdError::SectionNameNotValid(
-                program.data().section_name.borrow().clone(),
+                program.data().section_name.clone(),
             )),
         }?;
 
@@ -279,7 +279,7 @@ impl BpfManager {
         map_pin_path: String,
     ) -> Result<Uuid, BpfdError> {
         debug!("BpfManager::add_single_attach_program()");
-        let program_bytes = p.data().program_bytes().await?;
+        let program_bytes = p.data_mut().program_bytes().await?;
 
         let mut loader = BpfLoader::new();
 
@@ -291,9 +291,12 @@ impl BpfManager {
             .map_pin_path(map_pin_path.clone())
             .load(&program_bytes)?;
 
-        let raw_program = loader.program_mut(&p.data().section_name.borrow()).ok_or(
-            BpfdError::SectionNameNotValid(p.data().section_name.borrow().clone()),
-        )?;
+        let raw_program =
+            loader
+                .program_mut(&p.data().section_name)
+                .ok_or(BpfdError::SectionNameNotValid(
+                    p.data().section_name.clone(),
+                ))?;
 
         match p.clone() {
             Program::Tracepoint(program) => {
@@ -637,7 +640,7 @@ impl BpfManager {
                         prog_id,
                         ProgramInfo {
                             id: Some(*id),
-                            name: Some(p.data.section_name.borrow().to_string()),
+                            name: Some(p.data.section_name.clone()),
                             program_type: Some(ProgramType::Xdp as u32),
                             location,
                             global_data: Some(p.data.global_data.clone()),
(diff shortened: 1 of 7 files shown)


bpfd/src/command.rs
```diff
@@ -89,6 +89,41 @@ pub(crate) enum Location {
     File(String),
 }
 
+impl Location {
```
astoycos, line 92: Looking at this more... I __think__ it'd be best to only have one function here called `get_program_bytes` it can do all of the verification in one step that we're currently doing in a few steps and a few different places and do it down in `add_*_program` i.e something along the lines of this sudo code (not checked for bugs :))

``` rust
fn get_program_bytes(&self, provided_section_name: String) -> Result<Vec<u8>, BpfdError> {
    match self {
        Location::File(l) => read(l).await,
        Location::Image(l) => {
            let program_overrides = l
                .get_image(None)
                .await
                .map_err(|e| BpfdError::BpfBytecodeError(e.into()))?;

            // If section name isn't provided and we're loading from a container
            // image use the section name provided in the image metadata, otherwise
            // always use the provided section name.
            if section_name.is_empty() {
                section_name = program_overrides.image_meta.section_name
            } else if program_overrides.image_meta.section_name != section_name {
                return Err(BpfdError::BytecodeMetaDataMismatch {
                    image_sec_name: program_overrides.image_meta.section_name,
                    provided_sec_name: section_name,
                });
            }
            get_bytecode_from_image_store(program_overrides.path).await
        }
}
```

That can be called [here](https://github.com/bpfd-dev/bpfd/blob/main/bpfd/src/bpf.rs#L160) in bpf.rs and tc.rs/xdp.rs which should condense all of this logic

bpfd/src/oci_utils/image_manager.rs
```diff
@@ -316,7 +316,9 @@ fn load_image_manifest(image_dir: PathBuf) -> Result<OciImageManifest, anyhow::E
     Ok(serde_json::from_str::<OciImageManifest>(&file_content)?)
 }
 
-fn load_image_meta(image_content_path: PathBuf) -> Result<ContainerImageMetadata, anyhow::Error> {
+pub(crate) fn load_image_meta(
```
astoycos, line 319: So with my previous comment this won't need to be public.

bpfd/src/command.rs
```diff
@@ -227,41 +261,23 @@ impl ProgramData {
...
-                    location,
-                    path: program_overrides.path,
-                    section_name,
-                    global_data,
-                    owner,
-                })
-            }
+        let image_section = location.image_section().await;
+
+        if section_name.is_empty() {
```
astoycos, line 266: Based on my other comment there won't be any more error checking here

bpfd/src/multiprog/tc.rs
```diff
@@ -195,10 +195,11 @@ impl TcDispatcher {
                 let path = format!("{base}/dispatcher_{if_index}_{}/link_{k}", self.revision);
                 new_link.pin(path).map_err(BpfdError::UnableToPinLink)?;
             } else {
-                let program_bytes = if v.data.path.clone().contains(BYTECODE_IMAGE_CONTENT_STORE) {
-                    get_bytecode_from_image_store(v.data.path.clone()).await?
+                let path = v.data.location.path().await?.to_string_lossy().to_string();
+                let program_bytes = if path.contains(BYTECODE_IMAGE_CONTENT_STORE) {
```
astoycos, line 199: Now this will just be something like let program_bytes = v.data.location.get_program_bytes(&program.data().section_name)?

astoycos (changes requested):
Thanks for the work here @ykaliuta and sorry I'm changing up what the issue is asking for, but I feel like we can condense ALL of this logic + err checking down to the event that's really important i.e getting the bpfprogram bytes :)

bpfd/src/command.rs
ykaliuta: Thanks for comments! Trying to implement (pushed).

bpfd/src/bpf.rs
```diff
@@ -635,35 +619,30 @@ impl BpfManager {
...
+            .await;
+            let prog = Program::Tc(TcProgram {
+                data: prog_data.clone(),
+                direction: args.direction,
+                info: TcProgramInfo {
+                    if_index,
+                    current_position: None,
+                    metadata: command::Metadata {
+                        priority: args.priority,
+                        // This could have been overridden by image tags
```
astoycos, line 637: Not anymore :)

bpfd/src/bpf.rs
```diff
@@ -635,35 +619,30 @@ impl BpfManager {
...
+                        name: prog_data.section_name,
```
astoycos, line 638: This can just be args.section_name

bpfd/src/bpf.rs
```diff
@@ -266,17 +260,11 @@ impl BpfManager {
...
-                .clone()
-                .contains(BYTECODE_IMAGE_CONTENT_STORE)
-            {
-                get_bytecode_from_image_store(program.data.path.clone()).await?
-            } else {
-                read(program.data.path.clone()).await?
-            };
-
+                .location
+                .get_program_bytes(&program.data.section_name)
```
astoycos, line 266: So here we **may** need to change the programData's section name based on the image right? 

I think `program_bytes` will actually need to be a method on the `ProgramData` stuct, which will then internally call the `Location.get_program_bytes()` and override `section_name` if needed.

Does that make sense?

bpfd/src/command.rs
```diff
@@ -89,6 +89,39 @@ pub(crate) enum Location {
     File(String),
 }
 
+impl Location {
+    pub(crate) async fn get_program_bytes(
+        &self,
+        section_name: &String,
```
astoycos, line 95: Instead of passing section_name here just return the section name (if any) from this function, and compare it in the outer `ProgramData.program_bytes()` function

astoycos (changes requested):
So this is almost here, sorry for the super slow review :(!! We just need to ensure that program.Data.section_name is correct, i.e it must get updated if needed.  I suggested a way to do that in the comments but feel free to push whatever you think is best!

astoycos:
@ykaliuta PTAL at 47098fcd2406dddbbcb734f441e4de034dbe4114 and lmk what you think :) 

I found it easier to just push a commit rather than comment, I hope that's alright with you! 

If you'd like to drop my commit and just update yours I'm alright with that too!

PR #523, opened 2023-05-26
Add setting of global variables and proceed-on to integration tests

anfredette:
Also includes eBPF source files for programs used for integration tests.

### DIFF
--- a/tests/integration-test/bpf/tc_pass.bpf.c
+++ b/tests/integration-test/bpf/tc_pass.bpf.c
@@ -1,0 +1,15 @@
+#include <linux/bpf.h>
+#include <linux/pkt_cls.h>
+#include <bpf/bpf_helpers.h>
+
+volatile const __u32 GLOBAL_1 = 0;
+volatile const __u32 GLOBAL_2 = 0;
+
+SEC("classifier/pass")
+int pass(struct __sk_buff *skb)
+{
+	bpf_printk(" TC: GLOBAL_1: 0x%08X, GLOBAL_2: 0x%08X", GLOBAL_1, GLOBAL_2);
+	return TC_ACT_OK;
+}
+
+char _license[] SEC("license") = "GPL";
--- a/tests/integration-test/bpf/tc_pass.bpf.c
+++ b/tests/integration-test/bpf/tc_pass.bpf.c
@@ -2,13 +2,13 @@
 #include <linux/pkt_cls.h>
 #include <bpf/bpf_helpers.h>
 
-volatile const __u32 GLOBAL_1 = 0;
-volatile const __u32 GLOBAL_2 = 0;
+volatile const __u8 GLOBAL_u8 = 0;
+volatile const __u32 GLOBAL_u32 = 0;
 
 SEC("classifier/pass")
 int pass(struct __sk_buff *skb)
 {
-	bpf_printk(" TC: GLOBAL_1: 0x%08X, GLOBAL_2: 0x%08X", GLOBAL_1, GLOBAL_2);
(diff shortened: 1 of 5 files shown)


tests/integration-test/bpf/tc_pass.bpf.c
```diff
@@ -0,0 +1,15 @@
+#include <linux/bpf.h>
+#include <linux/pkt_cls.h>
+#include <bpf/bpf_helpers.h>
+
+volatile const __u32 GLOBAL_1 = 0;
+volatile const __u32 GLOBAL_2 = 0;
```
astoycos, line 6: Should we make one of these be a u8 just for a little bit of extra verification that the global data feature is functioning correctly?

tests/integration-test/src/tests/utils.rs
```diff
@@ -49,13 +49,25 @@ pub fn add_xdp_pass(iface: &str, priority: u32) -> Result<String> {
     let output = Command::cargo_bin("bpfctl")?
         .args([
             "load-from-image",
+            "--global",
```
astoycos, line 52: Can we make sure we still have a follow up issue for an E2E test with the global data + proceed_on features that actually loads AND verifies them?

tests/integration-test/bpf/tc_pass.bpf.c
```diff
@@ -0,0 +1,15 @@
+#include <linux/bpf.h>
```
astoycos, line 1: Let's make sure we automatically build any bytecode images for programs which we maintain in tree, just add these three programs to https://github.com/bpfd-dev/bpfd/blob/main/.github/workflows/image-build.yaml#L251

anfredette: I was thinking that would be a good idea but I wasn't sure how to build the bpf bytecode, so I thought it was a good first step to at least check in the source code.  I looked at the file you referenced, but have a few questions.  I'll ping you later.

tests/integration-test/bpf/tc_pass.bpf.c
anfredette: Yes, I agree.  I'm working on the e2e testing and was also thinking that u8 would be nice because the endianness wouldn't matter.

tests/integration-test/src/tests/utils.rs
anfredette: I think that's covered in https://github.com/bpfd-dev/bpfd/issues/394.

astoycos:
Apart from those comments this is looking good!

anfredette:
@astoycos @Billy99 FYI, I just needed to add the bpf source code to the `bpfd/bpf` directory to have it built with `cargo xtask build-ebpf`.  I still have some more to do on this one, so I'm going to add a hold.

tests/integration-test/bpf/tc_pass.bpf.c
anfredette: Done

tests/integration-test/bpf/tc_pass.bpf.c
anfredette: I've updated build_ebpf so that the integration test eBPF code gets built with `cargo xtask build-ebpf`.  However, I've opened issue #533 for the automatic building and pushing of the bytecode container files because I think there are a couple of open questions about how to do it.

anfredette:
@Billy99 this should be ready to go if you have time to take a look.  FYI, I'm also working on a separate pr for the functional testing, but I think it's better to keep that separate.

Billy99 (approved):
LGTM

PR #529, opened 2023-05-29
One bpfProgram K8s object per eBPF program

### LINKED ISSUE #459: Convert from a Single BpfProgram per *Program Type to Multiple
Today in our Kubernetes deployment we always create a single BpfProgram object for each explicit `*Program` type that is created . 

With the addition of the ability to specify UUID along with the desired future feature to list bpfPrograms already running on the system we would like to make an explicit BpfProgram Object for every individual instance of a bpfProgram. 

This will mean that multiple BpfPrograms could now be created for a single `*Program` object.

astoycos:
Fixes #459


astoycos:
I need to build example images in actions here

astoycos:
Blocked by #555

dave-tucker (approved):
LGTM

PR #534, opened 2023-06-02
Pushing images to quay.io when the code changes in bpfctl

### LINKED ISSUE #532: Image builds are not using the cache
Originally, @anfredette noted that:

> Currently, the images defined [here](https://github.com/bpfd-dev/bpfd/blob/main/.github/workflows/image-build.yaml) get pushed to quay.io every time a pr is merged.  We should only do that if the image changes.

# Expected Result

What should happen, is that:

- We build images on the same triggers we do now
- We store layers in a cache and re-use them to make image builds faster
- We push to the registry on same triggers as today, but if nothing has changed then the push should be a noop since all the layers already exist in quay.io.

# What's actually happening

I started working on #469 and #484 to speed up our container builds and I noticed that the podman/buildah actions don't store any layer cache in GH actions (see: https://github.com/redhat-actions/buildah-build/issues/90) so I started by moving over to the docker/build-and-push action which does.

In doing so I noticed that *something* from the context is getting copied on a `COPY` in the Dockerfile somewhere and invalidating the cache.

So not only do we not store layers between builds... but even if we did, in cases where no relevant files changed, we're copying these in a `COPY` dockerfile statement anyway, meaning the cache won't be hit.

# What needs to be done

- Move all images builds to docker/build-and-push action and use the gha cache
- .dockerignore all the things that aren't strictly necessary to build images
- only `COPY` the necessary things into the Dockerfile that are needed for each build (since if the payload is the EXACTLY the same then the layer will be re-used from cache)

krishnaduttPanchagnula:
closes #532


krishnaduttPanchagnula:
@anfredette is there any other changes from my end for this pr?

Billy99 (commented):
Thank you for your submission. I am not a github actions expert, so I had a question or two on PR regarding your change.

Also, when Andre wrote the Issue, we knew bpfd was pushing images even if it didn't need to, but we weren't sure on how to fix it. Just checking the sub-directory for changes doesn't take into account dependency changes. I don't have the answer, but I think more is going to have to be done.

anfredette (changes requested):
@krishnaduttPanchagnula I also thank you for this pr, but I'm also not a github actions expert and had some of the same questions as @Billy99.  I was planning to talk with some people today about it.  Effectively, what we are looking for is that if anything changes that would cause the executable image to change, we'd want to build a new container image and push it to quay.  Also, we want to do this for all of the images that we push to quay.  I was using bpfctl as an example.  However, I think it would be good if we could get this working for bpfctl first, and then we can repeat the pattern of the others.

anfredette:
> we can run a shell script as a step in which we can check for all the paths desired

Would we need to manually identify all of the relevant paths in this script?  If so, that seems error-prone.  On a separate note, our colleague @dave-tucker had some other ideas and is going follow up.

dave-tucker:
Thanks for the contribution @krishnaduttPanchagnula

Unfortunately I hadn't got around to triaging #532 just yet 😢 
The approach taken here is not going to build/push bpfd if/when we have changes to dependencies etc...
I know a shell script could be used to accomplish this, but it's missing the root cause.
Based on that, I'll close this PR for now.

I'm going to add some more information about what we need in the original issue and we'd welcome any help you have to offer.

PR #555, opened 2023-06-21
Example build target

astoycos:
Add targets to the examples makefile to build all the user space images and to easily load them into kind.

Also update our Kubernetes integration tests to use fresh userspace images in CI.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -104,6 +104,22 @@ build: fmt ## Build all the userspace example code.
 generate: ## Run `go generate` to build the bytecode for each of the examples.
 	go generate ./...
 
+.PHONY: build-us-images
+build-us-images: ## Build all example userspace images
+	docker build -t ${IMAGE_TC_US} -f ./go-tc-counter/container-deployment/Containerfile.go-tc-counter ../
+	docker build -t ${IMAGE_TP_US} -f ./go-tracepoint-counter/container-deployment/Containerfile.go-tracepoint-counter ../
+	docker build -t ${IMAGE_XDP_US} -f ./go-xdp-counter/container-deployment/Containerfile.go-xdp-counter ../
+
+.PHONY: push-us-images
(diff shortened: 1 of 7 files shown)


astoycos:
@Billy99 PTAL

examples/Makefile
```diff
@@ -122,6 +138,7 @@ IMAGE_TP_US ?= quay.io/bpfd-userspace/go-tracepoint-counter:latest
 IMAGE_XDP_BC ?= quay.io/bpfd-bytecode/go-xdp-counter:latest
 IMAGE_XDP_US ?= quay.io/bpfd-userspace/go-xdp-counter:latest
 KUST_DIR=default
+KIND_CLUSTER_NAME ?= bpfd-deployment
```
Billy99, line 141: Add `KIND_CLUSTER_NAME` to the Makefile comments section above, something like:
```
KIND_CLUSTER_NAME: ## Name of the deployed cluster to load to, defaults to `bpfd-deployment`
```

examples/Makefile
```diff
@@ -104,6 +104,22 @@ build: fmt ## Build all the userspace example code.
 generate: ## Run `go generate` to build the bytecode for each of the examples.
 	go generate ./...
 
+.PHONY: build-userspace-images
```
Billy99, line 107: Probably fine if we think this will mostly be used from Github workflow, but if the user will call a lot, may want to shorten all the names from *-userspace-images to *-us-images

Billy99 (commented):
Two nits.

examples/Makefile
astoycos: Done

examples/Makefile
astoycos: I wen't ahead and did this here 👍

Billy99 (approved):
LGTM

PR #556, opened 2023-06-23
Uprobe support

anfredette:
This pr adds support to bpfd for installing uprobe programs.

Example cli:
bpfctl load-from-image -g GLOBAL_u8=01 GLOBAL_u32=0D0C0B0A --image-url quay.io/bpfd-bytecode/mallocprobe:latest uprobe -f "malloc" -t "libc"


anfredette:
@dave-tucker @astoycos This pr is ready for review when you get a chance.

bpfd/src/bpf.rs
```diff
@@ -333,9 +337,88 @@ impl BpfManager {
                     Err(BpfdError::UnableToPinProgram(e))
                 })?;
 
+            Ok(id)
+        } else if let Program::Uprobe(ref program) = p {
```
anfredette, line 341: We should be able to streamline the add_single_attach_program() function and have less duplicated code, but I decided that I'd tackle that when I handle the other probe types.

dave-tucker (changes requested):
A couple of small changes but overall this looks great. Thanks @anfredette.

tests/integration-test/bpf/build_push_images.sh
```diff
@@ -35,3 +35,15 @@ docker build \
 
 docker push quay.io/bpfd-bytecode/tracepoint
 
+docker build \
+ --build-arg PROGRAM_NAME=mallocprobe \
+ --build-arg SECTION_NAME=malloc \
+ --build-arg PROGRAM_TYPE=uprobe \
+ --build-arg BYTECODE_FILENAME=mallocprobe.bpf.o \
+ --build-arg KERNEL_COMPILE_VER=$(uname -r) \
```
astoycos, line 43: Please remove the Kernel_Compile_Ver tag from here RE https://github.com/bpfd-dev/bpfd/pull/505/commits/125b7c357a496f3480ef4dc0c013afacb979ca77

bpfctl/src/main.rs
```diff
@@ -165,6 +165,31 @@ enum LoadCommands {
...
+        /// Optional: function to attach the uprobe to.
+        #[clap(short, long)]
+        fn_name: Option<String>,
+
+        /// Optional: offset added to the address of the target function
+        /// (or beginning of target if no function is identified)
+        #[clap(short, long)]
+        offset: Option<u64>,
+
+        /// Required: absolute path to a binary or library, or a library name
```
astoycos, line 179:
```suggestion
        /// Required: absolute path to a binary, library, or library name
```

bpfctl/src/main.rs
```diff
@@ -165,6 +165,31 @@ enum LoadCommands {
...
+        /// Example: --target "libc".
+        #[clap(short, long)]
+        target: String,
+
+        /// Optional: process identification number (PID) to run uprobe for.
```
astoycos, line 184:
```suggestion
        /// Optional: process identification number (PID) to attach uprobe to.
```

bpfd-api/src/lib.rs
```diff
@@ -147,6 +149,7 @@ impl TryFrom<i32> for ProgramType {
             29 => ProgramType::Lsm,
             30 => ProgramType::SkLookup,
             31 => ProgramType::Syscall,
+            32 => ProgramType::Uprobe,
```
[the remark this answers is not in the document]
anfredette: I was wondering why uprobe wasn't in the list.  So, I'll use kprobe. I think we'll want to be able to differentiate between the different probe types for the user.  I have a couple of ideas but will come up with a proposal after I do a little more investigation.

bpfd/src/bpf.rs
```diff
@@ -333,9 +337,88 @@ impl BpfManager {
...
+                .attach(
+                    program.info.fn_name.as_deref(),
+                    program.info.offset.unwrap_or_default(),
+                    program.info.target.clone(),
+                    program.info.pid,
+                )
+                .or_else(|e| {
+                    p.delete(id).map_err(|_| {
+                        BpfdError::Error(
+                            "new dispatcher cleanup failed, unable to delete program data"
```
astoycos, line 392: This comment is wrong i think :) 

There's no dispatcher here!

bpfd/src/bpf.rs
anfredette: https://github.com/bpfd-dev/bpfd/issues/564

astoycos (commented):
Pretty much there just some NITs :)

tests/integration-test/bpf/build_push_images.sh
astoycos: + anywhere else you might have added it :)

bpfctl/src/main.rs
anfredette: "absolute path" only applies to "binary" and "library", and not to "library name".  So I think the way I wrote it is technically correct.  However, it may not be clear.  I'll change it to the following:
"library name or the absolute path to a binary or library"

tests/integration-test/bpf/build_push_images.sh
anfredette: Ok.

anfredette: I also rebuilt and pushed the images.

bpfctl/src/main.rs
anfredette: I think that technically, the uprobe is attached to the function in the target, but if the PID is set, it only executes when that function is executed by the given PID.  The doc comment from the Aya code is below:
```rust
    /// Attaches the uprobe to the function `fn_name` defined in the `target`.
    /// If `offset` is non-zero, it is added to the address of the target
    /// function. If `pid` is not `None`, the program executes only when the target
    /// function is executed by the given `pid`.
```

bpfd/src/bpf.rs
anfredette: Yup.  That was a copy-pasted from the tracepoint code (which I assume was copy-pasted from the multi-attach code).  I'll fix both places.

bpfctl/src/main.rs
```diff
@@ -328,6 +391,26 @@ impl LoadCommands {
...
                 }),
             )),
+            LoadCommands::Uprobe {
+                fn_name,
+                offset,
+                target,
+                pid,
+                namespace,
+            } => {
+                if !namespace.is_none() {
```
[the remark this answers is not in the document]
anfredette: Done.

bpfctl/src/main.rs
anfredette: Changed to the following:
"only execute uprobe for given process identification number (PID)"

anfredette:
@dave-tucker and @astoycos, thanks for the review!  I think I've addressed all the comments.

dave-tucker (approved):

PR #571, opened 2023-07-06
consolidate add_single_attach_program()

astoycos:
Consolidate some duplicated code in the add_single_attach_program() function.


astoycos:
@anfredette Small change

anfredette (commented):
lgtm

However, we're having a failure in Codecov which doesn't look related to your change.  I've seen something similar before and rerunning the test resulted in a pass.  I reran it once, but that didn't fix it.  I just tried one more time.

anfredette (approved):
3rd time's a charm.  All passing now.

Issue #564, opened 2023-06-30
Add support for uretprobe, kprobe and kretprobe


anfredette:
When this is done, also streamline the add_single_attach_program() function to remove duplicated code.

anfredette:
Still need to do uprobes

PR #572, opened 2023-07-07
Add bpfd support for uretprobes, kprobes and kretprobes

anfredette:
This pr expands on the previous uprobe pr (#556) to add the rest of the probe types.

Fixes issue #564.


anfredette:
Integration tests are passing for me locally, but failing upstream, so I'll have to look into that.  Is it possible for me to get access to /tmp/bpfd_trace_pipe.log from the test system?

astoycos:
@anfredette The easiest way would be to add a step to the tests that dumps those logs to a file on a failure

If not you can try using 

https://github.com/marketplace/actions/debugging-with-ssh

Billy99:
Can we not do something like:
https://github.com/bpfd-dev/bpfd/blob/main/.github/workflows/build.yml#L240

But for `basic-integration-tests`?

astoycos:
Yep we can do that!

anfredette:
Thanks for the tips.  I'll look into it when I have the chance.

tests/integration-test/bpf/build_push_images.sh
```diff
@@ -33,13 +33,43 @@ docker build \
 docker push quay.io/bpfd-bytecode/tracepoint
 
 docker build \
```
astoycos, line 35: Do we have an issue for building and pushing these in actions?

tests/integration-test/src/tests/e2e.rs
```diff
@@ -23,7 +23,10 @@ const TC_EG_GLOBAL_4_LOG: &str = "bpf_trace_printk:  TC: GLOBAL_u8: 0x35";
 const TC_EG_GLOBAL_5_LOG: &str = "bpf_trace_printk:  TC: GLOBAL_u8: 0x3B";
 const TC_EG_GLOBAL_6_LOG: &str = "bpf_trace_printk:  TC: GLOBAL_u8: 0x3D";
 const TRACEPOINT_GLOBAL_1_LOG: &str = "bpf_trace_printk:  TP: GLOBAL_u8: 0x25";
-const UPROBE_GLOBAL_1_LOG: &str = "bpf_trace_printk:  TP: GLOBAL_u8: 0x25";
+// const UPROBE_GLOBAL_1_LOG: &str = "bpf_trace_printk:  UP: GLOBAL_u8: 0x25";
```
astoycos, line 26: nit extra comment?

tests/integration-test/src/tests/e2e.rs
```diff
@@ -387,8 +417,14 @@ fn test_program_execution_with_global_variables() {
     debug!("Successfully validated tc egress global variable");
     assert!(trace_pipe_log.contains(TRACEPOINT_GLOBAL_1_LOG));
     debug!("Successfully validated tracepoint global variable");
-    assert!(trace_pipe_log.contains(UPROBE_GLOBAL_1_LOG));
-    debug!("Successfully validated uprobe global variable");
+    assert!(trace_pipe_log.contains(KPROBE_GLOBAL_1_LOG));
+    debug!("Successfully validated kprobe global variable");
+    assert!(trace_pipe_log.contains(KRETPROBE_GLOBAL_1_LOG));
+    debug!("Successfully validated kretprobe global variable");
+    // assert!(trace_pipe_log.contains(UPROBE_GLOBAL_1_LOG));
```
astoycos, line 424: nit extra comment?

tests/integration-test/src/tests/utils.rs
```diff
@@ -200,7 +203,7 @@ pub fn add_uprobe(globals: Option<Vec<&str>>) -> Result<String> {
         "Always",
         "uprobe",
         "-f",
-        "malloc",
+        "fwrite",
```
astoycos, line 206: Any reason for this change?

astoycos (changes requested):
Some comments questions and Nits

tests/integration-test/bpf/build_push_images.sh
```diff
@@ -33,13 +33,43 @@ docker build \
...
- --build-arg PROGRAM_NAME=mallocprobe \
- --build-arg SECTION_NAME=malloc \
- --build-arg PROGRAM_TYPE=uprobe \
- --build-arg BYTECODE_FILENAME=mallocprobe.bpf.o \
+ --build-arg PROGRAM_NAME=uprobe \
+ --build-arg SECTION_NAME=probe \
+ --build-arg PROGRAM_TYPE=kprobe \
```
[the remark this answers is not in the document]
anfredette: I agree that it's confusing. Let's discuss how to handle this.

tests/integration-test/src/tests/e2e.rs
anfredette: As mentioned on the call, I have the uprobe tests commented out currently because they aren't working upstream.  I still need to work it out.

tests/integration-test/src/tests/e2e.rs
anfredette: Same as above.

tests/integration-test/src/tests/utils.rs
anfredette: It seemed to be a better choice because the test itself is using fwrite when saving the log to a file, but it still didn't work upstream.

tests/integration-test/bpf/build_push_images.sh
anfredette: I thought we did, but I can't find it right now.  I'll do a little more searching and will add one if I can't find one.  However, we usually do this when a pr is merged, but, when you're adding new test cases, the images need to be available for the new test cases to pass.  So, I think you also need to be able to push them manually.

anfredette:
The latest set of commits changes the API to use separate AttachInfo messages for kprobes and uprobes.

The integration tests work for me locally, but I still need to figure out why the uprobe integration tests are failing upstream, so I'm still expecting the uprobe tests to fail.

anfredette:
I reordered and squashed some of my commits before rebasing. This should be ready to go once everything passes.

astoycos (approved):
/LGTM

PR #581, opened 2023-07-14
Add pull-bytecode API

### LINKED ISSUE #429: Allow for pre-pulling of bytecode images

Molter73:
Implement the pull-bytecode API, which allows for users to pre-pull bytecode images as a caching method for future load commands.

Fixes #429


Molter73:
For anyone reading this, I still need to do some more testing and maybe some cleanup, but if people have comments on what the API looks like, feel free to comment away.

astoycos (approved):
/LGTM

Really great work here I couldn't find anything wrong, thanks @Molter73 :) 

As a follow up issue to this  (can you make one please?) we also need to figure out how to load an image which already on the machine (either in docker/podman local image store).  Then we need to document the process of rebuilding and loading local bytecode images all without pushing/pulling from a remote repository

PR #588, opened 2023-07-17
fix for ubuntu mirror issues

astoycos:
As advised here -> https://docs.github.com/en/actions/using-github-hosted-runners/customizing-github-hosted-runners#installing-software-on-ubuntu-runners


anfredette (approved):
LGTM

Issue #320, opened 2023-02-02
Reflect all bpf program state from nodes into k8s


dave-tucker:
bpfd should iterate all bpf programs on the host and load those into the k8s api.
to do this we will need to:

1. Support displaying information for all program types, even if we don't support loading them
2. Maintain a mapping between bpf fd and uuid on each node

astoycos:
Maintain a mapping b/w Kernel ID (Int BpfProgram Counter for a specific Node) Can we turn this into a namespaced UUID (UUID5 RFC namespace for each node) to correlate it to a specific node?

astoycos:
Blocked on API refactor #340

astoycos:
Done in #570 and #605

PR #605, opened 2023-07-21
List programs k8s

astoycos:
Works towards finishing #320 

This adds a new controller that creates bpfProgram object for programs not owned by bpfd.

These bpfProgram objects contain annotations which reflect kernel state as shown below

```bash
kubectl get bpfprograms -l bpfd.dev/unownedProgram -l kubernetes.io/hostname=bpfd-deployment-worker
NAME                                           AGE
dump-bpf-map-693-bpfd-deployment-worker        17h
dump-bpf-prog-694-bpfd-deployment-worker       17h
my-uretprobe-93125-bpfd-deployment-worker      17h
restrict-filesy-63130-bpfd-deployment-worker   17h
sd-fw-egress-63149-bpfd-deployment-worker      17h
sd-fw-egress-63171-bpfd-deployment-worker      17h
sd-fw-egress-63173-bpfd-deployment-worker      17h
sd-fw-egress-63175-bpfd-deployment-worker      17h
sd-fw-egress-63177-bpfd-deployment-worker      17h
sd-fw-ingress-63150-bpfd-deployment-worker     17h
sd-fw-ingress-63172-bpfd-deployment-worker     17h
sd-fw-ingress-63174-bpfd-deployment-worker     17h
sd-fw-ingress-63176-bpfd-deployment-worker     17h
sd-fw-ingress-63178-bpfd-deployment-worker     17h
```

```bash
kubectl get bpfprogram dump-bpf-map-693-bpfd-deployment-worker -o jsonpath='{.metadata.annotations}'

{"BTF-ID":"154","GPL-Compatible":"true","JITed":"true","Kernel-Allocated-Memory-Bytes":"4096","Kernel-ID":"693","Loaded-At":"2023-03-02T18:15:06+0000","Map-IDs":"[45]","Name":"dump_bpf_map","Size-JITed-Bytes":"287","Size-Translated-Bytes":"264","Tag":"749172daffada61f","Type":"tracing","Verified-Instruction-Count":"34"}
```

I decided to use annotations instead of status to store this metadata based on the existing [Kubernetes API conventions ](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md)

More specifically Annotations should be used to 

```
annotations: a map of string keys and values that can be used by external tooling to store and retrieve arbitrary metadata about this object (see [the annotations docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/))
```

and 

```
Annotations enable third-party automation and tooling to decorate objects with additional metadata for their own use.
```

While the status field contains conditions that

```
Conditions provide a standard mechanism for higher-level status reporting from a controller.
```

IMO in this case the metadata we're reporting to the users isn't actually a controller status, additionally there is no Key/Value store in the existing `metav1.Conditions` object which makes storing all the kernel information much more difficult. 

TODO

We still need to make sure a program loaded by bpfd ALSO get's these annotations, but that will require core some design changes to be done correctly so I'll work on it in another PR.

### DIFF
--- a/bpfd-operator/internal/constants.go
+++ b/bpfd-operator/internal/constants.go
@@ -19,30 +19,47 @@ package internal
 import "fmt"
 
 const (
-	XdpProgramControllerFinalizer        = "bpfd.io.xdpprogramcontroller/finalizer"
-	TcProgramControllerFinalizer         = "bpfd.io.tcprogramcontroller/finalizer"
-	TracepointProgramControllerFinalizer = "bpfd.io.tracepointprogramcontroller/finalizer"
-	XdpProgramInterface                  = "bpfd.io.xdpprogramcontroller/interface"
-	TcProgramInterface                   = "bpfd.io.tcprogramcontroller/interface"
-	TracepointProgramTracepoint          = "bpfd.io.tracepointprogramcontroller/tracepoint"
-	BpfProgramOwnerLabel                 = "bpfd.io/ownedByProgram"
-	K8sHostLabel                         = "kubernetes.io/hostname"
-	BpfdNs                               = "bpfd"
-	BpfdOperatorName                     = "bpfd-operator"
-	BpfdDsName                           = "bpfd-daemon"
-	BpfdConfigName                       = "bpfd-config"
-	BpfdDaemonManifestPath               = "./config/bpfd-deployment/daemonset.yaml"
-	BpfdMapFs                            = "/run/bpfd/fs/maps"
-	DefaultConfigPath                    = "/etc/bpfd/bpfd.toml"
-	DefaultRootCaPath                    = "/etc/bpfd/certs/ca/ca.crt"
-	DefaultCertPath                      = "/etc/bpfd/certs/bpfd/tls.crt"
-	DefaultKeyPath                       = "/etc/bpfd/certs/bpfd/tls.key"
-	DefaultClientCertPath                = "/etc/bpfd/certs/bpfd-client/tls.crt"
-	DefaultClientKeyPath                 = "/etc/bpfd/certs/bpfd-client/tls.key"
-	DefaultType                          = "tcp"
-	DefaultPath                          = "/run/bpfd/bpfd.sock"
-	DefaultPort                          = 50051
-	DefaultEnabled                       = true
+	XdpProgramInterface         = "bpfd.dev.xdpprogramcontroller/interface"
+	TcProgramInterface          = "bpfd.dev.tcprogramcontroller/interface"
+	TracepointProgramTracepoint = "bpfd.dev.tracepointprogramcontroller/tracepoint"
+	BpfProgramOwnerLabel        = "bpfd.dev/ownedByProgram"
+	K8sHostLabel                = "kubernetes.io/hostname"
+	DiscoveredLabel             = "bpfd.dev/discoveredProgram"
+	BpfdNs                      = "bpfd"
+	BpfdOperatorName            = "bpfd-operator"
+	BpfdDsName                  = "bpfd-daemon"
+	BpfdConfigName              = "bpfd-config"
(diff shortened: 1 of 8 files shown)


astoycos:
@anfredette This is a good example of a super basic bpfd-agent controller (doesn't use any of the shared logic in `common.go` and only lists programs from bpfd)

bpfd-operator/controllers/bpfd-agent/unowned_program_test.go
```diff
@@ -0,0 +1,211 @@
...
+*/
+
+package bpfdagent
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	//"github.com/google/go-cmp/cmp"
```
astoycos, line 24: Nit remoe

bpfd-operator/controllers/bpfd-agent/unowned_program.go
```diff
@@ -0,0 +1,147 @@
...
+package bpfdagent
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"strings"
+	"time"
+
+	//corev1 "k8s.io/api/core/v1"
```
astoycos, line 23: nit: remove commented deps

bpfd-operator/internal/constants.go
```diff
@@ -21,12 +21,13 @@ import "fmt"
 const (
 	XdpProgramControllerFinalizer        = "bpfd.io.xdpprogramcontroller/finalizer"
```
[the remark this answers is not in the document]
astoycos: Good point I probably just missed all these others let me update all of them

bpfd-operator/controllers/bpfd-agent/unowned_program.go
```diff
@@ -0,0 +1,147 @@
+/*
+Copyright 2022.
```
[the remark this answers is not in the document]
astoycos: :) Time to update!

bpfd-operator/internal/constants.go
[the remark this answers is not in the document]
astoycos: https://github.com/bpfd-dev/bpfd/pull/613

astoycos:
Should be gtg @Billy99

bpfd-operator/controllers/bpfd-agent/unowned_program.go
```diff
@@ -0,0 +1,147 @@
...
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	syncDurationUnownedController = 30 * time.Second
+)
+
+type UnownedProgramReconciler struct {
```
dave-tucker, line 43: Naming is hard 🤣 
I'm not a fan of `Unowned` programs - they are owned by someone, but that someone isn't bpfd.

Some alternatives:
- `RoguePrograms`
- `DiscoveredPrograms`
- `ForeignPrograms`
- `NodeLocalPrograms`

dave-tucker (changes requested):

bpfd-operator/controllers/bpfd-agent/unowned_program.go
astoycos: LOL Fair enough https://martinfowler.com/bliki/TwoHardThings.html

Billy99: Of that list, `DiscoveredPrograms` mkaes the most sense to me.

Billy99 (approved):
LGTM

bpfd-operator/controllers/bpfd-agent/unowned_program.go
astoycos: Done

dave-tucker (approved):

PR #612, opened 2023-07-24
go: go mod tidy

dave-tucker:


astoycos:
Good look we should probably have a go-mod-tidy step in CI

PR #629, opened 2023-08-03
move direction to tcInfo

astoycos:
The direction field for tc programs was at the wrong level internally fix this by moving the field into tcProgramInfo. Add a comment to `ProgramData`.


anfredette (approved):
LGTM

PR #637, opened 2023-08-03
Remove check to block unloading of a Map Owner

Billy99:
There was a check in bpfd which prevent an eBPF program from being deleted if it's maps were being referenced by another eBPF Program. The check was added during the original map sharing PR because it caused problems, but that is no longer the case.

bpfd used to remove the Map Owner from the list of UUIDs in the MapsUsedBy field, because it looked cleaner. The Map Owner is now included in the list of UUIDs in the MapsUsedBy because the Map Owner eBPF Program can now be unloaded, so it is useful to know if it is there or not.

### DIFF
--- a/bpfd/src/bpf.rs
+++ b/bpfd/src/bpf.rs
@@ -501,11 +501,6 @@ impl BpfManager {
             if !(prog.owner() == &owner || owner == SUPERUSER) {
                 return Err(BpfdError::NotAuthorized);
             }
-            if !self.is_map_safe_to_delete(id, prog.data().map_owner_uuid) {
-                return Err(BpfdError::Error(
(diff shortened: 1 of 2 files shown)


Billy99:
> Sweet thanks for checking this @Billy99
> 
> So now do "updates" just work as it pertains to K8s? i.e if program A is loaded and reloaded will the maps persist the process (if another program B is using them?)

I think it will. Just need some more K8s testing. All my testing was on host. Created XDP-Counter-A, created XDP-Counter-B sharing maps with A, unloaded A, created XDP-Counter-C sharing maps with A, and unloaded B & C. Everything was stabled and worked fine.

astoycos (approved):
LGTM

PR #638, opened 2023-08-04
Fix  bpfd-agent race

astoycos:
fix a race that @Billy99 found where we were creating a bpfProgram even though the parent *Program was being deleted.

In the meantime try and cleanup the core shared code to be a bit easier to understand. 

Functionally the main change here is that now a given BpfProgram will be created and reconciled in the same loop meaning that all bpfProgram objects wont be created before any program is actually loaded via bpfd.

### DIFF
--- a/bpfd-operator/controllers/bpfd-agent/common.go
+++ b/bpfd-operator/controllers/bpfd-agent/common.go
@@ -236,7 +236,9 @@ func (r *ReconcilerCommon) createBpfProgram(ctx context.Context,
 // object(s), and ultimately telling the custom controller types to load real
 // bpf programs on the node via bpfd. Additionally it acts as a central point for
 // interacting with the K8s API. This function will exit if any action is taken
-// against the K8s API.
+// against the K8s API. If the function returns a retry boolean and error, the
+// reconcile will be retried based on a default 5 second interval if the retry
+// boolean is set to `true`.
 func reconcileProgram(ctx context.Context,
 	rec bpfdReconciler,
 	program client.Object,
@@ -247,7 +249,13 @@ func reconcileProgram(ctx context.Context,
 	// initialize reconciler state
 	r := rec.getRecCommon()
 
-	// determine which node local actions should be taken based on wether the node is selected
+	// Get program bytecode, source could be an OCI container image or filepath
+	bytecode, err := bpfdagentinternal.GetBytecode(r.Client, &common.ByteCode)
+	if err != nil {
+		return false, fmt.Errorf("failed to process bytecode selector: %v", err)
+	}
+
+	// Determine which node local actions should be taken based on whether the node is selected
 	// OR if the *Program is being deleted.
 	isNodeSelected, err := isNodeSelected(&common.NodeSelector, ourNode.Labels)
 	if err != nil {
@@ -256,23 +264,19 @@ func reconcileProgram(ctx context.Context,
 
 	isBeingDeleted := !program.GetDeletionTimestamp().IsZero()
 
-	// Get existing bpfPrograms for a *Program.
+	// Query the K8s API to get a list of existing bpfPrograms for this *Program
+	// on this node.
 	existingPrograms, err := r.getExistingBpfProgs(ctx, program)
 	if err != nil {
-		return false, fmt.Errorf("failed to get bpfPrograms: %v", err)
+		return false, fmt.Errorf("failed to get existing bpfPrograms: %v", err)
 	}
 
-	// Get expected bpfPrograms for a *Program.
+	// Generate the list of BpfPrograms for this *Program. This handles the one
+	// *Program to many BpfPrograms (i.e. One *Program maps to multiple
+	// interfaces because of PodSelector)
 	expectedPrograms, err := rec.expectedBpfPrograms(ctx)
 	if err != nil {
-		r.Logger.Error(err, "failed to create bpfPrograms")
-		return true, nil
-	}
-
-	// Get program bytecode, source could be an OCI container image or filepath
-	bytecode, err := bpfdagentinternal.GetBytecode(r.Client, &common.ByteCode)
-	if err != nil {
-		return false, fmt.Errorf("failed to process bytecode selector: %v", err)
+		return true, fmt.Errorf("failed to get expected bpfPrograms: %v", err)
 	}
 
 	// multiplex signals into kubernetes API actions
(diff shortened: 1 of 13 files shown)


bpfd-operator/controllers/bpfd-agent/common.go
```diff
@@ -234,109 +235,120 @@ func (r *ReconcilerCommon) createBpfProgram(ctx context.Context,
...
 	ourNode *v1.Node,
 	programMap map[string]*gobpfd.ListResponse_ListResult) (bool, error) {
 
+	// initialize reconciler state
 	r := rec.getRecCommon()
-	r.bpfPrograms = map[string]bpfdiov1alpha1.BpfProgram{}
 
-	// Populate already existing bpfPrograms for *Program.
-	err := r.getBpfPrograms(ctx, program)
+	// determine which node local actions should be taken based on wether the node is selected
```
Billy99, line 250:
```suggestion
	// Determine which node local actions should be taken based on whether the node is selected
```

bpfd-operator/controllers/bpfd-agent/common.go
```diff
@@ -234,109 +235,120 @@ func (r *ReconcilerCommon) createBpfProgram(ctx context.Context,
...
-		if !exists {
-			opts := client.CreateOptions{}
-			r.Logger.Info("creating bpfProgram", "Name", prog.Name)
-			if err := r.Create(ctx, &prog, &opts); err != nil {
-				return true, fmt.Errorf("failed to create bpfProgram object: %v", err)
-			}
-			r.bpfPrograms[prog.Name] = prog
-			return false, nil
-		}
+	// Get existing bpfPrograms for a *Program.
```
Billy99, line 259:
```suggestion
	// Call K8s to get the list of existing bpfPrograms for this *Program on this node.
```

bpfd-operator/controllers/bpfd-agent/common.go
```diff
@@ -234,109 +235,120 @@ func (r *ReconcilerCommon) createBpfProgram(ctx context.Context,
...
+	existingPrograms, err := r.getExistingBpfProgs(ctx, program)
+	if err != nil {
+		return false, fmt.Errorf("failed to get bpfPrograms: %v", err)
 	}
 
-	isNodeSelected, err := isNodeSelected(&common.NodeSelector, ourNode.Labels)
+	// Get expected bpfPrograms for a *Program.
```
Billy99, line 265:
```suggestion
	// Create a list of BpfPrograms for this *Program. This handles the one *Program to many
	// BpfPrograms (i.e. One *Program maps to multiple interfaces because of PodSelector)
```

bpfd-operator/controllers/bpfd-agent/common.go
```diff
@@ -234,109 +235,120 @@ func (r *ReconcilerCommon) createBpfProgram(ctx context.Context,
...
-		// Only stop reconcile loop if we updated the status for a bpfprogram
-		if !alreadySet {
-			return false, nil
+			// If bpfProgram Maps isn't up to date just update it and return
+			if !reflect.DeepEqual(prog.Spec.Maps, r.expectedMaps) && len(r.expectedMaps) != 0 {
+				r.Logger.V(1).Info("Updating bpfProgram Object", "Maps", r.expectedMaps, "bpfProgram", prog.Name)
+				prog.Spec.Maps = r.expectedMaps
+				if err := r.Update(ctx, &prog, &client.UpdateOptions{}); err != nil {
+					r.Logger.Error(err, "failed to update bpfProgram's Programs")
+					return true, nil
```
Billy99, line 340: Confused on the `retry` flag. Here `return true, nil`, but most everywhere else, it is either `return false, nil` or `return true, fmt.Errorf()`.

In `Reconcile()` in xdp-program.go
```
func Reconcile() {
	:
	for _, xdpProgram := range xdpPrograms.Items {
		:
		retry, err := reconcileProgram(ctx, r, r.currentXdpProgram, &r.currentXdpProgram.Spec.BpfProgramCommon, r.ourNode, programMap)
		if err != nil {
			r.Logger.Error(err, "Reconciling XdpProgram Failed", "XdpProgramName", r.currentXdpProgram.Name, "Retrying", retry)
			return ctrl.Result{Requeue: retry, RequeueAfter: retryDurationAgent}, nil
		}
	}
	return ctrl.Result{Requeue: false}, nil
}
```
So retry flag is only honored in there is an error returned.

Billy99 (commented):
Couple of comments, mostly more descriptive comments.

I reran my tests that I saw the original issue with (moved it back to 3 nodes for testing) and no issues after about 8 cycles.

bpfd-operator/controllers/bpfd-agent/common.go
```diff
@@ -234,109 +235,120 @@ func (r *ReconcilerCommon) createBpfProgram(ctx context.Context,
...
 	if err != nil {
-		return false, fmt.Errorf("failed to check if node is selected: %v", err)
+		r.Logger.Error(err, "failed to create bpfPrograms")
+		return true, nil
 	}
 
-	isBeingDeleted := !program.GetDeletionTimestamp().IsZero()
-
+	// Get program bytecode, source could be an OCI container image or filepath
 	bytecode, err := bpfdagentinternal.GetBytecode(r.Client, &common.ByteCode)
```
anfredette, line 273: The reset of the functions above "should just work", but this one can fail if the authentication is incorrect.  As an optimization, this could be moved to the top of the function so it can fail early.

anfredette (commented):
The new reconcileProgram() is much more clear than the existing one. Other than the potential optimization, which could be done later, this LGTM.  I'll let Billy give it the final approval though.

bpfd-operator/controllers/bpfd-agent/common.go
```diff
@@ -234,109 +235,120 @@ func (r *ReconcilerCommon) createBpfProgram(ctx context.Context,
...
+	if err != nil {
+		return false, fmt.Errorf("failed to get bpfPrograms: %v", err)
 	}
 
-	isNodeSelected, err := isNodeSelected(&common.NodeSelector, ourNode.Labels)
+	// Get expected bpfPrograms for a *Program.
+	expectedPrograms, err := rec.expectedBpfPrograms(ctx)
 	if err != nil {
-		return false, fmt.Errorf("failed to check if node is selected: %v", err)
+		r.Logger.Error(err, "failed to create bpfPrograms")
```
anfredette, line 268:
```suggestion
		r.Logger.Error(err, "failed to get expected bpfPrograms")
```

bpfd-operator/controllers/bpfd-agent/common.go
```diff
@@ -234,109 +235,120 @@ func (r *ReconcilerCommon) createBpfProgram(ctx context.Context,
...
-			if err := r.Create(ctx, &prog, &opts); err != nil {
-				return true, fmt.Errorf("failed to create bpfProgram object: %v", err)
-			}
-			r.bpfPrograms[prog.Name] = prog
-			return false, nil
-		}
+	// Get existing bpfPrograms for a *Program.
+	existingPrograms, err := r.getExistingBpfProgs(ctx, program)
+	if err != nil {
+		return false, fmt.Errorf("failed to get bpfPrograms: %v", err)
```
anfredette, line 262:
```suggestion
		return false, fmt.Errorf("failed to get existing bpfPrograms: %v", err)
```

bpfd-operator/controllers/bpfd-agent/common.go
astoycos: Hrm good point I think this is broken... 

Basically what I wanted to do was ensure that we NEVER return and error for `Reconcile` because it will trigger a TON of instant reconciles and will not honor the retry period...basically just infinite looping most of the time :/ .

What this should express is the opposite... If we return an Error, print it and retry with the default retry duration (5 sec)

Let me try and clean this up in a new commit.

bpfd-operator/controllers/bpfd-agent/common.go
```diff
@@ -234,109 +235,120 @@ func (r *ReconcilerCommon) createBpfProgram(ctx context.Context,
...
-		if isBeingDeleted {
-			changed, err := r.removeFinalizer(ctx, &bpfProg, rec.getFinalizer())
+	// multiplex signals into kubernetes API actions
+	switch isBeingDeleted {
+	// Deletion of a *Program takes a few steps if there's existing bpfPrograms:
+	// 1. Reconcile the bpfProgram (take bpfd cleanup steps).
+	// 2. Remove any finalizers from the bpfProgram Object.
+	// 3. Update the condition on the bpfProgram to BpfProgCondUnloaded so the
+	//    operator knows it's safe to remove the parent Program Object, which
+	//	  is when the bpfProgram is automatically deleted by the owner-reference.
```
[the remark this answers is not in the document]
astoycos: Done

bpfd-operator/controllers/bpfd-agent/common.go
astoycos: PTAL at 7b71886fb1a03aa1cace7fdc1b0a44c9499a5933

astoycos:
Hrm 

```
#4 ERROR: failed to do request: Head "https://registry.fedoraproject.org/v2/fedora-minimal/manifests/latest": dial tcp 38.145.60.20:443: i/o timeout
```

not related to this PR

Billy99 (approved):
LGTM

PR #642, opened 2023-08-04
Proto idea

anfredette:


anfredette:
This pr was for discussion only.

Issue #640, opened 2023-08-04
bpfd-agent gets in a bad state when authentication fails


anfredette:
I attempted to load [bpfd.io_v1alpha1_xdp_pass_private_xdpprogram.yaml](https://github.com/bpfd-dev/bpfd/blob/main/bpfd-operator/config/samples/bpfd.io_v1alpha1_xdp_pass_private_xdpprogram.yaml), but it had an error.  The `kubectl create secret ` command given in the comments uses namespace `bpfd`, but the spec has namespace `default`.  As a result, the authentication fails and the image isn't pulled.

However, we say that it was loaded successfully anyway:
```
$ kubectl get xdpprograms.bpfd.dev 
NAME                         SECTIONNAME   NODESELECTOR
xdp-pass-private-all-nodes   pass          {}
$ kubectl get xdpprograms.bpfd.dev -o yaml 
apiVersion: v1
items:
- apiVersion: bpfd.dev/v1alpha1
  kind: XdpProgram
  metadata:
    annotations:
      kubectl.kubernetes.io/last-applied-configuration: |
        {"apiVersion":"bpfd.dev/v1alpha1","kind":"XdpProgram","metadata":{"annotations":{},"labels":{"app.kubernetes.io/name":"xdpprogram"},"name":"xdp-pass-private-all-nodes"},"spec":{"bytecode":{"image":{"imagepullsecret":{"name":"regcred","namespace":"default"},"url":"quay.io/bpfd-bytecode/xdp_pass_private:latest"}},"interfaceselector":{"primarynodeinterface":true},"nodeselector":{},"priority":0,"sectionname":"pass"}}
    creationTimestamp: "2023-08-04T15:08:30Z"
    deletionGracePeriodSeconds: 0
    deletionTimestamp: "2023-08-04T15:19:15Z"
    finalizers:
    - bpfd.dev.operator/finalizer
    generation: 2
    labels:
      app.kubernetes.io/name: xdpprogram
    name: xdp-pass-private-all-nodes
    resourceVersion: "8772"
    uid: c61f2850-2822-4017-83d0-99bd81b12982
  spec:
    bytecode:
      image:
        imagepullpolicy: IfNotPresent
        imagepullsecret:
          name: regcred
          namespace: default
        url: quay.io/bpfd-bytecode/xdp_pass_private:latest
    interfaceselector:
      primarynodeinterface: true
    nodeselector: {}
    priority: 0
    proceedon:
    - pass
    - dispatcher_return
[...]

PR #649, opened 2023-08-10
Handle and report bytecode selector errors

anfredette:
With the original bug, if there was a problem with GetBytecode(), the bpfProgram object would have been created, but the program would fail to load because GetBytecode() would fail and return.  We didn’t tell the user why it failed and you couldn’t delete the *program object because in order to do that, you need to call reconcileBpfdProgram(), which requires the bytecode interface.

Andrew’s recent pr changed the behavior so that the bpfProgram objects get created only after the GetBytecode() succeeds.  This is better because you don’t have the bpfProgram object that can’t be deleted, but because you don’t have the bpfProgram object, you don’t know why it failed unless you look at the agent logs.

Both versions still had a problem where if the secret gets deleted (or changed so it doesn’t work after a bpf program is loaded) you can’t unload programs that use it even though the bytecode  interface isn’t necessary for unloading programs.

This pr moves the GetBytecode() call into reconcileBpfdProgram(), and only calls it when we need to do a load.  This allows the bpfProgram object to get created and reports the error with a new condition called BpfProgCondBytecodeSelectorError.  It also allows you to delete the program.

This pr also changes the result returned from reconcileProgram to one of "Unchanged", "Updated", or "Requeue".

Fixes: issue #640

### DIFF
--- a/bpfd-operator/apis/v1alpha1/shared_types.go
+++ b/bpfd-operator/apis/v1alpha1/shared_types.go
@@ -237,9 +237,13 @@ const (
 	// BpfProgCondMapOwnerNotLoaded indicates that the eBPF program sharing a map with another
 	// eBPF program and that program is not loaded.
 	BpfProgCondMapOwnerNotLoaded BpfProgramConditionType = "MapOwnerNotLoaded"
+
 	// BpfProgCondByteCodeError indicates that an error occured when trying to
 	// process the bytecode selector.
 	BpfProgCondBytecodeSelectorError BpfProgramConditionType = "BytecodeSelectorError"
+
+	// None of the above conditions apply
+	BpfProgCondNone BpfProgramConditionType = "None"
 )
 
 // Condition is a helper method to promote any given BpfProgramConditionType to
@@ -304,6 +308,14 @@ func (b BpfProgramConditionType) Condition() metav1.Condition {
 			Status:  metav1.ConditionTrue,
 			Reason:  "bytecodeSelectorError",
 			Message: "There was an error processing the provided bytecode selector",
+		}
+
+	case BpfProgCondNone:
+		cond = metav1.Condition{
+			Type:    string(BpfProgCondNone),
+			Status:  metav1.ConditionTrue,
+			Reason:  "None",
+			Message: "None of the conditions apply",
 		}
 	}
 
--- a/bpfd-operator/apis/v1alpha1/shared_types.go
+++ b/bpfd-operator/apis/v1alpha1/shared_types.go
@@ -237,6 +237,9 @@ const (
 	// BpfProgCondMapOwnerNotLoaded indicates that the eBPF program sharing a map with another
 	// eBPF program and that program is not loaded.
 	BpfProgCondMapOwnerNotLoaded BpfProgramConditionType = "MapOwnerNotLoaded"
+	// BpfProgCondByteCodeError indicates that an error occured when trying to
+	// process the bytecode selector.
+	BpfProgCondBytecodeSelectorError BpfProgramConditionType = "BytecodeSelectorError"
 )
 
 // Condition is a helper method to promote any given BpfProgramConditionType to
@@ -293,6 +296,14 @@ func (b BpfProgramConditionType) Condition() metav1.Condition {
 			Status:  metav1.ConditionTrue,
 			Reason:  "mapOwnerNotLoaded",
 			Message: "BpfProgram map owner not loaded",
+		}
+
+	case BpfProgCondBytecodeSelectorError:
+		cond = metav1.Condition{
+			Type:    string(BpfProgCondBytecodeSelectorError),
+			Status:  metav1.ConditionTrue,
+			Reason:  "bytecodeSelectorError",
+			Message: "There was an error processing the provided bytecode selector",
 		}
 	}
 
--- a/bpfd-operator/controllers/bpfd-agent/common.go
+++ b/bpfd-operator/controllers/bpfd-agent/common.go
@@ -236,6 +236,36 @@ func (r *ReconcilerCommon) createBpfProgram(ctx context.Context,
 	return bpfProg, nil
 }
 
+type ReconcileResult uint8
+
+const (
+	// No changes were made to k8s objects, and rescheduling another reconcile
+	// is not necessary.
+	Unchanged ReconcileResult = 0
+	// Changes were made to k8s objects that we know will trigger another
+	// reconcile.  Calling code should return immediately to avoid multiple
+	// concurrent reconcile threads.
+	Updated ReconcileResult = 1
+	// A retry should be scheduled. This should only be used when "Updated"
+	// doesn't apply, but we want to trigger another reconcile anyway.  For
+	// example, there was a transient error. The calling code may continue
+	// reconciling other programs in it's list.
+	Requeue ReconcileResult = 2
+)
+
+func ReconcileResultToString(result ReconcileResult) string {
+	switch result {
+	case Unchanged:
+		return "Unchanged"
+	case Updated:
+		return "Updated"
+	case Requeue:
+		return "Requeue"
+	default:
+		return fmt.Sprintf("Unknown (%d)", result)
+	}
+}
+
 // reconcileProgram is called by ALL *Program controllers, and contains much of
 // the core logic for taking *Program objects, turning them into bpfProgram
 // object(s), and ultimately telling the custom controller types to load real
@@ -249,7 +279,7 @@ func reconcileProgram(ctx context.Context,
 	program client.Object,
 	common *bpfdiov1alpha1.BpfProgramCommon,
 	ourNode *v1.Node,
-	programMap map[string]*gobpfd.ListResponse_ListResult) (bool, error) {
+	programMap map[string]*gobpfd.ListResponse_ListResult) (ReconcileResult, error) {
 
 	// initialize reconciler state
 	r := rec.getRecCommon()
@@ -258,7 +288,7 @@ func reconcileProgram(ctx context.Context,
 	// OR if the *Program is being deleted.
 	isNodeSelected, err := isNodeSelected(&common.NodeSelector, ourNode.Labels)
 	if err != nil {
-		return false, fmt.Errorf("failed to check if node is selected: %v", err)
+		return Requeue, fmt.Errorf("failed to check if node is selected: %v", err)
 	}
 
 	isBeingDeleted := !program.GetDeletionTimestamp().IsZero()
@@ -267,7 +297,7 @@ func reconcileProgram(ctx context.Context,
 	// on this node.
 	existingPrograms, err := r.getExistingBpfProgs(ctx, program)
 	if err != nil {
-		return false, fmt.Errorf("failed to get existing bpfPrograms: %v", err)
+		return Requeue, fmt.Errorf("failed to get existing bpfPrograms: %v", err)
 	}
 
 	// Generate the list of BpfPrograms for this *Program. This handles the one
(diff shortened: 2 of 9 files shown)


bpfd-operator/apis/v1alpha1/shared_types.go
```diff
@@ -237,6 +237,9 @@ const (
 	// BpfProgCondMapOwnerNotLoaded indicates that the eBPF program sharing a map with another
 	// eBPF program and that program is not loaded.
 	BpfProgCondMapOwnerNotLoaded BpfProgramConditionType = "MapOwnerNotLoaded"
+	// BpfProgCondByteCodeError indicates that an error occured when trying to
```
astoycos, line 240: There's only a few failure modes here right? i.e something along the lines of 

1. Failure to pull the image (various errors)
2. Failure to find the file specified at PATH

Should we have two explicit condition types for this?

bpfd-operator/controllers/bpfd-agent/common.go
```diff
@@ -254,12 +254,6 @@ func reconcileProgram(ctx context.Context,
 	// initialize reconciler state
 	r := rec.getRecCommon()
```
astoycos, line 256: The only issue I can forsee by moving this down is that now we might pull for EVERY bpfProgram rather than once for all bpfPrograms

bpfd-operator/controllers/bpfd-agent/common.go
```diff
@@ -350,24 +342,27 @@ func reconcileProgram(ctx context.Context,
...
+				&common.ByteCode,
 				&prog,
 				isNodeSelected,
 				isBeingDeleted,
 				mapOwnerStatus,
 			)
 			if err != nil {
-				r.updateStatus(ctx, &prog, cond)
-				return true, fmt.Errorf("failed to reconcile bpfd program: %v", err)
+				if r.updateStatus(ctx, &prog, cond) {
```
astoycos, line 352: Wait I don't think we should change this we need to return `true` and the err so that the error is printed and we retry see the comment for this function 

```
If the function returns a retry boolean and error, the
// reconcile will be retried based on a default 5 second interval if the retry
// boolean is set to `true`.
```

Please leave as is

bpfd-operator/controllers/bpfd-agent/kprobe-program.go
```diff
@@ -213,6 +219,13 @@ func (r *KprobeProgramReconciler) reconcileBpfdProgram(ctx context.Context,
 		}
 
 		// otherwise load it
+		bytecode, err := bpfdagentinternal.GetBytecode(r.Client, bytecodeSelector)
+		if err != nil {
+			return bpfdiov1alpha1.BpfProgCondBytecodeSelectorError, fmt.Errorf("failed to process bytecode selector: %v", err)
```
astoycos, line 224: Hrm I know we might not always need the bytecode.. but can you just call this function and buildKprobeLoadRequest once up near the entry of reconcileBpfdProgram? I don't like having the code block duplicated, and IMO this whole thing should ALWAYS fail if we don't have access to the bytecode

If you really don't want to do that please just put it in an in-line anonymous and call it when needed.

astoycos (changes requested):
The comments not in common.go apply to all the individual controllers 👍

bpfd-operator/controllers/bpfd-agent/common.go
anfredette: I don't think it does that.  GetBytecode just builds the interface which get's sent to bpfd, which does the image pulling.  So, I think we'll build the interface each time, but I don't think it changes how many times we actually pull the image.

bpfd-operator/controllers/bpfd-agent/common.go
anfredette: I was seeing that when I always return true, the reconcile would sometimes take forever to complete.  I was thinking it would be better to only report the error the first time because on subsequent occurences, no state is changing and we can continue reconciling other programs.  In any case, I did have a question on this point, and I wanted to discuss it with you.  In particular, I wanted to better understand how you want this reconcile loop to work.  Can we talk tomorrow?

bpfd-operator/controllers/bpfd-agent/kprobe-program.go
anfredette: If we just call it once at the top and return if we get an error, then we can't delete the bpfProgram objects that are created.  I could call it at the top, remember whether there's an error, and then allow deletes, but only do loads if it was successful.

bpfd-operator/apis/v1alpha1/shared_types.go
anfredette: BTW, I think it would be good to be able to report these errors as well, but I think those are different failures and could be handled in another pr.  If you agree, I can open an issue for them.

bpfd-operator/controllers/bpfd-agent/common.go
astoycos: > I was seeing that when I always return true, the reconcile would sometimes take forever to complete.

well you should only return true on failure and in that case it _should_ take 5 seconds to reconcile (hoping bpfd is back up by then) 

Let's talk

the pattern we've followed here all boils down to https://github.com/bpfd-dev/bpfd/blob/main/bpfd-operator/controllers/bpfd-agent/kprobe-program.go#L147 where we are in the main `Reconcile` for the controller and decide to ALWAYS print the error and sometimes decide to retry (based on various scenarios)

bpfd-operator/controllers/bpfd-agent/kprobe-program.go
```diff
@@ -188,6 +177,23 @@ func (r *KprobeProgramReconciler) reconcileBpfdProgram(ctx context.Context,
...
+	bytecodeSelector *bpfdiov1alpha1.BytecodeSelector,
+	bpfProgram *bpfdiov1alpha1.BpfProgram,
+	isNodeSelected bool,
+	isBeingDeleted bool,
+	mapOwnerStatus *MapOwnerParamStatus) (bpfdiov1alpha1.BpfProgramConditionType, error) {
+
+	r.Logger.V(1).Info("Existing bpfProgram", "ExistingMaps", bpfProgram.Spec.Maps, "UUID", bpfProgram.UID, "Name", bpfProgram.Name, "CurrentKprobeProgram", r.currentKprobeProgram.Name)
+
+	id := string(bpfProgram.UID)
+
```
astoycos, line 196:
```suggestion
getAndBuild := func() { 
     		bytecode, err := bpfdagentinternal.GetBytecode(r.Client, bytecodeSelector)
		if err != nil {
			return bpfdiov1alpha1.BpfProgCondBytecodeSelectorError, fmt.Errorf("failed to process bytecode selector: %v", err) }
			
		loadRequest := r.buildKprobeLoadRequest(bytecode, id, bpfProgram, mapOwnerStatus.mapOwnerUuid)
}
```

Something like this is what I meant when I said "just put it in an in-line anonymous and call it when needed."

bpfd-operator/controllers/bpfd-agent/kprobe-program.go
astoycos: > If we just call it once at the top and return if we get an error, then we can't delete the bpfProgram objects that are created.

Hrm yeah that's an issue :( 

see below comment, it allows us to deduplicate that code block without adding a new function def

bpfd-operator/controllers/bpfd-agent/common.go
anfredette: I haven't confirmed this, but here's what I envision will happen if we always return an error:
1. Get list of programs
2. program with error is first on the list.
3. We hit the error 
4. We return the error
5. The reconcile loop quits without reconciling the rest of the programs (one of which may be the delete we're trying to do)
6. Another reconcile is scheduled for 5 seconds later.
7. Start back at #1 and repeat forever.

bpfd-operator/controllers/bpfd-agent/kprobe-program.go
anfredette: I still need to make this change.

bpfd-operator/controllers/bpfd-agent/common.go
```diff
@@ -236,6 +236,36 @@ func (r *ReconcilerCommon) createBpfProgram(ctx context.Context,
 	return bpfProg, nil
 }
 
+type ReconcileResult uint8
```
astoycos, line 239: This may apply to the operator code as well, although it's less complicated I think we'll still have the same enum values. If so this should live in /pkg/internal rather than nested in the bpfd-agent

bpfd-operator/controllers/bpfd-agent/tc-program.go
```diff
@@ -186,41 +188,50 @@ func (r *TcProgramReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 		}
 
-		retry, err := reconcileProgram(ctx, r, r.currentTcProgram, &r.currentTcProgram.Spec.BpfProgramCommon, r.ourNode, existingPrograms)
+		result, err := reconcileProgram(ctx, r, r.currentTcProgram, &r.currentTcProgram.Spec.BpfProgramCommon, r.ourNode, existingPrograms)
 		if err != nil {
-			r.Logger.Error(err, "Reconciling TcProgram Failed", "TcProgramName", r.currentTcProgram.Name, "Retrying", retry)
-			return ctrl.Result{Requeue: retry, RequeueAfter: retryDurationAgent}, nil
+			r.Logger.Error(err, "Reconciling TcProgram Failed", "TcProgramName", r.currentTcProgram.Name, "ReconcileResult", ReconcileResultToString(result))
+		}
+		switch result {
```
astoycos, line 195: This exact logic is duplicated for every bpfd-agent controller right? If so is it easy to internalize into a function in common.go

anfredette: I know.  It bothered me too, but was trying not to do too much surgery.  However, I'll take another look.

bpfd-operator/controllers/bpfd-agent/common.go
anfredette: I'll take a look at the operator code.

anfredette: Maybe, but I'm not sure. The current operator reconcile code is a bit different in that it looks like we just reconcile the one *Program that we were notified about instead of all of them every time.  Also, reconcileBpfProgram returns the reconcile.Result directly which we return immediately.  However, I moved ReconcileResult to bpfd-operator/internal/constants.go (I assume that's what you meant) because it seems like a good central location and makes it easier if we want to use it elsewhere in the future.  I also made the ToString function a method on ReconcileResult.

bpfd-operator/apis/v1alpha1/shared_types.go
anfredette: https://github.com/bpfd-dev/bpfd/issues/659

bpfd-operator/controllers/bpfd-agent/kprobe-program.go
anfredette: Done.

bpfd-operator/controllers/bpfd-agent/tc-program.go
anfredette: I'm not sure if you're just referring to the switch statement or the whole function.  While the switch statement is duplicated, it's controlling the flow of the loop, so I don't think I can just offload that part.  A macro would work, but Go doesn't have them.  I did summarize the comments because I was pretty much duplicating the comments on the result constants.  Actually, almost all the code in this whole file is the same for all the program types except for the types.  Do you have any ideas?

bpfd-operator/controllers/bpfd-agent/kprobe-program.go
anfredette: Done.

bpfd-operator/controllers/bpfd-agent/common.go
astoycos: ++ thanks!

astoycos (approved):
LGTM

thanks!

PR #656, opened 2023-08-14
Use kernel Id as the one true program identifier + cleanup a bunch of techdebt

### LINKED ISSUE #580: kernel ids - the one true program identifier
bpfd currently allocates a UUID for every loaded program.
this serves a few purposes:

1. It gives us a sentinel value that we can use for storing things on the filesystem
2. It helps us stay in sync when bpfd is driven by an orchestrator
3. Our UUIDs *should* be unique across the cluster

However, the kernel has it's own program/link/map identifiers.
- These are a u32, you provided by the kernel at load/create time.
- They are locally significant

We should preferably have only one unique means of identifying programs when using bpfd/bpfctl.
This should be the kernel ID for consistency with other tools, like bpftool.

Therefore, any UUID from an orchestrator should get stored in the program metadata instead.
Map pin paths will no longer be easily used from K8s - this now needs to be fixed in the design of #522

astoycos:
fixes #580


bpfd/src/multiprog/tc.rs
```diff
@@ -191,35 +202,51 @@ impl TcDispatcher {
...
 
-                for (name, value) in &v.data.global_data {
-                    bpf.set_global(name, value.as_slice(), true);
+                match &v.data.global_data {
+                    Some(d) => {
+                        for (name, value) in d {
+                            bpf.set_global(name, value.as_slice(), true);
+                        }
+                    }
+                    None => debug!("no global data to set for xdpProgram"),
```
astoycos, line 217:
```suggestion
                    None => debug!("no global data to set for Tc Program"),
```

tests/integration-test/src/tests/basic.rs
```diff
@@ -296,7 +296,11 @@ fn test_load_unload_kretprobe() {
 
 #[integration_test]
 fn test_pull_bytecode() {
-    std::fs::remove_dir_all(BYTECODE_IMAGE_CONTENT_STORE).unwrap();
+    let path = PathBuf::from(BYTECODE_IMAGE_CONTENT_STORE);
```
anfredette, line 299: I hit that one too, and fixed it like this in my last pr:
    let _ = std::fs::remove_dir_all(BYTECODE_IMAGE_CONTENT_STORE);

astoycos: Hrm I don't want to remove the error though... i.e we should know it failed unless it was because the dir already existed

anfredette: True.  Your solution will cover the case where the dir exists, but the delete fails anyway.

anfredette: I'll just drop that change from my pr.  I hit the issue because I ran `sudo ./scripts/setup.sh uninstall`, which deletes that dir, before running int test.

astoycos:
So this is mostly ready for review @Billy99 can you start by checking out the proto API changes? 

I'm failing some k8s integration test issues which I'm still debugging

bpfd/src/bpf.rs
```diff
@@ -1143,134 +908,154 @@ impl BpfManager {
...
-// of another program that map_owner_uuid references.
-// This function also returns a bool, which indicates if the input UUID
-// is the owner of the map (map_owner_uuid is not set) or not (map_owner_uuid
-// is set so the eBPF program is referencing another eBPF programs maps).
-fn get_map_index(id: Uuid, map_owner_uuid: Option<Uuid>) -> (bool, Uuid) {
-    if let Some(uuid) = map_owner_uuid {
-        (false, uuid)
-    } else {
-        (true, id)
+fn sort_dispatcher_programs(mut programs: Vec<Program>) -> Vec<Program> {
```
dave-tucker, line 952: This is not the algorithm we use to sort programs :(
See [here](
https://github.com/xdp-project/xdp-tools/blob/master/lib/libxdp/protocol.org#loading-and-attaching-component-programs) where our previous sort relied on the `Metadata` struct which contained the name + whether a program was already loaded or not. It's PartialEq impl, used by sort, did the hard work for us. Size, tag and load_time should also have been added.

bpfd/src/command.rs
```diff
@@ -231,176 +165,51 @@ impl TryFrom<AyaProgInfo> for KernelProgramInfo {
...
+#[derive(Debug, Serialize, Deserialize, Clone)]
 pub(crate) struct ProgramData {
+    // known at load time, set by user
+    pub(crate) name: String,
     pub(crate) location: Location,
-    pub(crate) section_name: String,
-    pub(crate) global_data: HashMap<String, Vec<u8>>,
-    pub(crate) owner: String,
-    pub(crate) map_owner_uuid: Option<Uuid>,
+    pub(crate) uuid: Option<Uuid>,
```
dave-tucker, line 175: Per comment on proto this should be a `HashMap<String, String>: metadata`

bpfd/src/command.rs
```diff
@@ -231,176 +165,51 @@ impl TryFrom<AyaProgInfo> for KernelProgramInfo {
...
+    pub(crate) map_owner_id: Option<u32>,
+
+    // populated after load
     pub(crate) kernel_info: Option<KernelProgramInfo>,
+    pub(crate) map_pin_path: Option<PathBuf>,
+    pub(crate) map_used_by: Option<Vec<u32>>,
+}
+
+// Only compare fields which are known at load time
+impl PartialEq for ProgramData {
```
dave-tucker, line 186: Why wouldn't we want to include fields that are known after load also?

bpfd/src/command.rs
```diff
@@ -413,10 +222,10 @@ impl ProgramData {
                         // If section name isn't provided and we're loading from a container
                         // image use the section name provided in the image metadata, otherwise
                         // always use the provided section name.
-                        let provided_sec_name = self.section_name.clone();
+                        let provided_sec_name = self.name.clone();
```
dave-tucker, line 225: Not a section name anymore.

bpfd/src/command.rs
```diff
@@ -565,34 +495,48 @@ impl Program {
             Program::Tracepoint(p) => p.data.kernel_info = Some(info),
             Program::Kprobe(p) => p.data.kernel_info = Some(info),
             Program::Uprobe(p) => p.data.kernel_info = Some(info),
+            // has to be set at variant creation
```
dave-tucker, line 498: Is it a noop? Or an error?
Feels to me like it should probably be an error.

bpfd/src/command.rs
```diff
@@ -565,34 +495,48 @@ impl Program {
...
-    pub(crate) fn save(&self, uuid: Uuid) -> Result<(), anyhow::Error> {
-        let path = format!("{RTDIR_PROGRAMS}/{uuid}");
+    pub(crate) fn get_kernel_info(self) -> Option<KernelProgramInfo> {
+        match self {
+            Program::Xdp(p) => p.data.kernel_info,
+            Program::Tc(p) => p.data.kernel_info,
+            Program::Tracepoint(p) => p.data.kernel_info,
+            Program::Kprobe(p) => p.data.kernel_info,
+            Program::Uprobe(p) => p.data.kernel_info,
+            // has to be set at variant creation
```
dave-tucker, line 510: Comment is misleading.

bpfd/src/command.rs
```diff
@@ -601,37 +545,81 @@ impl Program {
 
     pub(crate) fn if_index(&self) -> Option<u32> {
         match self {
-            Program::Xdp(p) => Some(p.info.if_index),
+            Program::Xdp(p) => p.if_index,
```
dave-tucker, line 548: We should probably just use `_ = > None` to avoid having to list out every enum variant that doesn't use an if_index here.

bpfd/src/command.rs
```diff
@@ -601,37 +545,81 @@ impl Program {
...
             Program::Uprobe(_) => None,
+            Program::Unsupported(_) => None,
+        }
+    }
+
+    pub(crate) fn set_if_index(&mut self, if_index: u32) {
+        match self {
+            Program::Xdp(p) => p.if_index = Some(if_index),
+            Program::Tc(p) => p.if_index = Some(if_index),
+            _ => (),
```
dave-tucker, line 561: Yep, same pattern in the fn above please.

bpfd/src/command.rs
```diff
@@ -601,37 +545,81 @@ impl Program {
...
-            Program::Xdp(p) => Some(p.info.if_name.clone()),
+            Program::Xdp(p) => Some(p.iface.clone()),
             Program::Tracepoint(_) => None,
-            Program::Tc(p) => Some(p.info.if_name.clone()),
+            Program::Tc(p) => Some(p.iface.clone()),
             Program::Kprobe(_) => None,
             Program::Uprobe(_) => None,
+            Program::Unsupported(_) => None,
+        }
+    }
```
dave-tucker, line 574: And again.

bpfd/src/command.rs
```diff
@@ -601,37 +545,81 @@ impl Program {
...
+
+    pub(crate) fn priority(&self) -> Option<i32> {
```
dave-tucker, line 576: and again.

bpfd/src/command.rs
```diff
@@ -601,37 +545,81 @@ impl Program {
...
         match self {
             Program::Xdp(_) => None,
             Program::Tracepoint(_) => None,
-            Program::Tc(p) => Some(p.info.direction),
+            Program::Tc(p) => Some(p.direction),
             Program::Kprobe(_) => None,
             Program::Uprobe(_) => None,
+            Program::Unsupported(_) => None,
+        }
+    }
```
dave-tucker, line 596: ok we're getting a little carried away on providing accessors here.
if it's available in a majority of cases, cool.
but for cases like this where it's only applicable to one prog type we might as well make callers pattern match or if let.

e.g

```
if let Program::Tc(p) = program {
    p.direction
}
```

dave-tucker (changes requested):
I'd really like to see this broken into smaller PRs.
The refactoring could be done seperately to the API changes and would be easier to review.
As it stands:
- The protobuf changes need work
- I'm not confident the refactor won't break the XDP/TC sorting, which highlights that we need for testing there.
- I've seen one sneaky change to prep for the map API work that I don't think needed to be here, there could be more.

These are all welcome changes, but we need to be mindful of how we make them.

proto/bpfd.proto
```diff
@@ -25,9 +25,9 @@ message LoadRequestCommon {
     } ;
     string section_name = 3;
     uint32 program_type = 4;
-    optional string id = 5;
+    optional string uuid = 5;
     map<string, bytes> global_data = 6;
-    optional string map_owner_uuid = 7;
+    optional uint32 map_owner_id = 7;
```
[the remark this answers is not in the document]
Billy99: Agree that this is how it should be done, but we have done other API changes (proto changes) without being backwards compatible. Should we say from Release 0.3 and on we follow the programming-guides or start it mid release?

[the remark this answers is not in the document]
dave-tucker: have we made incompatible proto changes between 0.2 and 0.3?
we need to follow the guidelines at some point 😆 no strong preference on when though.

astoycos:
CI _should_ be good now, working on responding to @dave-tucker's comments now

Adding hold so that this doesn't merge pointing to my aya fork

bpfctl/src/main.rs
```diff
@@ -321,294 +327,245 @@ struct GlobalArg {
...
+        ]);
+        if r.map_owner_id.is_none() {
+            table.add_row(vec!["Map Owner UUID:", "None"]);
         } else {
-            format!(
-                r#"Map Owner UUID:                     {}"#,
-                r.map_owner_uuid.clone()
-            )
+            table.add_row(vec![
+                "Map Owner UUID:",
```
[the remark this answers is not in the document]
astoycos: Nope I missed this

bpfctl/src/main.rs
```diff
@@ -915,44 +873,44 @@ async fn execute_request(command: &Commands, channel: Channel) -> anyhow::Result
         }
 
         Commands::Unload(l) => {
-            let request = tonic::Request::new(UnloadRequest {
-                id: l.id.to_string(),
-            });
+            let request = tonic::Request::new(UnloadRequest { id: l.id });
```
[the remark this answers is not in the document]
astoycos: Hrm I __think__ I'm following you here :laughing let me take a stab

proto/bpfd.proto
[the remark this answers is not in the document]
astoycos: Yes we have... I'd argue lets NOT guarentee compat until we have at least SOME stability... trying to do it now would lead to a bunch of issues... Especially with https://github.com/bpfd-dev/bpfd/pull/641 coming in right after this

bpfctl/src/main.rs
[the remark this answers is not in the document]
astoycos: Wait so I don't actually follow you here, were' already using `tonic::Request` do you mean we should actually be using the response here? 

OR do you mean we should be setting up the client using the built-in methods like 

```
let mut client = BpfdClient::connect("http://[::1]:50051").await?;
```

bpfctl/src/main.rs
```diff
@@ -9,8 +9,11 @@ use bpfd_api::{
     config::{self, Config},
     util::directories::*,
     v1::{
-        list_response::{self, list_result::Location},
-        load_request::{self, AttachInfo},
+        list_response::{
+            self,
```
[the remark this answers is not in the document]
astoycos: Good point! ... Fixed

bpfctl/src/main.rs
```diff
@@ -321,294 +327,245 @@ struct GlobalArg {
...
+        match r.location.clone() {
+            Some(l) => match l {
+                Location::Image(i) => {
+                    table.add_row(vec!["Image URL:", &i.url]);
+                    table.add_row(vec!["Pull Policy:", &format!{ "{}", TryInto::<ImagePullPolicy>::try_into(i.image_pull_policy)?}]);
+                }
+                Location::File(p) => {
+                    table.add_row(vec!["Path:", &p]);
+                }
+                _ => panic!("TODO(astoycos) remove NoLocation from API"),
```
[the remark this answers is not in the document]
astoycos: Yeah let me tackle this

proto/bpfd.proto
```diff
@@ -25,9 +25,9 @@ message LoadRequestCommon {
     } ;
     string section_name = 3;
     uint32 program_type = 4;
-    optional string id = 5;
+    optional string uuid = 5;
```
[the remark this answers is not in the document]
astoycos: ++ I'm alright with that!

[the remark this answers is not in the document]
astoycos: The biggest negative would be that it's another peice of info that client's need to understand that's not conveyed by the API... i.e now they have to know the "bpfd.dev/uuid" is special.

proto/bpfd.proto
```diff
@@ -79,23 +79,24 @@ message LoadRequest {
...
-    string id = 1;
+    uint32 id = 1;
 }
 
 message UnloadResponse {}
 
 message ListRequest {
     optional uint32 program_type = 1;
     optional bool bpfd_programs_only = 2;
+    optional string uuid = 3;
```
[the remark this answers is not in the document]
astoycos: > Why do we need this? You will only have 1 result for UUID

As of now we treat uuid as just an additional piece of metadata... so if client's mess up and add two programs with the same UUID this _could_ end up being >1 results

> , which is a Get no?

No not necessarily... I don't think we should allow GET to be called with anything other than a program ID.... Let's make "querying by label, uuid etc" Always be a list so that we don't have to deal with that complexity internally

bpfctl/src/main.rs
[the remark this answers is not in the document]
dave-tucker: I mean `s/tonic::Request::new/Request/g`

proto/bpfd.proto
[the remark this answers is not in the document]
dave-tucker: Define client in that statement 😏 
If you mean `bpfctl` - it doesn't need to understand the semantics of any of the K/V pairs. Just read/write them.

If you mean bpfd-agent/bpfd-operator, it can define it's own semantics on whatever it writes.
It could use the key "kittens" to store UUIDs if it wanted, although I wouldn't recommend it 😆

proto/bpfd.proto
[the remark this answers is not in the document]
dave-tucker: I think going for `list by metadata` is fine. so long as it's generic and we're not special casing uuid.
e.g I can search by any random piece of metadata that fits in a map<string, string>

bpfctl/src/main.rs
```diff
@@ -50,14 +53,15 @@ enum Commands {
     /// Get a program's metadata by kernel id.
     Get {
         /// An eBPF program's kernel id.
-        kernel_id: u32,
+        id: u32,
```
Billy99, line 56: Line 49:
```
-        /// Unload an eBPF program using the UUID.
+        /// Unload an eBPF program by id.
```

Line 51:
```
-        /// List all eBPF programs loaded via bpfd.
+        /// List all eBPF programs.
```

bpfctl/src/main.rs
```diff
@@ -105,9 +112,9 @@ struct LoadFileArgs {
     /// Only used when multiple eBPF programs need to share a map. If a map is being
     /// shared with another eBPF program, the eBPF program that created the map can not
     /// be unloaded until all eBPF programs referencing the map are unloaded.
-    /// Example: --map-owner-uuid 989958a5-b47b-47a5-8b4c-b5962292437d
+    /// Example: --map-owner-id 63178
```
Billy99, line 115: Line 111 and 146:
```
-    /// Optional: UUID of loaded eBPF program this eBPF program will share a map with.
-    /// Only used when multiple eBPF programs need to share a map. If a map is being
-    /// shared with another eBPF program, the eBPF program that created the map can not
-    /// be unloaded until all eBPF programs referencing the map are unloaded.

+    /// Optional: Id of loaded eBPF program this eBPF program will share a map with.
+    /// Only used when multiple eBPF programs need to share a map.
```

bpfctl/src/main.rs
```diff
@@ -71,6 +75,10 @@ struct ListArgs {
     #[clap(short, long, verbatim_doc_comment, hide_possible_values = true)]
     program_type: Option<ProgramType>,
 
+    // Optional: List programs with a specified uuid
```
Billy99, line 78:
```suggestion
    /// Optional: List programs with a specified uuid
```

Billy99 (commented):
Some bpfctl comments

astoycos (commented):
Vscode GH editor has me do this as a review? 🤷‍♂️

bpfctl/src/main.rs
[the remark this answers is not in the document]
astoycos: Still need to do this

proto/bpfd.proto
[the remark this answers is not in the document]
astoycos: Done!

proto/bpfd.proto
```diff
@@ -79,23 +79,24 @@ message LoadRequest {
 };
 
 message LoadResponse {
-    string id = 1;
+    uint32 id = 1;
```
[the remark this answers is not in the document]
astoycos: see above conversation

proto/bpfd.proto
```diff
@@ -79,23 +79,24 @@ message LoadRequest {
...
 message ListRequest {
     optional uint32 program_type = 1;
     optional bool bpfd_programs_only = 2;
+    optional string uuid = 3;
 }
 
 message ListResponse {
   message ListResult {
-    optional string id = 1;
+    optional string uuid = 1;
```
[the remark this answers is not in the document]
astoycos: Done

proto/bpfd.proto
```diff
@@ -104,7 +105,7 @@ message ListResponse {
     } ;
     uint32 program_type = 6;
     map<string, bytes> global_data = 7;
-    string map_owner_uuid = 8;
+    optional uint32 map_owner_id = 8;
```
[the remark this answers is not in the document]
astoycos: see above

bpfd/src/multiprog/tc.rs
```diff
@@ -52,21 +51,25 @@ impl TcDispatcher {
...
         if_name: String,
-        programs: &mut [(Uuid, Program)],
+        programs: &mut [Program],
         revision: u32,
         old_dispatcher: Option<Dispatcher>,
     ) -> Result<TcDispatcher, BpfdError> {
         debug!("TcDispatcher::new() for if_index {if_index}, revision {revision}");
-        let mut extensions: Vec<(&mut Uuid, &mut TcProgram)> = programs
+        // Cast program list to TcPrograms.
+        // TODO (astoycos) Maybe in a step earlier?
```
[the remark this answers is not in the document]
astoycos: Still need to possibly remove this TODO

bpfd/src/command.rs
astoycos: Done

bpfd/src/command.rs
```diff
@@ -190,7 +124,7 @@ impl std::fmt::Display for Direction {
 
 /// KernelProgramInfo stores information about ALL bpf programs loaded
 /// on a system.
-#[derive(Serialize, Deserialize, Debug, Clone)]
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
```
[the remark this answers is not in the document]
astoycos: see other comment on ProgramData

bpfd/src/command.rs
```diff
@@ -28,100 +28,34 @@ type Responder<T> = oneshot::Sender<T>;
...
-    pub(crate) map_owner_uuid: Option<Uuid>,
-    pub(crate) fn_name: Option<String>,
-    pub(crate) offset: u64,
-    pub(crate) target: String,
-    pub(crate) retprobe: bool,
-    pub(crate) pid: Option<i32>,
-    pub(crate) _namespace: Option<String>,
-    pub(crate) username: String,
-    pub(crate) responder: Responder<Result<Uuid, BpfdError>>,
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
```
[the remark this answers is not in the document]
astoycos: See other comment on ProgramData

bpfd/src/command.rs
astoycos: decided to panic here instead of returning an error

bpfd/src/command.rs
```diff
@@ -484,77 +411,84 @@ impl Program {
...
         }
     }
 
     pub(crate) fn set_position(&mut self, pos: Option<usize>) {
         match self {
-            Program::Xdp(p) => p.info.current_position = pos,
-            Program::Tc(p) => p.info.current_position = pos,
+            Program::Xdp(p) => p.current_position = pos,
+            Program::Tc(p) => p.current_position = pos,
             Program::Tracepoint(_) => (),
```
astoycos, line 488: Still need to fix this up

bpfctl/src/main.rs
astoycos: Superseded by metadata selector

bpfctl/src/main.rs
```diff
@@ -71,6 +75,10 @@ struct ListArgs {
...
+    #[clap(short, long, verbatim_doc_comment)]
+    uuid: Option<String>,
+
     // Optional: List all programs
```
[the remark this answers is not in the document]
astoycos: Still need to do

bpfctl/src/main.rs
astoycos: Done

astoycos:
Thanks for the reviews @dave-tucker @Billy99 I tackled as much as I could today/tonight if I missed anything I tried to let you know in the comments, please take another pass when you get the chance.

Additionally here are the aya changes we'll need to support this -> https://github.com/aya-rs/aya/compare/main...astoycos:map_pin

bpfctl/src/main.rs
astoycos: TODO

bpfd/src/bpf.rs
dave-tucker: That links to an empty playground, you needed the permalink 😉 
I'm not advocating that you do the work of fixing the sort to be libxdp compatible - that's on me.
I'd only ask that you don't change the sorting algorithm in this PR and maintain the old behaviour.

bpfd/src/multiprog/xdp.rs
```diff
@@ -158,57 +161,94 @@ impl XdpDispatcher {
             .program_mut(self.progam_name.clone().unwrap().as_str())
             .unwrap()
             .try_into()?;
-        extensions.sort_by(|(_, a), (_, b)| a.info.current_position.cmp(&b.info.current_position));
-        for (i, (k, v)) in extensions.iter_mut().enumerate() {
-            if v.info.metadata.attached {
-                let mut ext = Extension::from_pin(format!("{RTDIR_FS}/prog_{k}"))?;
+        // astoycos: we just sorted this.
+        // extensions.sort_by(|(_, a), (_, b)| a.info.current_position.cmp(&b.info.current_position));
```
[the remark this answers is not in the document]
dave-tucker: Sure. It's just this code is a little fragile and we don't have any test coverage on sort ordering.
We should manually test that this works for both the add/delete cases before merging.

bpfd/src/multiprog/tc.rs
```diff
@@ -177,11 +180,19 @@ impl TcDispatcher {
             .unwrap()
             .try_into()?;
 
-        extensions.sort_by(|(_, a), (_, b)| a.info.current_position.cmp(&b.info.current_position));
-
-        for (i, (k, v)) in extensions.iter_mut().enumerate() {
-            if v.info.metadata.attached {
-                let mut ext = Extension::from_pin(format!("{RTDIR_FS}/prog_{k}"))?;
+        // astoycos: Already sorted
```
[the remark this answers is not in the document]
dave-tucker: That's not actually sorting the programs though. Notice the `collect()`?
We're sorting the collected Vec of mut Programs and setting the current ID.

bpfctl/src/main.rs
Billy99: Nit: This was existing, but since you were there, this sentence needs to be removed:

```
If a map is being shared with another eBPF program, the eBPF program that created the map can not be unloaded until all eBPF programs referencing the map are unloaded.
```

bpfd/src/bpf.rs
```diff
@@ -354,6 +439,15 @@ impl BpfManager {
                 let kprobe: &mut KProbe = raw_program.try_into()?;
                 kprobe.load()?;
 
+                p.set_kernel_info(kprobe.program_info()?.try_into()?);
```
Billy99, line 442: Why do we need to call `set_kernel_info()` twice for kprobe and uprobe? For kprobe, it's called here and line 460. Similar pattern in uprobe.

bpfd/src/rpc.rs
```diff
@@ -273,130 +200,146 @@ impl Loader for BpfdLoader {
...
+                                    None => {
+                                        // skip programs not owned by bpfd
+                                        if request.get_ref().bpfd_programs_only() {
+                                            continue;
+                                        }
+                                        None
+                                    }
+                                };
+
+                                reply_entry.attach_info = match r.clone() {
```
dave-tucker, line 286:
```suggestion
                                reply_entry.attach_info = match r {
```

dave-tucker (changes requested):
Getting there. Some more comments.

astoycos:
Closed in favor of #701 

#701 pulls many of the API changes (since they were already reviewed here)  as well as the examples + operator changes from this PR

PR #658, opened 2023-08-22
bump aya up to main

astoycos:
use aya:main for the moment instead of our own fork. Additionally make some fixes caused by this update:

- fixup extension api

- update how we use the program_info API due to final version of the patch that landed in aya.

- With recent Aya changes we actually use the name of the bpf_function as the "section-name" fix this up after we started pointing back to aya-main.

We'll be reverting back shorty since I'll need some custom aya patches in #656 but this is a nice middle step


tests/integration-test/bpf/build_push_images.sh
```diff
@@ -24,7 +24,7 @@ docker push quay.io/bpfd-bytecode/tc_pass
 
 docker build \
  --build-arg PROGRAM_NAME=tracepoint \
- --build-arg SECTION_NAME=sys_enter_openat \
+ --build-arg SECTION_NAME=enter_openat \
```
anfredette, line 27: Doesn't this need to match the section name defined in the following?
https://github.com/bpfd-dev/bpfd/blob/main/tests/integration-test/bpf/tp_openat.bpf.c#L15

If so, one or the other will need to change and get rebuilt and pushed.

astoycos: Nope with https://github.com/aya-rs/aya/pull/413 it's now based on the actual function name 

( I already built and pushed the integration test images)

tests/integration-test/bpf/build_push_images.sh
```diff
@@ -24,7 +24,7 @@ docker push quay.io/bpfd-bytecode/tc_pass
...
  --build-arg PROGRAM_TYPE=tracepoint \
  --build-arg BYTECODE_FILENAME=tp_openat.bpf.o \
  -f ../../../packaging/container-deployment//Containerfile.bytecode \
```
astoycos, line 30: Add ability to make tag configurable here

Billy99 (approved):
LGTM

PR #661, opened 2023-08-23
Add setting of global data to sample yaml files

anfredette:
Note: This was just done for the sample programs that have global variables defined.


astoycos (approved):
LGTM

PR #691, opened 2023-09-05
separate dispatchers from programs in BpfManager

astoycos:
separate dispatchers from programs in BpfManager

Separate dispatchers and programs into their own unique types
since they currently have very different lifetime requirements
in the codebase today.

This includes adding getters/setters and helper traits for each
type.

Ultimately this now enables us to pass around a mutable reference
to a program throughout a program operation (load/delete/reload/etc)
which makes the entire cycle a bit more efficient now that we need
to update a program's kernel info AFTER load.

### DIFF
--- a/bpfd/src/bpf.rs
+++ b/bpfd/src/bpf.rs
@@ -48,11 +48,137 @@ const MAPS_MODE: u32 = 0o0660;
 
 pub(crate) struct BpfManager {
     config: Config,
-    dispatchers: HashMap<DispatcherId, Dispatcher>,
-    programs: HashMap<Uuid, Program>,
+    dispatchers: DispatcherMap,
+    programs: ProgramMap,
     maps: HashMap<Uuid, BpfMap>,
     commands: Receiver<Command>,
     image_manager: Sender<ImageManagerCommand>,
+}
+
+pub(crate) struct ProgramMap {
+    programs: HashMap<Uuid, Program>,
+}
+
+impl ProgramMap {
+    fn new() -> Self {
+        ProgramMap {
+            programs: HashMap::new(),
+        }
+    }
+
+    fn insert(&mut self, id: Uuid, prog: Program) -> Option<Program> {
+        self.programs.insert(id, prog)
+    }
+
+    fn remove(&mut self, id: &Uuid) -> Option<Program> {
+        self.programs.remove(id)
+    }
+
+    fn get_mut(&mut self, id: &Uuid) -> Option<&mut Program> {
+        self.programs.get_mut(id)
+    }
+
+    fn get(&self, id: &Uuid) -> Option<&Program> {
+        self.programs.get(id)
+    }
+
+    fn contains_key(&self, id: &Uuid) -> bool {
+        self.programs.contains_key(id)
+    }
+
+    fn programs_mut<'a>(
+        &'a mut self,
+        program_type: &'a ProgramType,
+        if_index: &'a Option<u32>,
+        direction: &'a Option<Direction>,
+    ) -> impl Iterator<Item = (&'a Uuid, &'a mut Program)> {
+        self.programs.iter_mut().filter(|(_, p)| {
+            p.kind() == *program_type && p.if_index() == *if_index && p.direction() == *direction
+        })
+    }
+
+    // Sets the positions of programs that are to be attached via a dispatcher.
+    // Positions are set based on order of priority. Ties are broken based on:
+    // - Already attached programs are preferred
+    // - Program name. Lowest lexical order wins.
+    fn set_program_positions(
+        &mut self,
+        program_type: ProgramType,
+        if_index: Option<u32>,
+        direction: Option<Direction>,
+    ) {
+        let mut extensions = self
+            .programs
+            .iter_mut()
+            .filter_map(|(k, v)| {
+                if v.kind() == program_type {
+                    if v.if_index() == if_index && v.direction() == direction {
+                        Some((k, v))
+                    } else {
+                        None
+                    }
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<(&Uuid, &mut Program)>>();
+
+        extensions.sort_by_key(|(_, b)| {
+            (
+                b.priority(),
+                b.attached().unwrap(),
+                b.data()
+                    .expect("All Bpfd programs should have ProgramData")
+                    .name()
+                    .to_owned(),
+            )
+        });
+        for (i, (_, v)) in extensions.iter_mut().enumerate() {
+            v.set_position(Some(i));
+        }
+    }
+
+    fn kernel_infos(&self) -> impl Iterator<Item = (u32, &Program)> {
+        self.programs.values().map(|p| {
+            let kernel_info = p
+                .data()
+                .expect("All Bpfd programs should have ProgramData")
+                .kernel_info()
+                .expect("Loaded Bpfd programs should have kernel information");
+
+            (kernel_info.id, p)
+        })
+    }
+}
+
+pub(crate) struct DispatcherMap {
+    dispatchers: HashMap<DispatcherId, Dispatcher>,
+}
+
+impl DispatcherMap {
+    fn new() -> Self {
+        DispatcherMap {
+            dispatchers: HashMap::new(),
+        }
+    }
+
+    fn remove(&mut self, id: &DispatcherId) -> Option<Dispatcher> {
+        self.dispatchers.remove(id)
+    }
+
+    fn insert(&mut self, id: DispatcherId, dis: Dispatcher) -> Option<Dispatcher> {
+        self.dispatchers.insert(id, dis)
+    }
+
+    /// Returns the number of extension programs currently attached to the dispatcher that
+    /// would be used to attach the provided [`Program`].
+    fn attached_programs(&self, did: &DispatcherId) -> usize {
+        if let Some(d) = self.dispatchers.get(did) {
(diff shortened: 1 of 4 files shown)


bpfd/src/bpf.rs
```diff
@@ -41,18 +41,145 @@ const MAPS_MODE: u32 = 0o0660;
...
+        self.programs
+            .iter()
+            .filter(|(_, p)| {
+                if p.kind() == program.kind() {
+                    p.if_index() == program.if_index() && p.direction() == program.direction()
+                } else {
+                    false
+                }
+            })
+            .collect::<HashMap<_, _>>()
```
dave-tucker, line 86: You can avoid the `collect().len()` here and use `count()`

bpfd/src/bpf.rs
```diff
@@ -41,18 +41,145 @@ const MAPS_MODE: u32 = 0o0660;
...
+            .len()
+    }
+
+    fn collect_programs(
+        &mut self,
+        program_type: ProgramType,
+        if_index: Option<u32>,
+        direction: Option<Direction>,
+    ) -> Vec<(&Uuid, &mut Program)> {
```
dave-tucker, line 95: In a local branch I played around with the return type being `impl Iterator<Item = (&Uuid, &mut Program)>` which lets us avoid allocating (via `collect()) here.

bpfd/src/bpf.rs
```diff
@@ -41,18 +41,145 @@ const MAPS_MODE: u32 = 0o0660;
...
+            .map(|p| {
+                let kernel_info = p
+                    .data()
+                    .expect("All Bpfd programs should have ProgramData")
+                    .kernel_info()
+                    .expect("Loaded Bpfd programs should have kernel information");
+
+                (kernel_info.id, p.to_owned())
+            })
+            .collect()
```
dave-tucker, line 154: Again, if this can be an iterator we avoid calling `collect()` and leave it up to the caller what to do with it.

bpfd/src/bpf.rs
```diff
@@ -41,18 +41,145 @@ const MAPS_MODE: u32 = 0o0660;
...
+    }
+}
+
 impl BpfManager {
     pub(crate) fn new(config: Config, commands: mpsc::Receiver<Command>) -> Self {
         Self {
             config,
-            dispatchers: HashMap::new(),
-            programs: HashMap::new(),
+            dispatchers: BpfDispatchers {
```
dave-tucker, line 177: Since you've gone through the bother of making them their own types, add a `new()` function 😉

bpfd/src/bpf.rs
astoycos: Done

bpfd/src/bpf.rs
astoycos: Let me give this a try.

bpfd/src/bpf.rs
astoycos: Gave this a try, just had to add some manual lifetimes

bpfd/src/bpf.rs
astoycos: Done just needed to add some custom lifetime specifiers.

bpfd/src/bpf.rs
astoycos: Done

bpfd/src/bpf.rs
astoycos: Lol done

bpfd/src/multiprog/xdp.rs
```diff
@@ -43,19 +43,20 @@ impl XdpDispatcher {
         mode: XdpMode,
         if_index: &u32,
         if_name: String,
-        programs: &mut [(Uuid, Program)],
+        programs: &mut [(&Uuid, &mut Program)],
```
dave-tucker, line 46: Yeah so if you wanted to take this all the way, this would become
```rust
programs: impl Iterator<Item=(&Uuid, &mut Program)>
```

Any calls to `.len()` get replaced with `.count()`
Any calls to `.first()` become `.peek()`
And then we're good 😉 

We might also - which I know you'll love - be able to remove the call to `sort` on the collected Vec.
It shouldn't matter if we attach the extensions out-of-order 😆 

Anyway, this is definitely new PR territory...

dave-tucker (changes requested):
Looking good! Couple of nits - which are mainly fixing the lack of docs and/or naming that were mostly my fault, but since you're refactoring... 🤣

bpfd/src/bpf.rs
```diff
@@ -48,13 +48,138 @@ const MAPS_MODE: u32 = 0o0660;
...
-    dispatchers: HashMap<DispatcherId, Dispatcher>,
-    programs: HashMap<Uuid, Program>,
+    dispatchers: BpfDispatchers,
+    programs: BpfPrograms,
     maps: HashMap<Uuid, BpfMap>,
     commands: Receiver<Command>,
     image_manager: Sender<ImageManagerCommand>,
 }
 
+pub(crate) struct BpfPrograms {
```
dave-tucker, line 58: Note: We *could* make this type `UuidMap` and it would become:

```rust
pub(crate) struct UuidMap<T> {
   inner: HashMap<Uuid, T>
}
```

You could then use this same struct for `BpfManager::maps`.
So you'd have `UuidMap<Dispatcher>` and `UuidMap<BpfMap>`...

bpfd/src/bpf.rs
```diff
@@ -48,13 +48,138 @@ const MAPS_MODE: u32 = 0o0660;
 
 pub(crate) struct BpfManager {
     config: Config,
-    dispatchers: HashMap<DispatcherId, Dispatcher>,
-    programs: HashMap<Uuid, Program>,
+    dispatchers: BpfDispatchers,
```
dave-tucker, line 51: I would name the new wrapper `DispatcherMap` since it's a map of dispatchers with some fancy helpers.

bpfd/src/bpf.rs
```diff
@@ -48,13 +48,138 @@ const MAPS_MODE: u32 = 0o0660;
...
+    programs: BpfPrograms,
```
dave-tucker, line 52: ProgramsMap or per 👇 UuidMap<Program>

bpfd/src/bpf.rs
```diff
@@ -619,85 +731,21 @@ impl BpfManager {
...
+        let mut bpfd_progs: HashMap<u32, &Program> = self.programs.kernel_infos().collect();
 
         loaded_programs()
             .map(|p| {
                 let prog = p.map_err(BpfdError::BpfProgramError)?;
                 let prog_id = prog.id();
 
                 match bpfd_progs.remove(&prog_id) {
-                    Some(p) => Ok(p),
+                    Some(p) => Ok(p.to_owned()),
```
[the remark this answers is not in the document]
astoycos: It's needed because 

`self.programs.kernel_infos()` is an iterator over program borrows i.e `(u32, &Program)`

bpfd/src/bpf.rs
```diff
@@ -48,13 +48,138 @@ const MAPS_MODE: u32 = 0o0660;
...
+
+    fn get(&self, id: &Uuid) -> Option<&Program> {
+        self.programs.get(id)
+    }
+
+    fn contains_key(&self, id: &Uuid) -> bool {
+        self.programs.contains_key(id)
+    }
+
+    fn next_avail_id(&self, program: &Program) -> usize {
```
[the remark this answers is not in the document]
astoycos: 1. is easy enough... but how do we get the number of attached extensions since we don't actually track the extensions in the dispatcher currently.

I added a new field to the Dispatcher's called `num_extensions` but PTAL

bpfd/src/multiprog/tc.rs
```diff
@@ -210,7 +210,7 @@ impl TcDispatcher {
                     bpf.set_global(name, value.as_slice(), true);
                 }
 
-                let (_, map_pin_path) = calc_map_pin_path(**k, v.data.map_owner_id());
+                let (_, map_pin_path) = calc_map_pin_path(k.to_owned(), v.data.map_owner_id());
```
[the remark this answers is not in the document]
astoycos: Yeah it could but unfortunately with the design of all the map stuff right now that change would have far-reaching effects for this PR. 

Let me tackle this in another one 👍

bpfd/src/multiprog/xdp.rs
astoycos: I like It! leaving these comments open for the next PR

bpfd/src/bpf.rs
astoycos: Done

bpfd/src/bpf.rs
astoycos: Done, ProgramMap for now

bpfd/src/bpf.rs
astoycos: Right but I think I'd rather have a dedicated Type for Maps especially as we enter the next stage of map development.

Specifically I prototyped removing the `BpfMap` type completely but ran into issues because now Maps have slightly different lifetimes then Programs (i.e a Map can hang around if a it's owner program was deleted but other programs are still using it) 

Therefore I think it will make the most sense to move all of the map functionality we have today into traits on a ProgramMap type (no more passing around Owned UUIDs /IDS everwhere YAY!)

dave-tucker (approved):
LGTM. Added the `hold` so you can squash your commits please 🙏

PR #694, opened 2023-09-07
Pin by name

astoycos:
This PR allows bpfd to pin maps AFTER load manually by name using the aya changes which can be previewed here -> https://github.com/aya-rs/aya/compare/main...astoycos:aya:map_pin

This is pre-work to allowing us to use the kernel ID as the one true prog ident i.e #656  since we won't know a programs map pin path until AFTER load. 

This has a hold since it will require our AYA fork to be used if @dave-tucker is alright with that diff ^

It also contains some new integration test updates + coverage which ensure maps are actually being pinned correctly by bpfd.


dave-tucker (approved):

tests/integration-test/src/tests/utils.rs
```diff
@@ -98,12 +102,14 @@ pub fn start_bpfd() -> Result<ChildGuard> {
 }
 
 /// Install an xdp_pass program with bpfctl
```
anfredette, line 104:
```suggestion
/// Install an xdp program with bpfctl
```

tests/integration-test/src/tests/utils.rs
```diff
@@ -142,21 +146,24 @@ pub fn add_xdp_pass(
...
     debug!(
-        "Successfully added xdp_pass program: {:?} from: {:?}",
+        "Successfully added xdp program: {:?} from: {:?}",
         uuid, load_type
     );
 
     Ok(uuid.to_string())
 }
 
 /// Install a tc_pass program with bpfctl
```
anfredette, line 156:
```suggestion
/// Install a tc program with bpfctl
```

tests/integration-test/src/tests/utils.rs
```diff
@@ -392,16 +408,11 @@ pub fn add_kretprobe(globals: Option<Vec<&str>>, load_type: &LoadType) -> Result
...
-            "--pull-policy",
-            "Always",
-        ]),
-        LoadType::File => args.extend(["-s", "my_kretprobe", "--path", KRETPROBE_FILE_LOC]),
+        LoadType::Image => args.extend(["--image-url", image_url, "--pull-policy", "Always"]),
+        LoadType::File => args.extend(["-s", "my_kretprobe", "--path", file_path]),
     }
 
-    args.extend(["kprobe", "-f", "try_to_wake_up", "-r"]);
+    args.extend(["kprobe", "--retprobe", "-f", "try_to_wake_up"]);
```
anfredette, line 415: I don't think this needed to change, but it should work the way you have it too.

tests/integration-test/src/tests/e2e.rs
```diff
@@ -396,3 +432,108 @@ fn test_program_execution_with_global_variables() {
...
+fn test_load_unload_tc_maps() {
+    let _namespace_guard = create_namespace().unwrap();
+    let _ping_guard = start_ping().unwrap();
+    let bpfd_guard = start_bpfd().unwrap();
+
+    assert!(iface_exists(DEFAULT_BPFD_IFACE));
+
+    debug!("Installing tc_counter program");
+
+    // Install an xdp counter program
```
anfredette, line 486:
```suggestion
    // Install a tc counter program
```

anfredette (commented):
Is it possible to read the maps from the integration test code?

Just a couple of nits on the comments. LGTM otherwise.

tests/integration-test/src/tests/utils.rs
```diff
@@ -251,7 +256,12 @@ pub fn add_tracepoint(globals: Option<Vec<&str>>, load_type: &LoadType) -> Resul
 }
 
 /// Attach a uprobe program to bpfctl with bpfctl
-pub fn add_uprobe(globals: Option<Vec<&str>>, load_type: &LoadType) -> Result<String> {
+pub fn add_uprobe(
```
anfredette, line 259: Just thinking.  Now that you've parameterized the image/file stuff, we could combine the add_*probe and add_*retprobe functions and have a boolean retprobe parameter.  That can be another pr, though.

astoycos: ++

PR #701, opened 2023-09-12
One program identifier

### LINKED ISSUE #580: kernel ids - the one true program identifier

astoycos:
closes and pull's from #656 (thanks @Billy99 For fixing up the examples 🥳  I'll fix up the commit so it says owned by you before merging)

Fixes #580


astoycos:
There's still some small spelling and nit stuff in comments but otherwise this should be read for review

astoycos:
Added an extra commit here that fixes filtering by metadata + adds an integration test to ensure it's working.

bpfd/src/bpf.rs
```diff
@@ -66,64 +65,67 @@ impl ProgramMap {
...
-            p.kind() == *program_type && p.if_index() == *if_index && p.direction() == *direction
+    ) -> impl Iterator<Item = &'a mut Program> {
+        self.programs.iter_mut().filter_map(|(_, p)| {
+            if p.kind() == *program_type && p.if_index() == *if_index && p.direction() == *direction
+            {
+                Some(p)
+            } else {
+                None
+            }
         })
```
dave-tucker, line 97: Use `values_mut()`

```suggestion
        self.programs.values_mut().filter(|p| {
```

dave-tucker (changes requested):

dave-tucker (changes requested):
A few nits but overall looking good. I haven't reviewed the Go code though.

dave-tucker (changes requested):
A few nits but overall looking good. I haven't reviewed the Go code though.

bpfd/src/bpf.rs
```diff
@@ -66,64 +65,67 @@ impl ProgramMap {
...
                     if v.if_index() == if_index && v.direction() == direction {
-                        Some((k, v))
+                        Some(v)
                     } else {
                         None
                     }
                 } else {
                     None
                 }
             })
```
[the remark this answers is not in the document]
astoycos: Arg good catch thanks!

dave-tucker (approved):

bpfd/src/bpf.rs
```diff
@@ -875,66 +868,71 @@ impl BpfManager {
...
-            let program = self
-                .programs
-                .get_mut(&id)
-                .expect("Program should be loaded");
-
-            let map = BpfMap { used_by: vec![id] };
-
-            self.maps.insert(id, map);
+        match map_owner_id {
+            Some(id) => {
```
Billy99, line 880:
```suggestion
            Some(owner_id) => {
```
`id` already exists and needs to be used in this case statement. Change `id` to something else in this case statement.

bpfd/src/bpf.rs
```diff
@@ -875,66 +868,71 @@ impl BpfManager {
...
+                if let Some(map) = self.maps.get_mut(&id) {
+                    map.used_by.push(id);
```
Billy99, line 882: Except here. Need to add this program's `id` to the map owner's used_by list.

bpfd/src/bpf.rs
```diff
@@ -875,66 +868,71 @@ impl BpfManager {
...
-                .programs
-                .get_mut(&id)
-                .expect("Program should be loaded");
-
-            let map = BpfMap { used_by: vec![id] };
-
-            self.maps.insert(id, map);
+        match map_owner_id {
+            Some(id) => {
+                if let Some(map) = self.maps.get_mut(&id) {
```
Billy99, line 881:
```suggestion
                if let Some(map) = self.maps.get_mut(&owner_id) {
```

bpfd/src/bpf.rs
```diff
@@ -875,66 +868,71 @@ impl BpfManager {
...
+                    map.used_by.push(id);
+
+                    // If map owner program still exists update it's maps_used_by_field
+                    if let Some(program) = self.programs.get_mut(&id) {
```
Billy99, line 885: Now `program.data.used_by` is what is returned in list/get, so this `if` block needs to be in a loop, and all programs with an instance in this `map.used_by` need to be updated with this new `id`.

The same will need to be done when a program in the `map.used_by` is unloaded. Also, `program.data.used_by` is not being updated on a bpfd restart.

The problem is you are trying to keep the same data in two different places and now you have to keep them in sync. And its a many `program.data.used_by` to one `map.used_by`. It's probably best to not keep `used_by` in ProgramData.

bpfctl/src/main.rs
```diff
@@ -88,12 +92,7 @@ struct LoadFileArgs {
 
     /// Required: Name of the ELF section from the object file.
     #[clap(short, long)]
-    section_name: String,
-
-    /// Optional: Program uuid to be used by bpfd. If not specified, bpfd will generate
-    /// a uuid.
-    #[clap(long, verbatim_doc_comment)]
-    id: Option<String>,
+    name: String,
```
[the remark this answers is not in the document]
astoycos: Nope, literally just the name

PR #702, opened 2023-09-13
fix bpfctl help command

astoycos:
Enable clap default features which automatically creates the `help` target for bpfctl.


astoycos:
@Billy99 Was actually doing this right and writing a test 👑

PR #704, opened 2023-09-13
"bpfctl --help" fails

Billy99:
All the "bpfctl --help" variations are failing:

```
$ bpfctl list --help
error: unexpected argument found
```

Update the bpfctl Cargo.toml file to use the "default" feature flags, which includes: std, color, help, usage, error-content, suggestions

Also add a test to detect this in the future.

### DIFF
--- a/bpfctl/Cargo.toml
+++ b/bpfctl/Cargo.toml
@@ -14,7 +14,14 @@ path = "src/main.rs"
 anyhow = { workspace = true, features = ["std"] }
 base64 = { workspace = true }
 bpfd-api = { workspace = true }
-clap = { workspace = true, features = ["derive", "std"] }
+clap = { workspace = true, features = [
+    "color",
+    "derive",
+    "help",
+    "std",
+    "suggestions",
+    "usage",
+] }
 comfy-table = { workspace = true, features = ["tty"] }
 env_logger = { workspace = true }
 hex = { workspace = true, features = ["std"] }
--- a/tests/integration-test/src/tests/basic.rs
+++ b/tests/integration-test/src/tests/basic.rs
@@ -1,3 +1,6 @@
+use std::process::Command;
+
+use assert_cmd::prelude::*;
 use bpfd_api::util::directories::{
     BYTECODE_IMAGE_CONTENT_STORE, RTDIR_FS_TC_INGRESS, RTDIR_FS_XDP,
 };
@@ -6,6 +9,19 @@ use rand::Rng;
 
 use super::{integration_test, IntegrationTest};
 use crate::tests::utils::*;
+
+#[integration_test]
+fn test_bpfctl_helptext() {
+    let args = vec!["list", "-help"];
+
+    assert!(!Command::cargo_bin("bpfctl")
+        .unwrap()
+        .args(args)
+        .ok()
+        .expect("bpfctl list --help failed")
+        .stdout
+        .is_empty());
+}
 
 #[integration_test]
 fn test_load_unload_xdp() {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -59,10 +59,52 @@ dependencies = [
 ]
 
 [[package]]
-name = "anstyle"
-version = "1.0.2"
+name = "anstream"
+version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "15c4c2c83f81532e5845a733998b6971faca23490340a418e9b72a3ec9de12ea"
+checksum = "b1f58811cfac344940f1a400b6e6231ce35171f614f26439e80f8c1465c5cc0c"
+dependencies = [
+ "anstyle",
+ "anstyle-parse",
+ "anstyle-query",
+ "anstyle-wincon",
+ "colorchoice",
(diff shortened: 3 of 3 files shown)


tests/integration-test/src/tests/basic.rs
```diff
@@ -7,6 +7,11 @@ use rand::Rng;
 use super::{integration_test, IntegrationTest};
 use crate::tests::utils::*;
 
+#[integration_test]
+fn test_bpfctl_helptext() {
+    let _ = get_bpfctl_list_helptext();
```
astoycos, line 12: No need to even use let here if we don't care about the string

astoycos: Just unwrap the result

tests/integration-test/src/tests/utils.rs
```diff
@@ -101,6 +101,20 @@ pub fn start_bpfd() -> Result<ChildGuard> {
...
 
+/// Retrieve bpfctl help test
+pub fn get_bpfctl_list_helptext() -> Result<String> {
+    let mut args = Vec::new();
+    args.push("list");
+    args.push("--help");
+    let output = Command::cargo_bin("bpfctl")?.args(args).ok();
+    let stdout = String::from_utf8(output.unwrap().stdout).unwrap();
+    let helptext = stdout.trim();
+    assert!(!helptext.is_empty());
```
anfredette, line 112: This works, but it's kind of changing the paradigm and doing the check in the util function instead of the test code.

bpfctl/Cargo.toml
```diff
@@ -14,7 +14,7 @@ path = "src/main.rs"
 anyhow = { workspace = true, features = ["std"] }
 base64 = { workspace = true }
 bpfd-api = { workspace = true }
-clap = { workspace = true, features = ["derive", "std"] }
+clap = { workspace = true, features = ["default", "derive"] }
```
dave-tucker, line 17: Can we not use `default`? What's the minimal feature set we need from `clap` for help to work?

dave-tucker (changes requested):

bpfctl/Cargo.toml
anfredette: I think we all thought that the other features in default looked like they'd be nice to have:
std
color
help
usage
error-context
suggestions

dave-tucker: If we want them, we should add them explicitly - it could be that clap decide to change what's in "default" when they next do a release and remove something we rely on. This happened in the `nix` crate recently, but thankfully dependabot caught it.

Billy99: I think we need all 7 features. I can use all 7 if you prefer.

tests/integration-test/src/tests/utils.rs
Billy99: Moved out of the utils.rs file

tests/integration-test/src/tests/basic.rs
Billy99: Reworked based on Andre's comment.

tests/integration-test/src/tests/utils.rs
```diff
@@ -101,6 +101,20 @@ pub fn start_bpfd() -> Result<ChildGuard> {
     Ok(bpfd_process)
 }
 
+/// Retrieve bpfctl help test
+pub fn get_bpfctl_list_helptext() -> Result<String> {
```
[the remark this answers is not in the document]
Billy99: Reworked based on Andre's comment.

anfredette (approved):
LGTM

tests/integration-test/src/tests/utils.rs
```diff
@@ -101,6 +101,23 @@ pub fn start_bpfd() -> Result<ChildGuard> {
...
+    match output {
+        Ok(o) => {
+            let stdout = String::from_utf8(o.unwrap().stdout).unwrap();
+            let helptext = stdout.trim();
+            Ok(helptext.to_string())
+        }
+        Err(e) => Err(e.into()),
+    }
+}
+
```
astoycos, line 120: Main thing is to remove the comments but you could also simplify like this...

```suggestion
/// Retrieve bpfctl help test
pub fn get_bpfctl_list_helptext() -> Result<String, anyhow::Error> {
    let args = vec!["list", "-help"];
    Command::cargo_bin("bpfctl")?
        .args(args)
        .ok()
        .map(|o| {
            String::from_utf8(o.stdout)
                .unwrap_or(String::new())
                .trim()
                .to_string()
        })
        .map_err(|e| e.into())
}
```

In general though (for next time) it probably makes sense to not even wrap this in a util function since it's generally a one off (i.e let's just make sure bpfctl has a help command 😆 ) 

Then it would turn all of this into like this simple thing 

```rust
#[integration_test]
fn test_bpfctl_helptext() {
    let args = vec!["list", "-help"];

    assert!(!Command::cargo_bin("bpfctl").unwrap()
        .args(args)
        .ok()
        .expect("bpfctl list --help failed")
        .stdout.is_empty());
}
```

dave-tucker (approved):

PR #705, opened 2023-09-13
k8s support for uprobes

### LINKED ISSUE #369: Support Uprobes
Note: This should probably happen after #342

As a user of bpfd, I would like the ability to load uprobes to trace my program.

Uprobes can be attached 3 different ways:

1. To a shared library - identified by name
2. To all instances of process - identified by path "/bin/bash"
3. To a specific instance of process - identified by path and pid e,g "/bin/bash" pid 9

### bpfd

1. Adjust the gRPC API to allow the attachment of uprobes, for the required fields in `AttachInfo` see: https://docs.aya-rs.dev/user/aya/programs/uprobe/struct.uprobe
1. Add logic to load uprobes

### bpfd-operator

1. Adjust the CRD to allow for the attachment of uprobe programs
2. This will include a `podSelector` to target specific pods within the cluster, similar to what is used in `NetworkPolicy`.

### Challenges

Containers make uprobe attachment a tricky business.
We will need to enforce [Process Namespace Sharing](https://kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/) on pods that want uprobes.
Then we will spawn a sidecar container to load programs with access to bpffs so the program can be pinned.
Unload is simply a case of removing the pin and doesn't require a sidecar.
The use of a sidecar container is needed to run in the same process namespace as the containers in a pod.

What we'll need to be sure of is whether attachment through `/proc/$pid/root` will work, where `$pid` is the entrypoint process of a container in the pod.

### Aya considerations

The existing uprobe API should work fine for our usecases, although we may want to add an alternative API that doesn't do symbol resolution, but instead allows the caller to provide the symbol offset directly.

### LINKED ISSUE #564: Add support for uretprobe, kprobe and kretprobe

anfredette:
Also includes a fix in DoesProgExist() for kprobes

resolves #369
resolves #564


bpfd-operator/controllers/bpfd-agent/uprobe-program.go
```diff
@@ -0,0 +1,340 @@
...
+			Namespace: &namespace,
+		},
+	}
+
+	return loadRequest
+}
+
+// reconcileBpfdPrograms ONLY reconciles the bpfd state for a single BpfProgram.
+// It does not interact with the k8s API in any way.
+func (r *UprobeProgramReconciler) reconcileBpfdProgram(ctx context.Context,
```
astoycos, line 209: You don't have to deal with this now... but it __should__ be farily simple to move all of this code into common.go the more I look at it.. That's a bit smaller diff everytime we add a new program :celebrate

bpfd-operator/controllers/bpfd-agent/uprobe-program_test.go
```diff
@@ -0,0 +1,211 @@
...
+		Spec: bpfdiov1alpha1.UprobeProgramSpec{
+			BpfProgramCommon: bpfdiov1alpha1.BpfProgramCommon{
+				SectionName:  sectionName,
+				NodeSelector: metav1.LabelSelector{},
+				ByteCode: bpfdiov1alpha1.BytecodeSelector{
+					Path: &bytecodePath,
+				},
+			},
+			FunctionName: functionName,
+			Targets:      []string{target},
```
astoycos, line 77: Do we actually support multiple targets here? In the controller it looks like we only support loading one

```
bpfProgram.Annotations[internal.UprobeProgramTarget]
```

bpfd-operator/controllers/bpfd-agent/uprobe-program_test.go
```diff
@@ -0,0 +1,211 @@
...
+		},
+	}
+
+	// First reconcile should create the bpf program object
+	res, err := r.Reconcile(ctx, req)
+	if err != nil {
+		t.Fatalf("reconcile: (%v)", err)
+	}
+
+	// Check the BpfProgram Object was created successfully
```
astoycos, line 127: Additionally for all this testing infra... I wonder if we can condense it somehow.... (followup)

bpfd-operator/controllers/bpfd-operator/uprobe-program_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+	// Mock request to simulate Reconcile() being called on an event for a
+	// watched resource .
+	req := reconcile.Request{
+		NamespacedName: types.NamespacedName{
+			Name: name,
+		},
+	}
+
+	// First reconcile should add the finalzier to the tcProgram object
```
astoycos, line 126:
```suggestion
	// First reconcile should add the finalizer to the uprobeProgram object
```

astoycos: This is probably wrong everywhere lol

astoycos (changes requested):
Just some NITS

I think we need to de-dup alot of this code sometime soon. 

Additionally... next time you might be able to use https://v0-19-x.sdk.operatorframework.io/docs/cli/operator-sdk_add_api/ to do some of the setup for you!

bpfd-operator/controllers/bpfd-operator/uprobe-program_test.go
anfredette: At least on tc forward!

anfredette: Nope.  It's wrong on all but tc 🤣.

bpfd-operator/controllers/bpfd-agent/uprobe-program.go
anfredette: Agreed.  I opened https://github.com/bpfd-dev/bpfd/issues/709.

bpfd-operator/controllers/bpfd-agent/uprobe-program_test.go
anfredette: I created https://github.com/bpfd-dev/bpfd/issues/710

bpfd-operator/controllers/bpfd-agent/internal/cmp.go
```diff
@@ -117,15 +117,40 @@ func DoesProgExist(actual *gobpfd.ListResponse_ListResult, expected *gobpfd.Load
...
+	}
+
 	if len(reasons) == 0 {
 		return true, reasons
 	} else {
 		return false, reasons
 	}
 }
+
+func strPntrsEqual(str1 *string, str2 *string) bool {
```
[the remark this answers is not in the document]
anfredette: Good point.  I'll change it.

bpfd-operator/controllers/bpfd-agent/uprobe-program_test.go
anfredette: You know, I haven't actually tested multiple targets.  I'll do that, but I though that was being handled here:
https://github.com/anfredette/bpfd/blob/bd3462b1af22031ce7c2f00bf0a310b98ded6a98/bpfd-operator/controllers/bpfd-agent/uprobe-program.go#L90

anfredette:
Thanks for the review @astoycos.  I think I've addressed everything locally, but I want to do some testing -- particularly on multiple targets before I push it up.

bpfd-operator/controllers/bpfd-agent/uprobe-program_test.go
astoycos: Ahhh right ok then it should be alright (We need to test it) but that should result in multiple bpfProgram objects being made (one for each target).

astoycos (approved):
LGTM

Thanks @anfredette Merging this for now but can you make sure to open an issue for updating the docs so @Billy99 Doesn't have to go back and do it all 🤣

PR #717, opened 2023-09-18
Address a few minor comments from astoycos on pr #705

anfredette:
Address a few minor comments from @astoycos on pr #705


astoycos (approved):
LGTM

thanks @anfredette!

PR #719, opened 2023-09-18
BPFD CSI Deployment Updates

astoycos:
Setup the deployment for the bpfd CSI plugin.

You can test this out with 

```
cd bpfd-operator
make run-on-kind-csi
cd ../examples
make deploy-xdp
```

currently the driver only logs some metadata from the CSI request and then sleeps for 60 seconds before returning.

Im the time that bpfd is working to make the driver the pod stay's in the container-creating state 

```
go-xdp-counter       go-xdp-counter-ds-rpkvb                                 0/1     ContainerCreating   0          33s
```

The Csi deployment can be configured by setting 

`  bpfd.enable.csi: "true"` 

in the bpfd configmap.


astoycos:
Breaking this into two PRs

dave-tucker (approved):

Issue #614, opened 2023-07-24
Checl `go mod` in CI


dave-tucker:
We shouldn't be checking in an un-tidied go.mod.
There should be a CI check to detect this and fail if we need to run `go mod tidy`

astoycos:
Also look into adding gofmt checks to our golang-ci run.

astoycos:
Shoutout to @navarrothiago for fixing this in #720

PR #720, opened 2023-09-18
ci: adds go mod check

navarrothiago:
Closes: #614


.github/workflows/build.yml
```diff
@@ -126,6 +126,11 @@ jobs:
         with:
           go-version: "1.20"
 
+      - name: Go mod check
+        run: |
```
astoycos, line 130:
```suggestion
        run: |
        working-directory: ${{ env.GOPATH }}/src/github.com/bpfd-dev/bpfd
```

astoycos:
Thanks for tackling this!!

PR #721, opened 2023-09-19
bpfd: Add bpffs helpers to StorageManager

dave-tucker:
Pulls in https://github.com/bpfd-dev/aya/commit/27df828e2f95799ecf1f977a003477bbb1b2c011

### DIFF
--- a/bpfd/src/storage.rs
+++ b/bpfd/src/storage.rs
@@ -1,7 +1,10 @@
 // SPDX-License-Identifier: (MIT OR Apache-2.0)
 // Copyright Authors of bpfd
 
+use std::path::Path;
+
 use async_trait::async_trait;
+use aya::maps::MapData;
 use bpfd_csi::v1::{
     controller_server::{Controller, ControllerServer},
     node_server::{Node, NodeServer},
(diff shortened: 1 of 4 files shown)


bpfd/src/storage.rs
```diff
@@ -203,4 +206,25 @@ impl StorageManager {
...
+    fn create_bpffs(&self, path: &Path) -> anyhow::Result<()> {
+        create_bpffs(
+            path.as_os_str()
+                .to_str()
+                .expect("unable to convert path to str"),
+        )
+    }
+
+    #[allow(dead_code)] // TODO: Remove this when the storage manager is fully implemented
+    fn pin_map_to_bpffs(
```
astoycos, line 220: So for duping a map to a new bpffs we'll just do an aya `from_pin` call to get the mapdata and then pin again to the new bpffs right?

dave-tucker: Yep. The whole process is:

- MapData::from_pin() - which behind the scenes does BPF_OBJ_GET from $basePath
- Then MapData::pin() - which does BPF_OBJ_PIN to $targetPath

On my TODO list is to check whether we can `cp` the pinned object from one bpffs to another.
That's riskier, so we'll stick with the above for now.

astoycos: Sounds good to me Thanks!

astoycos (approved):
LGTM

Thanks @dave-tucker

PR #722, opened 2023-09-19
bpfd/csi: ephemeral inline volume scaffolding

astoycos:
Implement the scaffolding required CSI endpoints for ephemeral inline volumes.

Currently the plugin just prints info on `NodePublishVolume` and `NodeUnPublishVolume`

i.e 

```
[INFO  bpfd::storage] Received publish volume request with :
    volume_caps: Some(
        VolumeCapability {
            access_mode: Some(
                AccessMode {
                    mode: SingleNodeWriter,
                },
            ),
            access_type: Some(
                Mount(
                    MountVolume {
                        fs_type: "",
                        mount_flags: [],
                        volume_mount_group: "",
                    },
                ),
            ),
        },
    ),
    volume_id: csi-413a3e6b41db24365286e94018a6644fa3190b8ab99b63269519562af55b8930,
    target_path: /var/lib/kubelet/pods/fcd164d3-cbe8-4044-8efd-c7f3e46b046d/volumes/kubernetes.io~csi/bpf-maps/mount,
    volume_context: {},
    read_only: false
```

### DIFF
--- a/bpfd-api/src/util.rs
+++ b/bpfd-api/src/util.rs
@@ -32,9 +32,13 @@ pub mod directories {
     pub const RTDIR_FS_XDP: &str = "/run/bpfd/fs/xdp";
     pub const RTDIR_FS_MAPS: &str = "/run/bpfd/fs/maps";
     pub const RTDIR_PROGRAMS: &str = "/run/bpfd/programs";
+    pub const STPATH_BPFD_SOCKET: &str = "/run/bpfd/bpfd.sock";
+    // The CSI socket must be in it's own sub directory so we can easily create a dedicated
+    // K8s volume mount for it.
+    pub const RTDIR_BPFD_CSI: &str = "/run/bpfd/csi";
+    pub const STPATH_BPFD_CSI_SOCKET: &str = "/run/bpfd/csi/csi.sock";
+
     // StateDirectory: /var/lib/bpfd/
     pub const STDIR: &str = "/var/lib/bpfd";
-    pub const STDIR_SOCKET: &str = "/var/lib/bpfd/sock";
-    pub const STPATH_BPFD_SOCKET: &str = "/var/lib/bpfd/sock/bpfd.sock";
     pub const BYTECODE_IMAGE_CONTENT_STORE: &str = "/var/lib/bpfd/io.bpfd.image.content";
 }
--- a/bpfd/src/serve.rs
+++ b/bpfd/src/serve.rs
@@ -23,11 +23,14 @@ use tonic::transport::{Server, ServerTlsConfig};
 
 pub use crate::certs::get_tls_config;
 use crate::{
-    bpf::BpfManager, errors::BpfdError, oci_utils::ImageManager, rpc::BpfdLoader,
-    static_program::get_static_programs, storage::StorageManager, utils::set_file_permissions,
+    bpf::BpfManager,
+    errors::BpfdError,
+    oci_utils::ImageManager,
+    rpc::BpfdLoader,
+    static_program::get_static_programs,
+    storage::StorageManager,
+    utils::{set_file_permissions, SOCK_MODE},
 };
-
-const SOCK_MODE: u32 = 0o0770;
 
 pub async fn serve(
     config: Config,
@@ -80,8 +83,6 @@ pub async fn serve(
         }
     }
 
-    let storage_manager = StorageManager::new();
-
     let (itx, irx) = mpsc::channel(32);
     let mut image_manager = ImageManager::new(BYTECODE_IMAGE_CONTENT_STORE, irx);
 
@@ -98,6 +99,8 @@ pub async fn serve(
         }
     };
     if csi_support {
+        let storage_manager = StorageManager::new();
+
         join!(
             join_listeners(listeners),
             bpf_manager.process_commands(),
--- a/bpfd/src/storage.rs
+++ b/bpfd/src/storage.rs
@@ -1,131 +1,81 @@
 // SPDX-License-Identifier: (MIT OR Apache-2.0)
 // Copyright Authors of bpfd
 
-use std::path::Path;
+use std::{collections::HashMap, fs::remove_file, path::Path};
 
 use async_trait::async_trait;
 use aya::maps::MapData;
+use bpfd_api::util::directories::STPATH_BPFD_CSI_SOCKET;
 use bpfd_csi::v1::{
-    controller_server::{Controller, ControllerServer},
+    identity_server::{Identity, IdentityServer},
     node_server::{Node, NodeServer},
-    ControllerExpandVolumeRequest, ControllerExpandVolumeResponse,
-    ControllerGetCapabilitiesRequest, ControllerGetCapabilitiesResponse,
-    ControllerGetVolumeRequest, ControllerGetVolumeResponse, ControllerModifyVolumeRequest,
-    ControllerModifyVolumeResponse, ControllerPublishVolumeRequest,
-    ControllerPublishVolumeResponse, ControllerUnpublishVolumeRequest,
-    ControllerUnpublishVolumeResponse, CreateSnapshotRequest, CreateSnapshotResponse,
-    CreateVolumeRequest, CreateVolumeResponse, DeleteSnapshotRequest, DeleteSnapshotResponse,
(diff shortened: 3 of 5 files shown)


bpfd/src/serve.rs
```diff
@@ -27,7 +27,8 @@ use crate::{
     static_program::get_static_programs, storage::StorageManager, utils::set_file_permissions,
 };
 
-const SOCK_MODE: u32 = 0o0770;
+pub(crate) const SOCK_MODE: u32 = 0o0770;
```
dave-tucker, line 30: nit: Should we pull this into utils?

bpfd/src/serve.rs
```diff
@@ -27,7 +27,8 @@ use crate::{
...
+const DRIVER_NAME: &str = "bpfd.csi.k8s.io";
```
dave-tucker, line 32:
```suggestion
const DRIVER_NAME: &str = "csi.bpfd.dev";
```

bpfd/src/serve.rs
```diff
@@ -98,6 +97,12 @@ pub async fn serve(
         }
     };
     if csi_support {
+        const VERSION: &str = env!("CARGO_PKG_VERSION");
+        let node_id = std::env::var("KUBE_NODE_NAME")
+            .expect("cannot start bpfd csi driver KUBE_NODE_NAME not set");
+        let storage_manager =
```
dave-tucker, line 103:
```suggestion
            .expect("cannot start bpfd csi if driver KUBE_NODE_NAME not set");
```

Just curious as to why that is?

bpfd/src/serve.rs
```diff
@@ -98,6 +97,12 @@ pub async fn serve(
...
+            StorageManager::new(DRIVER_NAME.to_string(), node_id, VERSION.to_string());
```
dave-tucker, line 104: This should all just be computed in `StorageManager::new()`.
No need to pass it into the constructor.
```suggestion
            StorageManager::new();
```

bpfd/src/storage.rs
```diff
@@ -1,206 +1,370 @@
...
-        todo!()
+        return Err(Status::unimplemented(""));
     }
+
     async fn controller_get_capabilities(
         &self,
         _request: Request<ControllerGetCapabilitiesRequest>,
     ) -> std::result::Result<Response<ControllerGetCapabilitiesResponse>, Status> {
-        todo!()
+        return Ok(Response::new(ControllerGetCapabilitiesResponse {
```
dave-tucker, line 169: Do we need to implement this at all?
Would be convenient if we didn't need to as we could drop the Controller impl entirely.

bpfd/src/storage.rs
```diff
@@ -1,209 +1,371 @@
...
+        };
+        let csi_node = CsiNode { driver };
         Self {
             csi_controller,
             csi_node,
+            csi_identity,
         }
     }
 
     pub async fn run(self) {
```
dave-tucker, line 341: Is this mandatory in the spec? What if something is already running on that socket?

bpfd/src/storage.rs
```diff
@@ -1,209 +1,371 @@
...
+            .serve_with_incoming_shutdown(uds_stream, shutdown_handler());
 
         tokio::spawn(async move {
-            info!("CSI Plugin Listening on {addr}");
+            info!("CSI Plugin Listening on {}", path.display());
             if let Err(e) = serve.await {
                 eprintln!("Error = {e:?}");
             }
-            info!("Shutdown CSI Plugin {}", addr);
+            info!("Shutdown CSI Plugin Handler {}", path.display());
```
dave-tucker, line 368:
```suggestion
            info!("Shutdown CSI Plugin {}", path.display());
```

bpfd/src/storage.rs
```diff
@@ -1,209 +1,371 @@
...
 pub(crate) struct StorageManager {
+    csi_identity: CsiIdentity,
     csi_controller: CsiController,
     csi_node: CsiNode,
 }
 
-struct CsiController {}
+struct CsiIdentity {
+    driver: StorageDriver,
+}
```
dave-tucker, line 66:
```suggestion
struct CsiIdentity {
    name: String,
    version: String,
}
```

bpfd/src/storage.rs
```diff
@@ -1,209 +1,371 @@
...
+
+struct CsiNode {
+    driver: StorageDriver,
+}
```
dave-tucker, line 70:
```suggestion
struct CsiNode {
    node_id: String,
}
```

bpfd/src/storage.rs
```diff
@@ -1,209 +1,371 @@
...
+
+struct CsiController {
+    driver: StorageDriver,
+}
```
dave-tucker, line 74:
```suggestion
struct CsiController {
    cap: Vec<ControllerServiceCapability>,
}
```

bpfd/src/storage.rs
```diff
@@ -1,209 +1,371 @@
...
+impl Identity for CsiIdentity {
+    async fn get_plugin_info(
+        &self,
+        _request: Request<GetPluginInfoRequest>,
+    ) -> Result<Response<GetPluginInfoResponse>, Status> {
+        return Ok(Response::new(GetPluginInfoResponse {
+            name: self.driver.name.to_string(),
+            vendor_version: self.driver.version.to_string(),
+            manifest: HashMap::new(),
+        }));
```
dave-tucker, line 86:
```suggestion
        return Ok(Response::new(GetPluginInfoResponse {
            name: self.name.clone(),
            vendor_version: self.version.clone(),
            manifest: HashMap::new(),
        }));
```

dave-tucker (changes requested):

dave-tucker (commented):
Looking good. Some cleanup required but a step in the right direction.

bpfd/src/storage.rs
astoycos: Done

bpfd/src/storage.rs
astoycos: Actually this is just mounted in the bpfd pod at /run/bpfd/sock/csi.sock.

on the host it will be mounted in the predefined `/var/lib/kubelet/plugins/csi-bpfd` so we will always own the socket there i.e we don't care if something else is running there.

astoycos: Additionally I moved the default bpfd socket to the runtime dir as well rather then the state dir `/var/lib`

astoycos (commented):
This got much simpler now @dave-tucker Thanks for the review! 

Eliminated all of the controller plugin endpoints 🎉

astoycos:
Eww I don't like replying in reviews sorry about that

bpfd-api/src/util.rs
```diff
@@ -32,9 +32,11 @@ pub mod directories {
     pub const RTDIR_FS_XDP: &str = "/run/bpfd/fs/xdp";
     pub const RTDIR_FS_MAPS: &str = "/run/bpfd/fs/maps";
     pub const RTDIR_PROGRAMS: &str = "/run/bpfd/programs";
+    pub const STDIR_SOCKET: &str = "/run/bpfd/sock";
+    pub const STPATH_BPFD_SOCKET: &str = "/run/bpfd/sock/bpfd.sock";
```
dave-tucker, line 36: nit: You can just leave them in `/run/bpfd`
No need for a subdirectory.

dave-tucker (approved):
LGTM. Just one nit that will bug me, so if you can take care of that one and remove the `hold` label this is then good to merge.

bpfd-api/src/util.rs
astoycos: Lol for sure!

astoycos: So I ended still having to create a `/run/bpfd/csi/` dir because I still needed a dedicated directory for the csi sock so that we could volume mount it separately in K8s.  
i.e today we do 

```
        - name: runtime
          mountPath: /run/bpfd
          mountPropagation: Bidirectional
``` 

and needed 

```
        - mountPath: /run/bpfd
           subPath: csi.sock
          name: socket-dir
```

as well which lead to daemonset issues.

However once we can configure this socket location (if we end up doing that) this issue will go away and by default it can all just go in /run/bpfd 👍  I hope that's alright

PR #723, opened 2023-09-19
bpfd: bump prost,prost-types and tonic

astoycos:
Bump these deps all together since they have build dependencies on eachother.


astoycos:
Also fixes https://github.com/bpfd-dev/bpfd/issues/396

PR #726, opened 2023-09-19
Add support for Get API

Billy99:
bpfd does not support a Get API. To retrieve a given program, the application must retrieve all programs and filter out the program it needs. This PR adds a Get API to bpfd, which allows an application to retrieve a given program by it's program id. As part of this PR, the Load commands also return the same data returned by the Get for the program that was just loaded.

The API structures were also reworked so that all the Kernel Data is in its own structure and the bpfd based data is in its own structure. These new structs are returned by the List, Get and Load commands.


proto/bpfd.proto
```diff
@@ -7,27 +7,71 @@ service Bpfd {
...
+ * internally by bpfd. */
+
+message ProgramInfo {
+    /* Name entered by user */
+    string name = 1;
+    BytecodeLocation bytecode = 2;
+    AttachInfo attach = 3;
+    map<string, bytes> global_data = 4;
+    optional uint32 map_owner_id = 5;
+    optional string map_pin_path = 6;
```
astoycos, line 67: At least in our internal api map_pin_path isn't optional, and i'm not sure it should be here either.

proto/bpfd.proto
```diff
@@ -75,59 +139,88 @@ message LoadRequest {
...
-    uint32 btf_id = 20;
-    uint32 bytes_xlated = 21;
-    bool jited = 22;
-    uint32 bytes_jited = 23;
-    uint32 bytes_memlock = 24;
-    uint32 verified_insns = 25;
-  } 
-  repeated ListResult results = 26;
+  message ListResult { 
+    ProgramInfo info = 1;
```
astoycos, line 195: `ProgramInfo` should be optional here I think since not every program will have bpfd specific informaton

proto/bpfd.proto
```diff
@@ -75,59 +139,88 @@ message LoadRequest {
...
     map<string, string> match_metadata = 3;
 }
 
+/* ListResponse represents a response a request to list loaded and attached
+ * eBPF programs. Based on the corresponding ListRequest programs will include
+ * those loaded by bpfd AND/OR those loaded by other users. The ListResults
+ * will ALWAYS contain a program's kernel info and if loaded by bpfd, will
+ * also contain the state kept by bpfd.
+ */
+
```
astoycos, line 192:
```suggestion
/* ListResponse represents a response from listing loaded and attached
 * eBPF programs. Based on the corresponding ListRequest programs will include
 * those loaded by bpfd AND/OR those loaded by other users. The ListResults
 * will ALWAYS contain a program's kernel info and if loaded by bpfd, will
 * also contain the state kept by bpfd.
 */

```

proto/bpfd.proto
```diff
@@ -75,59 +139,88 @@ message LoadRequest {
...
+message GetRequest {
+    uint32 id = 1;
+}
+
+/* GetResponse represents a response from getting an eBPF program. 
+ * It will ALWAYS contain a program's kernel info and if loaded by bpfd, will
+ * contain the state kept by bpfd. */
+
+message GetResponse { 
+    ProgramInfo info = 1;
```
astoycos, line 224: Optional here as above

astoycos (changes requested):
Just some comments + nits on the API to start with

proto/bpfd.proto
Billy99: It's not entered by the user, but returned from bpfd. So needs to be optional.

proto/bpfd.proto
```diff
@@ -7,27 +7,71 @@ service Bpfd {
...
+}
+
+/* KernelProgramInfo represents the state for a single ITS_BPF_NOT_EBPF program that is
+ * maintained internally by the Linux kernel within its bpf_prog_info
+ * structure.
+ */
+
+message KernelProgramInfo { 
+    uint32 id = 1;
+    /* Same a Name entered by user, but truncated to 16 byte by kernel */
```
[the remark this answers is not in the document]
Billy99: OK, tweaked some.
```
    /* Name as seen by the kernel. May be truncated version of user enter */
    /* name since limited to 16 byte by kernel */
```

proto/bpfd.proto
Billy99: OK

proto/bpfd.proto
Billy99: I'm fine changing it here in the external API to optional, but it is optional in the internal API (command.rs: ProgramData).

proto/bpfd.proto
Billy99: Done

proto/bpfd.proto
Billy99: Done

proto/bpfd.proto
Billy99: Done

bpfd/src/bpf.rs
```diff
@@ -772,6 +767,26 @@ impl BpfManager {
...
+    pub(crate) fn get_program(&mut self, id: u32) -> Result<Program, BpfdError> {
+        debug!("BpfManager::get_program({0})", id);
+        // If the program was loaded by bpfd, then us it.
+        // Otherwise, call Aya to get ALL the loaded eBPF programs, and convert the data
+        // returned from Aya into an Unsupported Program Object.
+        match self.programs.get(&id) {
+            Some(p) => Ok(p.to_owned()),
+            None => loaded_programs()
+                .find_map(|p| {
+                    let prog = p.map_err(BpfdError::BpfProgramError).ok()?;
```
astoycos, line 779: No need to map the error if we're just discarding it :) 
```suggestion
                    let prog = p.ok()?;
```

astoycos (changes requested):
BPFD Core Review

This just reviews the core bits, I saw you pushed an update while I was reviewing so please ignore any comments that are out of date!

bpfctl/src/main.rs
```diff
@@ -663,69 +697,49 @@ impl LoadCommands {
...
-                name: name.to_string(),
-                program_type: command.get_prog_type() as u32,
-                global_data: parse_global(global),
-                map_owner_id: *map_owner_id,
+                name: _,
+                metadata: _,
+                global: _,
+                map_owner_id: _,
+                command: _,
+            }) => Ok(Some(BytecodeLocation {
```
astoycos, line 725:
```suggestion
            Commands::LoadFromFile(LoadFileArgs {
                path,
                ..
            }) => Ok(Some(BytecodeLocation {
```

bpfctl/src/main.rs
```diff
@@ -663,69 +697,49 @@ impl LoadCommands {
...
-                    .unwrap_or(vec![])
-                    .iter()
-                    .map(|(k, v)| (k.to_owned(), v.to_owned()))
-                    .collect(),
-                location: Some(load_request_common::Location::Image(pull_args.try_into()?)),
-                name: name.to_string(),
-                program_type: command.get_prog_type() as u32,
-                global_data: parse_global(global),
-                map_owner_id: *map_owner_id,
+                name: _,
```
astoycos, line 730: same as above remove all the `_`

bpfctl/src/main.rs
```diff
@@ -874,23 +888,70 @@ async fn execute_request_tcp(
...
+                global_data: parse_global(&l.global),
+                uuid: None,
+                map_owner_id: l.map_owner_id,
             });
             let response = client.load(request).await?.into_inner();
-            println!("{}", response.id);
+
+            ProgTable::new_get_bpfd(&response.info)?.print();
+            ProgTable::new_get_unsupported(&response.kernel_info)?.print();
         }
```
astoycos, line 955: Both patterns here are doing the exact same thing can you just do something like this?

```suggestion
        _(l) => {
            let bytecode = match command.get_bytecode_location() {
                Ok(t) => t,
                Err(e) => bail!(e),
            };

            let attach = match command.get_attach_info() {
                Ok(t) => t,
                Err(e) => bail!(e),
            };

            let request = tonic::Request::new(LoadRequest {
                bytecode,
                name: l.name.to_string(),
                program_type: l.command.get_prog_type() as u32,
                attach,
                metadata: l
                    .metadata
                    .clone()
                    .unwrap_or(vec![])
                    .iter()
                    .map(|(k, v)| (k.to_owned(), v.to_owned()))
                    .collect(),
                global_data: parse_global(&l.global),
                uuid: None,
                map_owner_id: l.map_owner_id,
            });
            let response = client.load(request).await?.into_inner();

            ProgTable::new_get_bpfd(&response.info)?.print();
            ProgTable::new_get_unsupported(&response.kernel_info)?.print();
        }
```

bpfd-operator/controllers/bpfd-agent/internal/bpfd-core.go
```diff
@@ -246,21 +223,25 @@ func GetMapsForID(id uint32) (map[string]string, error) {
...
+			"Map-IDs":                       fmt.Sprintf("%v", kernelInfo.GetMapIds()),
+			"BTF-ID":                        fmt.Sprint(kernelInfo.GetBtfId()),
+			"Size-Translated-Bytes":         fmt.Sprint(kernelInfo.GetBytesXlated()),
+			"JITed":                         fmt.Sprintf("%v", kernelInfo.GetJited()),
+			"Size-JITed-Bytes":              fmt.Sprint(kernelInfo.GetBytesJited()),
+			"Kernel-Allocated-Memory-Bytes": fmt.Sprint(kernelInfo.GetBytesMemlock()),
+			"Verified-Instruction-Count":    fmt.Sprint(kernelInfo.GetVerifiedInsns()),
+		}
 	}
+	return map[string]string{}
```
astoycos, line 244:
```suggestion
	return nil
```

bpfd-operator/controllers/bpfd-agent/common.go
```diff
@@ -458,7 +458,11 @@ func ProcessMapOwnerParam(ctx context.Context,
 				return nil, fmt.Errorf("failed to get bpfd program for BpfProgram with UUID %s: %v", bpfProgramList.Items[0].GetUID(), err)
 			}
 
-			mapOwnerStatus.mapOwnerId = &prog.Id
+			kernelInfo := prog.GetKernelInfo()
+			if kernelInfo == nil {
+				return nil, fmt.Errorf("failed to process bpfd program for BpfProgram with UUID %s: %v", bpfProgramList.Items[0].GetUID(), err)
```
astoycos, line 463: NIT this and above should be UID not UUID (that's just how it's stated in the k8s spec)

bpfd-operator/controllers/bpfd-agent/discovered_program.go
```diff
@@ -100,21 +100,30 @@ func (r *DiscoveredProgramReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 	for _, p := range programs {
-		// skip bpf programs loaded by bpfd, their corresponding bpfProgram object
-		// will be managed by another controller.
-		if _, ok := p.Metadata[internal.UuidMetadataKey]; ok {
+		kernelInfo := p.GetKernelInfo()
+		if kernelInfo == nil {
 			continue
 		}
 
+		bpfProgName := ""
```
astoycos, line 108: nit: move this var back to where it was to avoid the diff

bpfd-operator/controllers/bpfd-agent/kprobe-program.go
```diff
@@ -174,30 +174,32 @@ func (r *KprobeProgramReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-	loadRequest.AttachInfo = &gobpfd.LoadRequest_KprobeAttachInfo{
-		KprobeAttachInfo: &gobpfd.KprobeAttachInfo{
-			FnName:    bpfProgram.Annotations[internal.KprobeProgramFunction],
-			Offset:    r.currentKprobeProgram.Spec.Offset,
-			Retprobe:  r.currentKprobeProgram.Spec.RetProbe,
-			Namespace: &namespace,
+
+	return &gobpfd.LoadRequest{
+		Bytecode:    bytecode,
+		Name:        r.currentKprobeProgram.Spec.SectionName,
```
astoycos, line 187: oof we need to change this to `ProgramName` or just `Name` do we have an issue for that?

bpfd-operator/controllers/bpfd-agent/internal/bpfd-core.go
```diff
@@ -246,21 +223,25 @@ func GetMapsForID(id uint32) (map[string]string, error) {
 
 // Convert a list result into a set of kernel info annotations
 func Build_kernel_info_annotations(p *gobpfd.ListResponse_ListResult) map[string]string {
```
astoycos, line 225: I didn't see this anywhere, but can you also now make sure to update the BpfPrograms for OUR programs (programs loaded by bpfd) with these annotations as well? 

(Doing this in a separate PR is alright)

tests/integration-test/src/tests/e2e.rs
```diff
@@ -341,67 +328,61 @@ fn test_program_execution_with_global_variables() {
         &LoadType::Image,
         TC_PASS_IMAGE_LOC,
         TC_PASS_FILE_LOC,
-    )
-    .unwrap();
+    );
 
-    uuids.push(uuid);
+    uuids.push(prog_id.unwrap());
```
astoycos, line 333: nit: it's probably just ids now but that's alrigt, ultimately all this needs to be deduplicated anyways

astoycos (changes requested):
Almost there 🥳 !!!!

bpfd/src/oci_utils/image_manager.rs
```diff
@@ -82,18 +82,25 @@ impl BytecodeImage {
 
 impl From<bpfd_api::v1::BytecodeImage> for BytecodeImage {
     fn from(value: bpfd_api::v1::BytecodeImage) -> Self {
-        BytecodeImage::new(
-            value.url,
-            value.image_pull_policy,
-            match value.username.as_ref() {
+        let username = if value.username.is_some() {
```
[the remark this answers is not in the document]
Billy99: Adding function back. This is what is causing the pull failures. username and password are already Option<String>, but looks like "" is not valid. So this function was converting "" into None.

bpfd/src/oci_utils/image_manager.rs
```diff
@@ -82,18 +82,25 @@ impl BytecodeImage {
 
 impl From<bpfd_api::v1::BytecodeImage> for BytecodeImage {
```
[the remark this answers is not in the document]
Billy99: Adding function back.

bpfd-operator/controllers/bpfd-agent/internal/bpfd-core.go
Billy99: #744

tests/integration-test/src/tests/e2e.rs
Billy99: Done, was already on my list to do.

bpfd-operator/controllers/bpfd-agent/internal/bpfd-core.go
Billy99: Done

bpfctl/src/main.rs
Billy99: I agree on the duplicate code, but stuck on how to do it in rust.

Billy99:
/hold
I think I found a bug. Let me investigate

Billy99:
Fixed the bug. `bpfctl list --all` was causing bpfd to panic when there was no ProgramInfo (programs not created by bpfd).
```
impl TryFrom<&Program> for V1ProgramInfo {
    type Error = BpfdError;

    fn try_from(program: &Program) -> Result<Self, Self::Error> {
-        let data = program.data().expect("program info not available");
+        let data = program.data()?;
```
And updated the code calling it to map to None when an error was seen (For example: rpc.rs line 129).

astoycos (approved):
LGTM

Thanks for all the work here @Billy99

PR #727, opened 2023-09-21
Implement bpffs support in the bpfd CSI plugin

astoycos:
This is the POC skeleton which allows k8s pods to request certain loaded maps from bpfd. For example with it you can run 

```bash
cat <<EOF | kubectl apply -f -
apiVersion: bpfd.dev/v1alpha1
kind: XdpProgram
metadata:
  labels:
    app.kubernetes.io/name: xdpprogram
  name: go-xdp-counter-example
spec:
  sectionname: xdp_stats
  # Select all nodes
  nodeselector: {}
  interfaceselector:
    primarynodeinterface: true
  priority: 55
  bytecode:
    image:
      url: quay.io/bpfd-bytecode/go-xdp-counter:latest
---
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  containers:
  - name: main
    image: nginx
    volumeMounts:
    - name: bpf-maps
      mountPath: /run/bpfd/fs/maps
      readOnly: true
   volumes:
   - name: bpf-maps
     csi:
       driver: csi.bpfd.dev
       volumeAttributes:
         csi.bpfd.dev/program: go-xdp-counter-example
         csi.bpfd.dev/maps: xdp_stats_map
EOF
```

Which will load a program (xdp-counter) which creates an `xdp_stats_map` and then startup a pod which requests for that map to be mounted in its /run/bpfd/fs for future use.

There's still some TODOs which need to be completed following this PR:

- Move all aya API calls + bpffs creation into the main bpfManager thread which is where any/all privileged operations should take place
- Make this configuration (csi) the default and update all example binaries + deployments to make use of it
- Ensure K8s deployment bpfd privileges are properly fixed.


bpfd/src/command.rs
```diff
@@ -480,6 +480,19 @@ impl Program {
         }
     }
 
+    pub(crate) fn take_data(&self) -> Result<ProgramData, BpfdError> {
```
dave-tucker, line 483: If you're cloning it, you're not taking it.

bpfd/src/storage.rs
```diff
@@ -100,21 +116,147 @@ impl Node for CsiNode {
...
+                        // Find the Program with the specified *Program CRD name
+                        Ok(results) => {
+                            let prog_data = results
+                                .into_iter()
+                                .flat_map(|p| p.take_data())
+                                .find(|d| {
+                                    d.metadata().get(OPERATOR_PROGRAM_KEY) == Some(program_name)
+                                })
+                                .ok_or(Status::new(
+                                    5.into(),
```
dave-tucker, line 158: What is this magic number and why is it a status?

bpfd/src/storage.rs
```diff
@@ -100,21 +116,147 @@ impl Node for CsiNode {
...
+                                    format!("Bpfd Program {program_name} not found"),
+                                ))?;
+
+                            Ok(prog_data
+                                .map_pin_path()
+                                .expect("map pin path should be set since the program is already loaded")
+                                .to_owned())
```
dave-tucker, line 165:
```suggestion
                            let prog_data = results
                                .into_iter()
                                .find(|p| {
                                    if let Ok(data) = p.data() {
                                        data.metadata().get(OPERATOR_PROGRAM_KEY)
                                            == Some(program_name)
                                    } else {
                                        false
                                    }
                                })
                                .ok_or(Status::new(
                                    5.into(),
                                    format!("Bpfd Program {program_name} not found"),
                                ))?;

                            Ok(prog_data
                                .data().expect("program data is valid because we just checked it")
                                .map_pin_path()
                                .expect("map pin path should be set since the program is already loaded")
                                .to_owned())
```

This gets rid of the need for `take_data`.

bpfd/src/storage.rs
```diff
@@ -100,21 +116,147 @@ impl Node for CsiNode {
...
+                // Create the Target Path if it doesn't exist
+                let target = Path::new(target_path);
+                if !target.exists() {
+                    create_dir_all(target).map_err(|e| {
+                        Status::new(
+                            5.into(),
+                            format!("failed creating target path {target_path:?}: {e}"),
+                        )
+                    })?;
+                    set_dir_permissions(target_path, 750).await;
```
dave-tucker, line 187: Another magic number. What's `750`

bpfd/src/storage.rs
```diff
@@ -42,6 +57,7 @@ struct CsiIdentity {
 
 struct CsiNode {
     node_id: String,
+    tx: Arc<Mutex<Sender<Command>>>,
 }
```
dave-tucker, line 61:
```suggestion
    tx: Sender<Command>,
```

This doesn't need to be wrapped in `Arc<Mutex<>>` as the channel is MPSC - there are supposed to be multiple producers. That's true for BpfLoader also.

bpfd/src/utils.rs
```diff
@@ -83,9 +82,25 @@ pub(crate) async fn set_dir_permissions(directory: &str, mode: u32) {
...
 }
 
-pub(crate) fn create_bpffs(directory: &str) -> anyhow::Result<()> {
-    debug!("Creating bpffs at {}", directory);
+pub(crate) fn create_bpffs(directory: &str) -> anyhow::Result<(), anyhow::Error> {
+    debug!("Creating bpffs at {directory}");
     let flags = MsFlags::MS_NOSUID | MsFlags::MS_NODEV | MsFlags::MS_NOEXEC | MsFlags::MS_RELATIME;
     mount::<str, str, str, str>(None, directory, Some("bpf"), flags, None)
-        .context("unable to mount bpffs")
+        .map_err(|e| anyhow::anyhow!("unable to create bpffs at {directory}: {e:?}"))
```
dave-tucker, line 89: `.context()` is better here.
See: https://docs.rs/anyhow/latest/anyhow/trait.Context.html

You can change the message though

bpfd/src/utils.rs
```diff
@@ -83,9 +82,25 @@ pub(crate) async fn set_dir_permissions(directory: &str, mode: u32) {
     }
 }
 
-pub(crate) fn create_bpffs(directory: &str) -> anyhow::Result<()> {
-    debug!("Creating bpffs at {}", directory);
+pub(crate) fn create_bpffs(directory: &str) -> anyhow::Result<(), anyhow::Error> {
```
dave-tucker, line 85: `anyhow::Result<()>` implies there is an anyhow error - no need to add it in the signature.

bpfd/src/utils.rs
```diff
@@ -83,9 +82,25 @@ pub(crate) async fn set_dir_permissions(directory: &str, mode: u32) {
...
+    debug!("Creating bpffs at {directory}");
     let flags = MsFlags::MS_NOSUID | MsFlags::MS_NODEV | MsFlags::MS_NOEXEC | MsFlags::MS_RELATIME;
     mount::<str, str, str, str>(None, directory, Some("bpf"), flags, None)
-        .context("unable to mount bpffs")
+        .map_err(|e| anyhow::anyhow!("unable to create bpffs at {directory}: {e:?}"))
+}
+
+pub(crate) fn unmount(directory: &str) -> anyhow::Result<(), anyhow::Error> {
```
dave-tucker, line 92: nit: Are these used anywhere else? Or will we need to use them anywhere else soon. If not they need to live in `storage.rs`

`creaet_bpffs` is unique since it's also used on startup (for now)

dave-tucker (changes requested):
I've not take a look at the Go code, but Rust-side is almost there - just a few issues to resolve.

bpfd/src/storage.rs
astoycos: :) https://github.com/container-storage-interface/spec/blob/master/spec.md#nodepublishvolume-errors

I'll make these constants

bpfd/src/storage.rs
astoycos: Ah yep looks like find doesn't coerce the type, sweet!

bpfd/src/utils.rs
astoycos: Done!

bpfd/src/utils.rs
astoycos: Done!

bpfd/src/storage.rs
astoycos: Done and also fixed for bpfLoader in a separate commit 👍

dave-tucker (approved):

Issue #430, opened 2023-04-26
Standardize BpfProgram K8s API Condition types.


astoycos:
Currently the available condition types for *Program k8s objects are defined as 

```
	BpfProgCondLoaded                    bpfProgramConditionType = "Loaded"
	BpfProgCondNotLoaded                 bpfProgramConditionType = "NotLoaded"
	BpfProgCondNotUnloaded               bpfProgramConditionType = "NotUnLoaded"
	BpfProgCondNotSelected               bpfProgramConditionType = "NotSelected"
	BpfProgCondUnloaded                  bpfProgramConditionType = "Unloaded"
```

These should be reanalyzed and instead defined explicitly as part of the k8s API rather than within the bpfd-agent.

astoycos:
Assign to @shaneutt

astoycos:
Possibly explore lower level status codes within bpfd internal API

astoycos:
Similar to how [gateway API does things](https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1beta1/gateway_types.go#L510) we should standardize our condition types within the API itself.  More specifically Condition (https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L1481) `Type` and `Reason` should be strongly defined while the `message` can be applied dynamically by the operator

for `BpfProgram` condition types we generally need to express a few cases (I'm sure this isn't exhaustive :)) 

- Program Loaded successfully
- Program Failed to load (with an accompanying message which provides the error to the user)
- Program Failed to Unload (with an accompanying message which provides the error to the user)
- Program Unloaded successfully
- Program was not assigned to this node

shaneutt:
I'm gonna break this down into two PRs:

- [ ] move the conditions into the API package, and rename them accordingly
- [ ] add meaningful reasons to the conditions, and emit them from `reconcileProgram` instead of using static reasons

shaneutt:
#530 only partially completed this, not quite done yet: still need to add more meaningful reasons to the conditions.

shaneutt:
Got that first big chunk of it done, but I wont be able to jump right on the next bit right away so I figure I'll unassign myself at least temporarily in case someone else wants to jump in here.

astoycos:
[...]

PR #734, opened 2023-09-26
Allow bpf programs with errors to be deleted in k8s

### LINKED ISSUE #733: Can't delete programs with errors in k8s
If you try to add a program, but there is a problem that prevents bpfd from loading it, you can't delete it or update it.

An example of such an error is a typo in the image url in the program spec.

anfredette:
Fixes #733, "Can't delete programs with errors in k8s"

The problem was that if a program was being deleted, and it wasn't loaded for some reason (most likely due to an error), the code returned BpfProgCondNotLoaded.  While this is true, the problem is that the code treats this condition as an error, so the delete never completes.  The fix is to return BpfProgCondUnloaded, which is what the code is expecting to see when a program is being deleted.

Also fixed a problem with the sample tracepoint yaml.

### DIFF
--- a/bpfd-operator/config/samples/bpfd.io_v1alpha1_tracepoint_tracepointprogram.yaml
+++ b/bpfd-operator/config/samples/bpfd.io_v1alpha1_tracepoint_tracepointprogram.yaml
@@ -5,7 +5,7 @@ metadata:
     app.kubernetes.io/name: tracepointprogram
   name: tracepoint-example
 spec:
-  sectionname: sys_enter_openat
+  sectionname: enter_openat
   # Select all nodes
   nodeselector: {}
   names: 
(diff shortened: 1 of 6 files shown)


Billy99:
Can you add a test (in this PR or write an issue)? I feel like I fixed a similar problem 6-8 months ago where the program was created, but not loaded, and could not be deleted.

anfredette:
Good point.  I'll add a test.

astoycos:
Ahh interesting and good catch @anfredette Hopefully it will be harder to make this type of mistake when we fixup the conditions.

anfredette:
@Billy99 @astoycos I think I'll need to do an integration test for adding and deleting a program with an error (as opposed to a unit test) because it involves an interaction between the agent and the operator.  I also decided to make another change to the way we handle conditions while I was in the code.  It doesn't go all the way to addressing issue #430, but I think it's a simple step in the right direction.  I need to clean it up a little before submitting it, but I based it on this change.  So, if it's okay with you guys, I'd like to merge this fix first, then I'll submit the conditions change, and then I'll write an integration test for issue #734.

Billy99 (approved):
LGTM

anfredette:
Thanks @Billy99

Issue #725, opened 2023-09-19
Scrub the bpfctl commands and output in the docs


anfredette:
Some things have changed including (but not necessarily limited to) the load commands and list/get output.

Billy99:
#670 is going to change some of the `bpfctl` output, so I'll do a scrub associated with those changes if no gets to this first.

PR #736, opened 2023-09-27
Update docs and helptext

Billy99:
Perform a scrub of bpfctl helptext and after the recent changes

Resolves: #725


bpfctl/src/main.rs
```diff
@@ -69,11 +69,15 @@ struct ListArgs {
     #[clap(short, long, verbatim_doc_comment, hide_possible_values = true)]
     program_type: Option<ProgramType>,
 
-    // Optional: List programs which contain a specific set of metadata labels
+    /// Optional: List programs which contain a specific set of metadata labels
+    /// that were applied when program was loaded with `--metadata` parameter.
```
astoycos, line 73:
```suggestion
    /// that were applied when the program was loaded with `--metadata` parameter.
```

bpfctl/src/main.rs
```diff
@@ -85,7 +89,7 @@ struct LoadFileArgs {
     #[clap(short, long, verbatim_doc_comment)]
     path: String,
 
-    /// Required: Name of the ELF section from the object file.
+    /// Required: The name of the function that is the entry point for the BPF program..
```
astoycos, line 92:
```suggestion
    /// Required: The name of the function that is the entry point for the BPF program.
```

bpfctl/src/main.rs
Billy99: done

bpfctl/src/main.rs
Billy99: Done

astoycos (approved):
LGTM

Thanks @Billy99!

Issue #630, opened 2023-08-03
Remove dependency on cert-manager


dave-tucker:
This will simplify the OLM installation process

PR #737, opened 2023-09-27
Remove Cert-manager as a dep

astoycos:
Allow the kubernetes based bpfd-agent to communicate with bpfd via an unencrypted unix socket.

- Updated bpfd to always set the permissions if it's listening for incoming connections on a unix socket.
- Update the bpfd-agent to use the bpfdconfig for setting up the desired unix socket location.
- Allow encryption to occur over TLS OR Unix.
- Fixup the default deployment manifests to share the new socket.
- Reduce permissions of the bpfd-agent when deployed on kubernetes.


astoycos:
fixes #630

bpfd-operator/internal/conn/conn.go
```diff
@@ -54,28 +53,29 @@ type Grpc struct {
...
-		Grpc: Grpc{
-			Endpoints: []Endpoint{
-				{
-					Type:    internal.DefaultType,
-					Port:    internal.DefaultPort,
-					Enabled: internal.DefaultEnabled,
-				},
-			},
-		},
+		// Don't default this
```
astoycos, line 62: nit:  remove comment

bpfd-operator/config/bpfd-deployment/daemonset-csi.yaml
```diff
@@ -27,6 +27,10 @@ spec:
       - key: node-role.kubernetes.io/master
         operator: Exists
         effect: NoSchedule
+      # Specify FS group so that the agent can use the bpfd socket
+      # without needing to be root.
+      securityContext:
+        fsGroup: 2000
```
Billy99, line 33: I read https://kubernetes.io/docs/tasks/configure-pod-container/security-context/ on the FS Group, but it just gave an example of 2000, didn't explain where 2000 came from. Is that just standard practice to use 2000?

bpfd-operator/pkg/helpers/helpers.go
```diff
@@ -169,13 +171,43 @@ func GetMaps(c *bpfdclientset.Clientset, ProgramName string, mapNames []string)
 
 	prog := bpfProgramList.Items[0]
 
+	id, ok := prog.Annotations[internal.IdAnnotation]
+	if !ok {
+		return nil, fmt.Errorf("BpfProgram %s does not have a program id", ProgramName)
+	}
+
+	maps := map[string]string{}
+	programMapPath := fmt.Sprintf("%s/%s", internal.BpfdMapFs, id)
```
Billy99, line 180: This doesn't work if this program is sharing another program's maps.

Billy99 (changes requested):

Billy99:
We should probably update the documentation somewhere regarding the different version of `make run-on-kind` and provide some guidance on when to use which.

bpfd/src/serve.rs
```diff
@@ -150,7 +145,12 @@ async fn serve_unix(
 
     let uds = UnixListener::bind(&path)?;
     let uds_stream = UnixListenerStream::new(uds);
-    set_file_permissions(&path, SOCK_MODE).await;
+    // Always set the file permissions of our listening socket.
+    if (tokio::fs::set_permissions(path.clone(), std::fs::Permissions::from_mode(SOCK_MODE)).await)
```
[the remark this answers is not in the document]
astoycos: The function only sets permissions on the socket if it's in the bpfd user group https://github.com/bpfd-dev/bpfd/blob/main/bpfd/src/utils.rs#L63

We should ALWAYS be setting the socket perms correctly since it's used by processes other than bpfd.

bpfd-operator/config/bpfd-deployment/daemonset-csi.yaml
astoycos: Good question I used it completely arbitrarily, do you have something else you'd rather have it be?

bpfd-operator/pkg/helpers/helpers.go
astoycos: good point... how did that work before though?

astoycos:
> We should probably update the documentation somewhere regarding the different version of make run-on-kind and provide some guidance on when to use which.

To be honest I'm trying to keep any new documentation as minimal as possible just because it's been alot of work to maintain.  For this specifically my ultimate goal is to remove/ deprecate the `run-on-kind-encrypted` target when we get to https://github.com/bpfd-dev/bpfd/issues/741

Billy99:
> > We should probably update the documentation somewhere regarding the different version of make run-on-kind and provide some guidance on when to use which.
> 
> To be honest I'm trying to keep any new documentation as minimal as possible just because it's been alot of work to maintain. For this specifically my ultimate goal is to remove/ deprecate the `run-on-kind-encrypted` target when we get to #741

I'm fine with limiting it because it will change pretty soon, but not OK because it is a lot of work. That's the price you pay for good documentation.

bpfd-operator/pkg/helpers/helpers.go
Billy99: I guess it didn't.

bpfd-operator/config/bpfd-deployment/daemonset-csi.yaml
Billy99: I am not familiar with it, so I was more asking.

astoycos:
Not sure why the golint is failing here :/

bpfd-operator/pkg/helpers/helpers.go
Billy99: Blocked by #744. Once ProgramData is written to annotations, the MapPinPath will be there and the path doesn't need to be calculated.

Billy99 (approved):
LGTM

Issue #710, opened 2023-09-15
deduplicate bpfd operator and agent test code.


anfredette:
Much of the code in the following files is the same.  

```bash
bpfd-operator/controllers/bpfd-agent/*-program_test.go 
bpfd-operator/controllers/bpfd-operator/*-program_test.go
```

Try to create common test functions that can be used for all program types.

astoycos:
Are you working on this currently @anfredette?

anfredette:
Yes.

Issue #709, opened 2023-09-15
deduplicate k8s agent reconcile code


anfredette:
For example, can we make reconcileBpfdProgram() common for all program types?

PR #738, opened 2023-09-27
Just keep one condition for each program object

anfredette:
All of the code assumes that only the last condition counts.  However, that's not how conditions work.  This change modifies our behavior so that we only maintain the one (most recent) condition for each program object, which is a step toward the resolution of issue #430.

Part of the solution is to create a common updateCondition() function for the bpfd-operator that is used for all program types which removes some code duplication and is a small step toward the resolution of issue #709

This also removes code duplication in xdp-program_test.go, which is a small step toward the resolution of issue #710.

### DIFF
--- a/bpfd-operator/controllers/bpfd-agent/common.go
+++ b/bpfd-operator/controllers/bpfd-agent/common.go
@@ -161,25 +161,46 @@ func (r *ReconcilerCommon) removeFinalizer(ctx context.Context, o client.Object,
 // false if the status was already set for the given bpfProgram, meaning reconciliation
 // may continue.
 func (r *ReconcilerCommon) updateStatus(ctx context.Context, prog *bpfdiov1alpha1.BpfProgram, cond bpfdiov1alpha1.BpfProgramConditionType) bool {
-	if prog.Status.Conditions != nil {
-		// Get most recent condition
-		recentIdx := len(prog.Status.Conditions) - 1
 
-		// If the most recent condition is the same as input, just return.
-		if prog.Status.Conditions[recentIdx].Type == string(cond) {
-			return false
-		} else {
-			// Remove the input condition from the list if it exists (may not exist)
-			// because the SetStatusCondition() doesn't append if it is already in the list.
-			meta.RemoveStatusCondition(&prog.Status.Conditions, string(cond))
+	r.Logger.V(1).Info("updateStatus()", "existing conds", prog.Status.Conditions, "new cond", cond)
+
+	if prog.Status.Conditions != nil {
+		r.Logger.V(1).Info("conditions != nil")
+		numConditions := len(prog.Status.Conditions)
+
+		switch {
+		case numConditions == 1:
+			{
+				if prog.Status.Conditions[0].Type == string(cond) {
+					r.Logger.V(1).Info("conditions match")
+					// No change, so just return false -- not updated
+					return false
+				} else {
+					// We're changing the condition, so delete this one.  The
+					// new condition will be added below.
+					r.Logger.V(1).Info("removing condition", "bpfProgram", prog.Name, "condition", prog.Status.Conditions[0].Type)
+					meta.RemoveStatusCondition(&prog.Status.Conditions, prog.Status.Conditions[0].Type)
+				}
+			}
+		case numConditions > 1:
+			{
+				// We should only ever have one condition, so we shouldn't hit
+				// this case.  However, if we do, delete the existing conditions
+				// and add the new one below.
+				r.Logger.V(1).Info("more than one BpfProgramCondition", "numConditions", numConditions)
+				for _, c := range prog.Status.Conditions {
+					r.Logger.Info("removing condition", "bpfProgram", prog.Name, "condition", c.Type)
+					meta.RemoveStatusCondition(&prog.Status.Conditions, c.Type)
+				}
+			}
 		}
 	}
 
 	meta.SetStatusCondition(&prog.Status.Conditions, cond.Condition())
-	r.Logger.Info("Updating bpfProgram condition", "bpfProgram", prog.Name, "condition", cond.Condition().Type)
+
+	r.Logger.V(1).Info("Updating bpfProgram condition", "bpfProgram", prog.Name, "condition", cond.Condition().Type)
 	if err := r.Status().Update(ctx, prog); err != nil {
-		r.Logger.Error(err, "failed to set bpfProgram object status")
-		return true
+		r.Logger.V(1).Error(err, "failed to set bpfProgram object status")
 	}
 
 	return true
--- a/bpfd-operator/controllers/bpfd-agent/common.go
+++ b/bpfd-operator/controllers/bpfd-agent/common.go
@@ -165,44 +165,35 @@ func (r *ReconcilerCommon) updateStatus(ctx context.Context, prog *bpfdiov1alpha
 	r.Logger.V(1).Info("updateStatus()", "existing conds", prog.Status.Conditions, "new cond", cond)
 
 	if prog.Status.Conditions != nil {
-		r.Logger.V(1).Info("conditions != nil")
 		numConditions := len(prog.Status.Conditions)
 
-		switch {
-		case numConditions == 1:
-			{
-				if prog.Status.Conditions[0].Type == string(cond) {
-					r.Logger.V(1).Info("conditions match")
-					// No change, so just return false -- not updated
-					return false
-				} else {
-					// We're changing the condition, so delete this one.  The
-					// new condition will be added below.
-					r.Logger.V(1).Info("removing condition", "bpfProgram", prog.Name, "condition", prog.Status.Conditions[0].Type)
-					meta.RemoveStatusCondition(&prog.Status.Conditions, prog.Status.Conditions[0].Type)
-				}
+		if numConditions == 1 {
+			if prog.Status.Conditions[0].Type == string(cond) {
+				// No change, so just return false -- not updated
+				return false
+			} else {
+				// We're changing the condition, so delete this one.  The
+				// new condition will be added below.
+				prog.Status.Conditions = nil
 			}
-		case numConditions > 1:
-			{
-				// We should only ever have one condition, so we shouldn't hit
-				// this case.  However, if we do, delete the existing conditions
-				// and add the new one below.
-				r.Logger.V(1).Info("more than one BpfProgramCondition", "numConditions", numConditions)
-				for _, c := range prog.Status.Conditions {
-					r.Logger.Info("removing condition", "bpfProgram", prog.Name, "condition", c.Type)
-					meta.RemoveStatusCondition(&prog.Status.Conditions, c.Type)
-				}
-			}
+		} else if numConditions > 1 {
+			// We should only ever have one condition, so we shouldn't hit this
+			// case.  However, if we do, log a message, delete the existing
+			// conditions, and add the new one below.
+			r.Logger.Info("more than one BpfProgramCondition", "numConditions", numConditions)
+			prog.Status.Conditions = nil
 		}
+		// if numConditions == 0, just add the new condition below.
 	}
 
 	meta.SetStatusCondition(&prog.Status.Conditions, cond.Condition())
 
 	r.Logger.V(1).Info("Updating bpfProgram condition", "bpfProgram", prog.Name, "condition", cond.Condition().Type)
 	if err := r.Status().Update(ctx, prog); err != nil {
-		r.Logger.V(1).Error(err, "failed to set bpfProgram object status")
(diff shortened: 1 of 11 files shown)


bpfd-operator/controllers/bpfd-agent/common.go
```diff
@@ -164,25 +164,50 @@ func (r *ReconcilerCommon) removeFinalizer(ctx context.Context, o client.Object,
...
+					// We're changing the condition, so delete this one.  The
+					// new condition will be added below.
+					r.Logger.V(1).Info("removing condition", "bpfProgram", prog.Name, "condition", prog.Status.Conditions[0].Type)
+					meta.RemoveStatusCondition(&prog.Status.Conditions, prog.Status.Conditions[0].Type)
+				}
+			}
+		case numConditions == 0:
+			{
+				// Fall through and add the new condition.
+			}
```
astoycos, line 191: nit remove this, just use `default:` and print 	`r.Logger.V(1).Info("Updating bpfProgram condition", "bpfProgram", prog.Name, "condition", cond.Condition().Type)`

bpfd-operator/controllers/bpfd-agent/xdp-program_test.go
```diff
@@ -43,184 +44,34 @@ import (
...
+		bpfProgName0 = fmt.Sprintf("%s-%s-%s", name, fakeNode.Name, fakeInt0)
+		bpfProgName1 = fmt.Sprintf("%s-%s-%s", name, fakeNode.Name, fakeInt1)
 		bpfProgEth0  = &bpfdiov1alpha1.BpfProgram{}
 		bpfProgEth1  = &bpfdiov1alpha1.BpfProgram{}
 		fakeUID0     = "ef71d42c-aa21-48e8-a697-82391d801a80"
 		fakeUID1     = "ef71d42c-aa21-48e8-a697-82391d801a81"
 	)
+
+	var fakeInts []string
+	if multiInterface {
```
astoycos, line 69: This if statement always does the same thing :/

bpfd-operator/controllers/bpfd-agent/xdp-program_test.go
```diff
@@ -424,43 +198,153 @@ func TestXdpProgramControllerCreateMultiIntf(t *testing.T) {
...
+
+		// Check that the bpfProgram's status was correctly updated
+		err = cl.Get(ctx, types.NamespacedName{Name: bpfProgName1, Namespace: metav1.NamespaceAll}, bpfProgEth1)
+		require.NoError(t, err)
+
+		require.Equal(t, string(bpfdiov1alpha1.BpfProgCondLoaded), bpfProgEth1.Status.Conditions[0].Type)
+	}
+}
+
+func TestXdpProgramControllerCreate(t *testing.T) {
```
astoycos, line 340: Don't we need to make these same changes for the tc program unit tests as well?

bpfd-operator/controllers/bpfd-operator/common.go
```diff
@@ -78,14 +80,14 @@ func reconcileBpfProgram(ctx context.Context, rec ProgramReconciler, prog client
...
-		r.Logger.Error(err, "failed to get freshPrograms for full reconcile")
+		r.Logger.V(1).Error(err, "failed to get freshPrograms for full reconcile")
 		return ctrl.Result{}, nil
 	}
 
 	// List all nodes since a bpfprogram object will always be created for each
 	nodes := &corev1.NodeList{}
 	if err := r.List(ctx, nodes, &client.ListOptions{}); err != nil {
-		r.Logger.Error(err, "failed getting nodes for full reconcile")
+		r.Logger.V(1).Error(err, "failed getting nodes for full reconcile")
```
astoycos, line 90: This is the only place where we see this error, I don't think you should bump these to V(1)

bpfd-operator/controllers/bpfd-operator/common.go
```diff
@@ -188,3 +190,50 @@ func statusChangedPredicate() predicate.Funcs {
 		},
 	}
 }
+
+func (r *ReconcilerCommon) updateCondition(ctx context.Context, obj client.Object, conditions *[]metav1.Condition, cond bpfdiov1alpha1.ProgramConditionType, message string) (ctrl.Result, error) {
```
astoycos, line 194: Can you move this and the one in bpfd-agent/common.go into `/bpfd-operator/internal/k8s.go` ?

bpfd-operator/controllers/bpfd-operator/common.go
```diff
@@ -188,3 +190,50 @@ func statusChangedPredicate() predicate.Funcs {
...
+					meta.RemoveStatusCondition(conditions, c.Type)
+				}
+			}
+		}
+	}
+
+	meta.SetStatusCondition(conditions, cond.Condition(message))
+
+	if err := r.Status().Update(ctx, obj); err != nil {
+		r.Logger.V(1).Info("failed to set KprobeProgram object status...requeuing")
```
astoycos, line 234:
```suggestion
		r.Logger.V(1).Info("failed to set *Program object status...requeuing")
```

bpfd-operator/controllers/bpfd-operator/kprobe-program_test.go
```diff
@@ -137,6 +140,21 @@ func TestKprobeProgramReconcile(t *testing.T) {
...
+		meta.SetStatusCondition(&Kprobe.Status.Conditions, bpfdiov1alpha1.ProgramDeleteError.Condition("bogus condition #1"))
+		if err := r.Status().Update(ctx, Kprobe); err != nil {
+			r.Logger.V(1).Info("failed to set KprobeProgram object status")
+		}
+		meta.SetStatusCondition(&Kprobe.Status.Conditions, bpfdiov1alpha1.ProgramReconcileError.Condition("bogus condition #2"))
+		if err := r.Status().Update(ctx, Kprobe); err != nil {
+			r.Logger.V(1).Info("failed to set KprobeProgram object status")
+		}
+		// Make sure we have 2 conditions
+		require.Equal(t, 2, len(Kprobe.Status.Conditions))
```
astoycos, line 155: Wait shouldn's we only have one condition?

astoycos (changes requested):
just some small stuff!! Let's condense those functions that the agent + operator are using to one central location if possible.

bpfd-operator/controllers/bpfd-agent/xdp-program_test.go
anfredette: We could, but I wasn't changing the tc cases, so I didn't do it.

bpfd-operator/controllers/bpfd-agent/xdp-program_test.go
anfredette: oops. fixed.

bpfd-operator/controllers/bpfd-operator/common.go
anfredette: good catch

bpfd-operator/controllers/bpfd-operator/kprobe-program_test.go
anfredette: This is a forced error test case where I explicitly added 2 conditions. This should never happen with the current code, but I wanted to handle it in case someone messes it up in the future.

bpfd-operator/controllers/bpfd-operator/common.go
anfredette: fixed

bpfd-operator/controllers/bpfd-operator/kprobe-program_test.go
anfredette: Also, since all the program types run the same code, I just randomly picked one in which to test this error path.

astoycos:
Shoot I forgot about https://github.com/bpfd-dev/bpfd/blob/main/bpfd-operator/pkg/helpers/helpers.go#L308

can you update all those helpers to not play the "look for last position" dance?

anfredette:
> Shoot I forgot about https://github.com/bpfd-dev/bpfd/blob/main/bpfd-operator/pkg/helpers/helpers.go#L308
> 
> can you update all those helpers to not play the "look for last position" dance?

Thanks for pointing that out.  I updated the code and moved the repeated condition-checking logic into an isProgLoaded() function.

bpfd-operator/pkg/helpers/helpers.go
```diff
@@ -305,6 +305,31 @@ func GetMaps(c *bpfdclientset.Clientset, ProgramName string, mapNames []string)
...
+// Returns true if loaded.  False if not.  Also returns the condition type.
+func isProgLoaded(conditions *[]metav1.Condition) (bool, string) {
+	// Get most recent condition
+	conLen := len(*conditions)
+
+	if conLen <= 0 {
+		return false, "None"
+	}
+
+	if conLen > 1 {
```
astoycos, line 317: We've decided in our core design that we should only ever have one condition. 

I don't think we should be "assuming" this may not hold true in a public library that we've designed for people to use. 

i.e this function should return the single condition `conditions[0]` and return (false, error)  or (false, "None") otherwise

astoycos (changes requested):

bpfd-operator/controllers/bpfd-operator/common.go
```diff
@@ -188,3 +190,46 @@ func statusChangedPredicate() predicate.Funcs {
...
+					meta.RemoveStatusCondition(conditions, (*conditions)[0].Type)
+				}
+			}
+		case numConditions > 1:
+			{
+				// We should only ever have one condition, so we shouldn't hit
+				// this case.  However, if we do, delete the existing conditions
+				// and add the new one below.
+				r.Logger.Info("more than one BpfProgramCondition", "numConditions", numConditions)
+				for _, c := range *conditions {
```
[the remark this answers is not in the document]
astoycos: ^^ is just https://github.com/kubernetes/apimachinery/blob/v0.28.2/pkg/api/meta/conditions.go#L30 with a single line changed to make the operation idempotent rather than additive

bpfd-operator/pkg/helpers/helpers.go
anfredette: That's what it's doing -- I'm just logging a message if the condition exists.  However, I'm okay with removing the check and log here in the public library.

bpfd-operator/pkg/helpers/helpers.go
```diff
@@ -437,10 +426,10 @@ func WaitForBpfProgConfLoad(c *bpfdclientset.Clientset, progName string, timeout
 		return wait.PollImmediate(time.Second, timeout, isXdpbpfdProgLoaded(c, progName))
 	case Tracepoint:
 		return wait.PollImmediate(time.Second, timeout, isTracepointbpfdProgLoaded(c, progName))
-	// case Uprobe: not covered because it doesn't have a ProgramType.  However,
-	// the only case currently used is Xdp, so it's not needed now.  If we need
-	// to support uprobes in the future, we could replace ProgramType with
-	// something else, like the string representation of the program type.
+	// TODO: case Uprobe: not covered.  Since Uprobe has the same ProgramType as
```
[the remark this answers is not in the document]
anfredette: Good idea.  Now or later?

bpfd-operator/controllers/bpfd-operator/common.go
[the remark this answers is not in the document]
anfredette: First of all, this should never happen, so we shouldn't execute this code and there shouldn't be any logs.  However, in my experience, I've seen many bugs in code that had comments saying "this should never happen".  If you ignore the possiblities, you wind up having bugs that are very hard to find.  And this issue isn't with how it works now.  The issue would happen a year or two from now when others are working on the project and someone decides it would be good to report a condition for something new and it messes up how the core logic works.  So, this is defensive programming.  If it happens, we log a message and clean things up.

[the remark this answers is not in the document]
anfredette: I think we can remove the debug logs that say we're removing the condtions since we already have the debug log at the beginning that reports the existing and new conditions.  I think it's good to at least have this log so we can understand how the state of the objects is changing if something goes wrong.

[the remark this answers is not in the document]
astoycos: OK i'm alright with it I just didn't like the extra loop :)

astoycos (approved):
LGTM 

Barring response to Billy's review comments + his LGTM

anfredette:
I changed the logs and made a few other changes too.

anfredette:
I removed the hold.  It's waiting on @Billy99 to ack the changes.

Billy99 (approved):
LGTM

PR #754, opened 2023-10-03
ci: build all bytecode images in ci

### LINKED ISSUE #716: Sign xdp_pass, tc_pass etc...
These images are not signed as they don't go through the same GH actions pipeline.
They probably should.

astoycos:
Add build commands to automatically build/sign the rest of the bytecode images in github actions.  All bpfPrograms that we maintain bpf code for in our repository will now be built.

fixes #716


Billy99 (approved):
LGTM

PR #757, opened 2023-10-04
Fix the map_used_by field for managing the  map_pin_path

Billy99:
The map_used_by field is not being set properly. Each program should be listing the map it is using and all the other programs that are using it.

### DIFF
--- a/bpfd/src/bpf.rs
+++ b/bpfd/src/bpf.rs
@@ -21,7 +21,7 @@ use bpfd_api::{
     ProbeType::{self, *},
     ProgramType, TcProceedOn,
 };
-use log::{debug, info};
+use log::{debug, info, warn};
 use tokio::{
     fs, select,
     sync::{
@@ -196,7 +196,7 @@ impl BpfManager {
             // TODO: Should probably check for pinned prog on bpffs rather than assuming they are attached
             program.set_attached();
             debug!("rebuilding state for program {}", id);
-            self.rebuild_map_entry(id, program.data()?.map_owner_id());
+            self.rebuild_map_entry(id, &mut program);
             self.programs.insert(id, program);
         }
         self.rebuild_dispatcher_state(ProgramType::Xdp, None, RTDIR_XDP_DISPATCHER)
@@ -298,9 +298,6 @@ impl BpfManager {
         // Program bytes MUST be cleared after load.
         program.data_mut()?.clear_program_bytes();
 
-        // map_pin_path MUST be set following load.
-        let map_pin_path = program.data()?.map_pin_path();
-
         match result {
             Ok(id) => {
                 info!(
@@ -311,18 +308,21 @@ impl BpfManager {
 
                 // Now that program is successfully loaded, update the id, maps hash table,
                 // and allow access to all maps by bpfd group members.
-                self.save_map(
-                    id,
-                    map_owner_id,
-                    map_pin_path.expect("map_pin_path must be set after successful load"),
-                )
-                .await?;
+                self.save_map(&mut program, id, map_owner_id).await?;
+
+                // Only add program to bpfManager if we've completed all mutations and it's successfully loaded.
+                self.programs.insert(id, program.to_owned());
 
                 Ok(program)
             }
             Err(e) => {
-                if let Some(pin_path) = map_pin_path {
-                    let _ = self.cleanup_map_pin_path(pin_path, map_owner_id).await;
+                // Cleanup any directories associated with the map_pin_path.
+                // Data and map_pin_path may or may not exist depending on where the original
+                // error occured, so don't error if not there and preserve original error.
+                if let Ok(data) = program.data() {
+                    if let Some(pin_path) = data.map_pin_path() {
+                        let _ = self.cleanup_map_pin_path(pin_path, map_owner_id).await;
+                    }
                 }
                 Err(e)
             }
@@ -410,13 +410,11 @@ impl BpfManager {
             .kernel_info()
             .expect("kernel info should be set after load")
             .id;
-        self.programs.insert(id, program.to_owned());
 
-        if let Some(p) = self.programs.get_mut(&id) {
-            p.set_attached();
-            p.save(id)
-                .map_err(|e| BpfdError::Error(format!("unable to save program state: {e}")))?;
-        };
+        program.set_attached();
+        program
+            .save(id)
+            .map_err(|e| BpfdError::Error(format!("unable to save program state: {e}")))?;
 
         Ok(id)
     }
@@ -599,11 +597,7 @@ impl BpfManager {
                     }
                 }
 
-                self.programs.insert(id, p.to_owned());
-                self.programs
-                    .get(&id)
-                    .unwrap()
-                    .save(id)
+                p.save(id)
                     // we might want to log or ignore this error instead of returning here...
                     // because otherwise it will hide the original error (from res above)
                     .map_err(|_| BpfdError::Error("unable to persist program data".to_string()))?;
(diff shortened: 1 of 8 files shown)


bpfd/src/bpf.rs
```diff
@@ -196,8 +196,14 @@ impl BpfManager {
             // TODO: Should probably check for pinned prog on bpffs rather than assuming they are attached
             program.set_attached();
             debug!("rebuilding state for program {}", id);
-            self.rebuild_map_entry(id, program.data()?.map_owner_id());
+            let map_owner_id = match program.data() {
```
astoycos, line 199: why do we need this map can't you just move

`self.rebuild_map_entry(id, program.data()?.map_owner_id());`

below the insert?

Additionally this is now ignoring the error is that what we want?

bpfd/src/bpf.rs
```diff
@@ -311,6 +317,18 @@ impl BpfManager {
                     map_pin_path.expect("map_pin_path must be set after successfult load"),
                 )
                 .await?;
+
+                // Update the local copy with the modified map_used_by
+                if let Some(prog) = self.programs.get_mut(&id) {
+                    let map_used_by = match prog.data() {
```
astoycos, line 323: If we actually want to ignore the error here (and above) just do 

` let map_used_by = prog.data().ok().map(|data| data.maps_used_by());
`

bpfd/src/bpf.rs
```diff
@@ -311,6 +317,18 @@ impl BpfManager {
                     map_pin_path.expect("map_pin_path must be set after successfult load"),
                 )
                 .await?;
+
+                // Update the local copy with the modified map_used_by
+                if let Some(prog) = self.programs.get_mut(&id) {
```
astoycos, line 322: Wait we shouldn't have to do a `self.programs.get_mut(&id)` here the `program` var is a pointer to the actual program we just loaded.

bpfd/src/bpf.rs
```diff
@@ -913,31 +931,33 @@ impl BpfManager {
...
                 } else {
                     return Err(BpfdError::Error("map_owner_id does not exists".to_string()));
                 }
             }
             None => {
-                let program = self
-                    .programs
-                    .get_mut(&id)
-                    .expect("Program should be loaded");
+                if let Some(program) = self.programs.get_mut(&id) {
```
astoycos, line 947: why remove the expect here? If map_owner_id isn't set then the program in question is the owner and __should__ have already been loaded right?

bpfd/src/bpf.rs
```diff
@@ -913,31 +931,33 @@ impl BpfManager {
...
 
-                self.maps.insert(id, map);
-
-                // TODO(astoycos) remove external self.maps, keep all map tracking info in Program
-                // update programs map_used_by
-                program.data_mut()?.set_maps_used_by(Some(vec![id]));
+                    // Update this program with the updated map_used_by
+                    if let Ok(data) = program.data_mut() {
+                        data.set_maps_used_by(Some(vec![id]));
+                    }
```
astoycos, line 955: program.data_mut().ok().map(|data| data.set_maps_used_by(Some(vec![id])))

bpfd/src/bpf.rs
```diff
@@ -913,31 +931,33 @@ impl BpfManager {
...
+                }
 
-                set_dir_permissions(map_pin_path.to_str().unwrap(), MAPS_MODE).await;
+                if let Some(path) = map_pin_path.to_str() {
```
astoycos, line 958: Wait we should panic if we can't convert the map_pin_path to a string right? (i.e that shouldn't happen)

bpfd/src/bpf.rs
```diff
@@ -961,21 +981,22 @@ impl BpfManager {
...
                 fs::remove_dir_all(path)
                     .await
                     .map_err(|e| BpfdError::Error(format!("can't delete map dir: {e}")))?;
+            } else {
+                // Update all the programs still using the same map with the updated map_used_by.
+                for id in map.used_by.iter() {
+                    if let Some(program) = self.programs.get_mut(id) {
+                        if let Ok(data) = program.data_mut() {
+                            data.set_maps_used_by(Some(map.used_by.clone()));
+                        }
```
astoycos, line 997: Change into a on-liner as above

bpfd/src/bpf.rs
```diff
@@ -992,9 +1013,25 @@ impl BpfManager {
 
         if let Some(map) = self.maps.get_mut(&index) {
             map.used_by.push(id);
+
+            // Update all the programs using the same map with the updated map_used_by.
+            for used_by_id in map.used_by.iter() {
+                if let Some(program) = self.programs.get_mut(used_by_id) {
+                    if let Ok(data) = program.data_mut() {
+                        data.set_maps_used_by(Some(map.used_by.clone()));
+                    }
```
astoycos, line 1022: one liner ^^^

bpfd/src/bpf.rs
```diff
@@ -992,9 +1013,25 @@ impl BpfManager {
...
+            }
         } else {
             let map = BpfMap { used_by: vec![id] };
             self.maps.insert(index, map);
+
+            // Update this programs with the updated map_used_by.
+            if let Some(program) = self.programs.get_mut(&id) {
+                if let Ok(data) = program.data_mut() {
+                    data.set_maps_used_by(Some(vec![id]));
+                }
```
astoycos, line 1033: one liner ^^^

bpfd/src/command.rs
```diff
@@ -331,6 +331,8 @@ pub(crate) struct ProgramData {
     // populated after load
     kernel_info: Option<KernelProgramInfo>,
     map_pin_path: Option<PathBuf>,
+
+    // Only populated on get or list calls.
```
astoycos, line 335: Wait is this true? We're doing all of the populating logic nested in add_program right?

bpfd/src/bpf.rs
```diff
@@ -913,31 +931,33 @@ impl BpfManager {
                 if let Some(map) = self.maps.get_mut(&m) {
                     map.used_by.push(id);
 
-                    // If map owner program still exists update it's maps_used_by_field
-                    if let Some(program) = self.programs.get_mut(&m) {
-                        program
-                            .data_mut()?
-                            .set_maps_used_by(Some(map.used_by.clone()));
+                    // Update all the programs using the same map with the updated map_used_by.
```
astoycos, line 934: I'm lost here for some reason....Say we have program A, B, and C Programs B/C are using the maps from Program A

When we add Program B all we have to do is update program A's maps_used_by right? 

So shouldn't this just be what I had before? 

Isn't this is now saying that programs A, B, AND C will all have the same "maps_used_by" field which makes no sense because only program A is the owner.

astoycos (changes requested):
Some questions here!

bpfd/src/bpf.rs
astoycos: Basically id map_owner_id is set I would expect maps_used_by to be NONE

Billy99: > When we add Program B all we have to do is update program A's maps_used_by right?
> So shouldn't this just be what I had before?

Yes, this is what we had before. And yes, `this is now saying that programs A, B, AND C will all have the same "maps_used_by" field`. With your implementation, remove Program A. How does the user know what programs are using Program A's maps now that it is gone? The user is flying blind at that point. Once we have the map object exposed to the user, this field can go away.

astoycos: I mean the user would just see "oh the map_owner_id doesn't actually exist anymore it must have been deleted" but I gotcha and am alright with it. Thanks for reminding me :)

bpfd/src/bpf.rs
Billy99: clippy and fmt forced it back to what I had originally.

bpfd/src/bpf.rs
Billy99: clippy and fmt forced it back to what I had originally.

bpfd/src/bpf.rs
Billy99: clippy and fmt forced it back to what I had originally.

bpfd/src/bpf.rs
astoycos: Then just rebuild the map entry before inserting into programs right? That __should__ allow you to get past the borrow errors

bpfd/src/bpf.rs
astoycos: Agreed but let's return an error then OR print a warning... currently we just do nothing except try and set the dir permissions even if the program isn't found.

bpfd/src/bpf.rs
```diff
@@ -311,7 +317,17 @@ impl BpfManager {
                     map_pin_path.expect("map_pin_path must be set after successfult load"),
                 )
                 .await?;
-                Ok(program)
+
+                // add_multi_attach_program() and add_single_attach_program() take ownership
+                // of program, so this function now has an outdated copy (for example, map_used_by
+                // is not set). Get the latest copy to return.
```
astoycos, line 323: This comment/diff is incorrect both add_single/multi_attach_program take mutable references to programs (&mut program)

So therefore after attaching map_pin_path and map_used_by should be set. See the existing logic on line 308 where we get the map pin path (which was set in the attach functions)

bpfd/src/bpf.rs
astoycos: I can hear the argument for not panicing but with this we're doing nothing if the `to_str()` fails

bpfd/src/bpf.rs
astoycos: We might need to consider calling save_map BEFORE inserting the program into the dispatcher 

regardless if the `program` variable in `add_program` is not up to date with what actually got inserted into `self.programs` here we're doing something wrong and should try and figure it out.  I really don't want to hack around this as we just spent a bunch of time refactoring the code to avoid getting/resetting self.programs

Billy99: So `map_used_by` was definitely `None`. So in `add_multi_attach_program()`, after the insert, even though we have the program, we get a fresh copy to do the `p.set_attached()` and `p.save(id)`:
```
        self.dispatchers.insert(did, dispatcher);
        let id = program
            .kernel_info()
            .expect("kernel info should be set after load")
            .id;
        self.programs.insert(id, program.to_owned());

        if let Some(p) = self.programs.get_mut(&id) {  <== Why do we do this then?
            p.set_attached();
            p.save(id)
                .map_err(|e| BpfdError::Error(format!("unable to save program state: {e}")))?;
        };
```

bpfd/src/bpf.rs
Billy99: Ok, done

bpfd/src/bpf.rs
```diff
@@ -991,17 +1007,51 @@ impl BpfManager {
...
+                    } else {
+                        warn!("unable to retrieve data for {used_by_id} when setting map_used_by on rebuild");
+                    }
+                }
+            }
         } else {
             let map = BpfMap { used_by: vec![id] };
             self.maps.insert(index, map);
+
+            // Update this programs with the updated map_used_by.
```
astoycos, line 1049: nit: I don't think you need this comment.
```suggestion
```

astoycos (changes requested):
Some small stuff!

bpfd/src/bpf.rs
```diff
@@ -311,18 +308,21 @@ impl BpfManager {
...
                 // and allow access to all maps by bpfd group members.
-                self.save_map(
-                    id,
-                    map_owner_id,
-                    map_pin_path.expect("map_pin_path must be set after successful load"),
-                )
-                .await?;
+                self.save_map(&mut program, id, map_owner_id).await?;
+
+                // Only add program to bpfManager if we've done everything
```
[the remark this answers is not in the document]
Billy99: Done

bpfd/src/bpf.rs
```diff
@@ -909,42 +903,63 @@ impl BpfManager {
...
 
-                set_dir_permissions(map_pin_path.to_str().unwrap(), MAPS_MODE).await;
+                // Set the permissions on the map_pin_path directory.
+                if let Some(map_pin_path) = data.map_pin_path() {
+                    if let Some(path) = map_pin_path.to_str() {
+                        set_dir_permissions(path, MAPS_MODE).await;
+                    } else {
+                        return Err(BpfdError::Error(
+                            "invalid map_pin_path {map_pin_path} for {id}".to_string(),
+                        ));
```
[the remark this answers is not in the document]
Billy99: Done

bpfd/src/bpf.rs
Billy99: Done

bpfd/src/bpf.rs
```diff
@@ -909,42 +903,63 @@ impl BpfManager {
...
+                    }
+                } else {
+                    return Err(BpfdError::Error(
+                        "map_pin_path should be set for {id}".to_string(),
```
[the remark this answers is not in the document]
Billy99: Done

astoycos (approved):
LGTM

Thanks @Billy99

Issue #468, opened 2023-05-02
Check for existence of `clsact` qdisc when call to create fails


anfredette:
The code currently ignores the error when creating the `clsact` qdisc assuming that if there's an error, it must already exist.  
https://github.com/bpfd-dev/bpfd/blob/main/bpfd/src/multiprog/tc.rs#L122

However, if another user has added the `ingress` qdisc, the call would also fail, but the subsequent attempt to add our dispatcher program to the `clsact` qdisc would fail.  

So, if we get an error on the create call, we need to handle it in some way.  Options include:

1. Use the existing ingress qdisc instead and only support ingress attachments on the given interface
2. Return an error
3. Delete the ingress qdisc and add clsact

I'm currently leaning toward 1. because it allows for better cooperation with other apps that are already running.

navarrothiago:
I've been looking into the issue. If another user has added the ingress qdisc, I guess we can assume that this user will manage the qdisc. I'm wondering if, following option 1, our program might be in an unknown state after the qdisc is deleted. This could mean that our BPF programs attached to the ingress qdisc would be removed unexpectedly. Then, I believe we can't rely on that. Does it make sense?

anfredette:
@navarrothiago thanks for looking at this.  It is true that if the other user/program deletes the qdisc, any programs we've attached will get deleted.  That's also true if we are using clsact and another user deletes it without us knowing.  In any case, it's probably safer for us to go with option 2 and return an error if the ingress qdisc already exists because that means that another user is already active on the system.  We may also want to check for the clsact qdisc before blindly adding it.  I'm not sure whether this is possible, but another thought is that it would be good if we could get notified if a clsact qdisc is deleted.  That would at least allow us to clean up our state and report an error.

Also, this illustrates some of the challenges we can have when multiple applications are managing their eBPF programs independently.  I.e., it's easy for them to step on each other.

navarrothiago:
@anfredette, could you assign the issue to me?

PR #759, opened 2023-10-04
bpf: adds cleanup for tc programs when qdisc is deleted

navarrothiago:
This PR introduces a new logic to monitor qdisc destroy events in the Linux Traffic Control subsystem. Upon detection of these events, the code performs a thorough cleanup operation for all TC programs attached to the affected network interface. This ensures that stale or orphaned TC programs do not persist in the bpfd.

Related: #468

[asciicast](https://asciinema.org/a/611968)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -251,6 +251,7 @@ dependencies = [
  "log",
  "object",
  "thiserror",
+ "tokio",
 ]
 
 [[package]]
@@ -377,6 +378,7 @@ dependencies = [
  "base16ct",
  "bpfd-api",
  "bpfd-csi",
+ "bytes",
  "caps",
  "chrono",
  "clap",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ base16ct = { version = "0.2.0", default-features = false }
 base64 = { version = "0.21.4", default-features = false }
 bpfd-api = { version = "0.2.0", path = "./bpfd-api" }
 bpfd-csi = { version = "1.8.0", path = "./csi" }
+bytes = { version = "1.5.0", default-features = false }
 caps = { version = "0.5.4", default-features = false }
 chrono = { version = "0.4.31", default-features = false }
 clap = { version = "4", default-features = false }
--- a/bpf/tc_qdisc.bpf.c
+++ b/bpf/tc_qdisc.bpf.c
@@ -1,0 +1,90 @@
+/* SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause) */
+/* Copyright Authors of bpfd */
+#include <linux/bpf.h>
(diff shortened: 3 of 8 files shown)


navarrothiago:
Still need to fix the jobs...

anfredette:
@navarrothiago thanks for submitting this.  I've taken a quick look at the code and have tried it out.  I think this is a really good function to have.  This could be a separate PR, but if we could also trigger the operator to do a reconcile for TC programs after your code does the cleanup, the bpfd programs would get reinstalled.  I'm also wondering if we could/should do something similar for XDP.  I.e., if someone replaces our XDP dispatcher on an interface, we could clean up the XDP state too.

I'd like to do a more thorough review of the code and would also like to discuss the approach in the community meeting tomorrow to get feedback from others.

I'm going to put a hold on it for now pending that feedback.

anfredette (changes requested):
As discussed, rather than install our own bpf tracepoint program, we should be able to get this information via netlink notifications.  Furthermore, it would be beneficial, in general, for bpfd to have a way to subscribe to arbitrary notifications that we may care about for other reasons.  For example, the case where the xdp_dispatcher is removed that I mentioned earlier.  Please take a look at the same netlink library that was suggested in pr #761.

Additionally, rather than cleaning everything up, I think we can rebuild and reattach a new dispatcher.  What happens when the clsact qdisc is deleted is that the tc_dispatcher becomes no longer attached, so the user bpf programs are no longer being executed.  However, all of the programs are still loaded in memory and bpfd state reflects what we'd like to be the case.  So, we should be able to do something very similar to what is being done in [rebuild_dispatcher_state()](https://github.com/bpfd-dev/bpfd/blob/main/bpfd/src/bpf.rs#L212).

It's a little different because rather than looping through the dispatcher directory entries like we're currently doing, for this case we'd look in the dispatcher directory to see if we have an ingress and/or egress dispatcher for the particular interface on which the qdisc was just deleted.  If we do, then we rebuild just those.  Also, we shouldn't need to call [self.dispatchers.insert()](https://github.com/bpfd-dev/bpfd/blob/main/bpfd/src/bpf.rs#L241C21-L241C45) because it's already on the list.

anfredette:
> look in the dispatcher directory to see if we have an ingress and/or egress dispatcher for the particular interface on which the qdisc was just deleted

Actually, rather than looking in the directory, we could just check if the dispatcher exists in our [dispatcher map](https://github.com/bpfd-dev/bpfd/blob/main/bpfd/src/bpf.rs#L50).

anfredette:
@navarrothiago, this has been inactive for a while now.  Unless I hear from you otherwise, I'm planning to close it.  Thanks, Andre

anfredette:
Closing pr because it has not been updated for a long time.

@navarrothiago we appreciate your contribution and welcome you to update this in the future.

PR #761, opened 2023-10-06
bpf: returns an error when adding a tc program to existence clsact qdisc

navarrothiago:
closes: #468

### DIFF
--- a/bpfd/src/multiprog/tc.rs
+++ b/bpfd/src/multiprog/tc.rs
@@ -13,7 +13,9 @@ use aya::{
     Bpf, BpfLoader,
 };
 use bpfd_api::util::directories::*;
+use futures::stream::TryStreamExt;
 use log::debug;
+use netlink_packet_route::tc::Nla;
 use serde::{Deserialize, Serialize};
 use tokio::sync::mpsc::Sender;
 
@@ -110,19 +112,61 @@ impl TcDispatcher {
             loader: Some(loader),
         };
         dispatcher.attach_extensions(&mut extensions).await?;
-        dispatcher.attach(old_dispatcher)?;
+        dispatcher.attach(old_dispatcher).await?;
         dispatcher.save()?;
         Ok(dispatcher)
     }
 
-    fn attach(&mut self, old_dispatcher: Option<Dispatcher>) -> Result<(), BpfdError> {
+    /// has_qdisc returns true if the qdisc_name is found on the if_index.
+    async fn has_qdisc(qdisc_name: String, if_index: i32) -> Result<bool, anyhow::Error> {
+        let (connection, handle, _) = rtnetlink::new_connection().unwrap();
+        tokio::spawn(connection);
+
+        let mut qdiscs = handle.qdisc().get().execute();
+        while let Some(qdisc_message) = qdiscs.try_next().await? {
+            if qdisc_message.header.index == if_index
+                && qdisc_message.nlas.contains(&Nla::Kind(qdisc_name.clone()))
+            {
+                return Ok(true);
+            }
+        }
+        Ok(false)
+    }
+
+    async fn attach(&mut self, old_dispatcher: Option<Dispatcher>) -> Result<(), BpfdError> {
         debug!(
             "TcDispatcher::attach() for if_index {}, revision {}",
             self.if_index, self.revision
         );
         let iface = self.if_name.clone();
-        // Add clsact qdisc to the interface. This is harmless if it has already been added.
-        let _ = tc::qdisc_add_clsact(&iface);
+
+        // Aya returns an error when trying to add a qdisc that already exists, which could be ingress or clsact. We
+        // need to make sure that the qdisc installed is the one that we want, i.e. clsact. If the qdisc is an ingress
+        // qdisc, we return an error. If the qdisc is a clsact qdisc, we do nothing. Otherwise, we add a clsact qdisc.
+
+        // no need to add a new clsact qdisc if one already exists.
+        if TcDispatcher::has_qdisc("clsact".to_string(), self.if_index as i32).await? {
+            debug!(
+                "clsact qdisc found for if_index {}, no need to add a new clsact qdisc",
+                self.if_index
+            );
+
+        // if ingress qdisc exists, return error.
+        } else if TcDispatcher::has_qdisc("ingress".to_string(), self.if_index as i32).await? {
+            debug!("ingress qdisc found for if_index {}", self.if_index);
+            return Err(BpfdError::InvalidAttach(format!(
+                "Ingress qdisc found for if_index {}",
+                self.if_index
+            )));
+
+        // otherwise, add a new clsact qdisc.
+        } else {
+            debug!(
+                "No qdisc found for if_index {}, adding clsact",
+                self.if_index
+            );
+            let _ = tc::qdisc_add_clsact(&iface);
+        }
 
         let new_dispatcher: &mut SchedClassifier = self
             .loader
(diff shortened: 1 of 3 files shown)


bpfd/src/multiprog/tc.rs
```diff
@@ -122,8 +122,14 @@ impl TcDispatcher {
             self.if_index, self.revision
         );
         let iface = self.if_name.clone();
-        // Add clsact qdisc to the interface. This is harmless if it has already been added.
-        let _ = tc::qdisc_add_clsact(&iface);
+
+        // let's return an error if the interface is not valid for now.
+        if let Err(e) = tc::qdisc_add_clsact(&iface) {
```
anfredette, line 127: Unfortunately, this is isn't going to work this easily because Aya returns the same error whether the `ingress` or `clsact` qdisc has already been added on the interface.  So, I think we're going to need to do something like:

```
qdiscs = Get existing qdiscs for interface (if any)
if qdiscs contains clsact
  we're good (no need to re-add the clsact qdisc)
else if qdiscs contains "ingress"
 return an error saying "ingress qdisc has been added on the interface"
else
  call qdisc_add_clsact() to add clsact qdisc.
```

Also unfortunately, I don't think Aya has an API to get the existing qdiscs, so, we'd need to add one.

anfredette (changes requested):

anfredette (commented):
@navarrothiago thanks for submitting this change!  Unfortunately, it's not going to work this easily.  Please see comments inline.

anfredette:
An alternative is the remember whether we've already created the clsact qdisc, and only try to create it if we haven't, but I'd rather not rely on the assumption that it's still there (i.e., that no one has deleted it).

anfredette:
As discussed, rather than adding a new netlink helper function to Aya to get the existing qdiscs, we'd rather use netlink directly.  

@dave-tucker suggests we try using the following library:
https://github.com/rust-netlink
Also from Dave:
> If you need to go low-level, you can find an example here: https://github.com/redhat-et/emerita/blob/main/emerita/src/netlink.rs
> It’s creating veth pairs, but you should be able to adapt that to our needs

Additionally, rather than cleaning everything up, I think we can pretty easily rebuild and reattach a new dispatcher.  What happens when the clsact qdisc is deleted is that the tc_dispatcher is no longer attached.  However, all of the programs are still loaded in memory and bpfd state reflects what we'd like to be the case.  So, we should be able to do something very similar to what is being done in [rebuild_dispatcher_state()](https://github.com/bpfd-dev/bpfd/blob/main/bpfd/src/bpf.rs#L212).

It's a little different because when we're currently rebuilding state, we loop through the entries in the dispatcher directory and rebuild them all.  In this case, we'd look in the dispatcher directory to see if we have an ingress and/or egress dispatcher for the particular interface on which the qdisc was just deleted.  If we do, then we rebuild just those.  Also, we shouldn't need to call [self.dispatchers.insert()](https://github.com/bpfd-dev/bpfd/blob/main/bpfd/src/bpf.rs#L241C21-L241C45) because it's already on the list.

bpfd/src/multiprog/tc.rs
```diff
@@ -110,19 +112,72 @@ impl TcDispatcher {
...
+                for nla in qdisc_message.nlas {
+                    if let Nla::Kind(nla_kind) = nla {
+                        if nla_kind == qdisc_name {
+                            return Ok(true);
+                        }
+                    }
+                }
+            }
+        }
+        Ok(false)
```
anfredette, line 144: I'm not necessarily saying to do this, but it seemed like we could use the try_filter() method to skip the messages we don't care about as an option to what you are doing on lines 126 & 127.  So, I played around with it and came up with the following.  What the following code does is skip everything except for what we are looking for.  So, after that's done, if anything is left, it's the qdisc we are looking for on the if_index provided.

I'd like to get @dave-tucker's opinion on two things.  First of all, the pattern @navarrothiago used seems to follow those in the rtnetlink examples (https://github.com/rust-netlink/rtnetlink/tree/main/examples).  However, the example of yours that you pointed us to seems a bit different and lower level.  Are you okay with this?  Also, do you have a preference for either Thiago's option or the one below?

```suggestion
        if (handle
            .qdisc()
            .get()
            .execute()
            .try_filter(|msg| {
                future::ready(
                    msg.header.index == if_index
                        && msg.nlas.contains(&Nla::Kind(qdisc_name.clone())),
                )
            })
            .try_next()
            .await?)
            .is_some()
        {
            return Ok(true);
        }        
        Ok(false)
```

bpfd/src/multiprog/tc.rs
```diff
@@ -110,19 +112,72 @@ impl TcDispatcher {
...
         Ok(dispatcher)
     }
 
-    fn attach(&mut self, old_dispatcher: Option<Dispatcher>) -> Result<(), BpfdError> {
+    /// has_qdisc returns true if the qdisc_name is found on the if_index.
+    async fn has_qdisc(qdisc_name: String, if_index: i32) -> Result<bool, anyhow::Error> {
+        let (connection, handle, _) = rtnetlink::new_connection().unwrap();
+        tokio::spawn(connection);
+
+        let mut qdiscs = handle.qdisc().get().execute();
```
anfredette, line 125: The QDiscHandle::get() method is equivalent to `tc qdisc show`.  So this works, but we get all the qdiscs on all the interfaces.  It would be nice if `rtnetlink` had a get option equivalent to `tc qdisc show dev <DEVICE>`.  It seems like it wouldn't be too hard to add, but this could be a future optimization.

bpfd/src/multiprog/tc.rs
```diff
@@ -110,19 +112,72 @@ impl TcDispatcher {
...
+        while let Some(qdisc_message) = qdiscs.try_next().await? {
+            if qdisc_message.header.index == if_index {
+                debug!(
+                    "Qdisc: {:?} - Parent: {:?} - Handle: {:?} - Kind: {:?}",
+                    qdisc_message.header.index,
+                    qdisc_message.header.parent,
+                    qdisc_message.header.handle,
+                    qdisc_message.header.info
+                );
+                for nla in qdisc_message.nlas {
```
anfredette, line 135: Even if we don't go with the option I gave below, it would be better to use the qdisc_message.nlas.contains() here instead of a for loop.

anfredette (changes requested):
@nav

anfredette (changes requested):
@navarrothiago thanks for reworking this!  It works and it looks pretty good.  I made a couple of minor comments and gave an alternative for the has_qdisc() function.  However, I'm not a Rust expert either, so I'd like to hear what @dave-tucker has to say before you change anything.

bpfd/src/multiprog/tc.rs
```diff
@@ -110,19 +112,72 @@ impl TcDispatcher {
...
+                    if let Nla::Kind(nla_kind) = nla {
+                        if nla_kind == qdisc_name {
+                            return Ok(true);
+                        }
+                    }
+                }
+            }
```
anfredette, line 142: I deleted my earlier suggestion.  I talked it over with Dave.  Either would work, but your way has the advantage that it will exit as soon as we find the qdisc in question, so it could exit earlier.  Im making the following new suggestion that just deletes your debug log (I think the logs in the calling code are sufficient) and replaces the for loop with a .contains().

```suggestion
            if qdisc_message.header.index == if_index
                && qdisc_message.nlas.contains(&Nla::Kind(qdisc_name.clone()))
            {
                return Ok(true);
            }
```

anfredette (changes requested):
I talked things over with Dave this morning.  I'm suggesting a small change then this should be good to go.

navarrothiago:
Thanks @anfredette for the review. Have just push another commit. Let me know if we need to change anything else.

anfredette (approved):
LGTM

PR #768, opened 2023-10-12
ci: add sha tags to all images

astoycos:
Cleanup out tag setup in ci, remove explicit priorities and add a sha tag.


astoycos:
@navarrothiago PTAL :) 

I also sent you an invite to out contributor's ORG!

navarrothiago (commented):
Thanks a lot to work on that! This is what we were looking for.

anfredette (approved):
LGTM

Issue #508, opened 2023-05-21
Release follow-up... Add Docs on how to run with release locally and in K8s 


astoycos:
Now that we have our first release https://github.com/bpfd-dev/bpfd/releases/tag/v0.2.0 we should have instructions on how to run bpfd locally and in K8s with the release payload materials. 

Locally it will just involve curling the bpfd and bpfctl binaries from the release and unpacking them to your bin for use

In k8s you can do something like 

1. Create a kind cluster 

`kind create cluster --name=test-bpfd` 

2. Install cert-manager

`kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.11.0/cert-manager.yaml`

3. Deploy bpfd Crds 

`kubectl apply -f  https://github.com/bpfd-dev/bpfd/releases/download/v0.2.0/bpfd-crds-install-v0.2.0.yaml`

4. Deploy bpfd-operator

`kubectl apply -f https://github.com/bpfd-dev/bpfd/releases/download/v0.2.0/bpfd-operator-install-v0.2.0.yaml`

5. Deploy an example 

`kubectl apply -f https://github.com/bpfd-dev/bpfd/releases/download/v0.2.0/go-xdp-counter-install-v0.2.0.yaml`

PR #771, opened 2023-10-13
docs: Add Docs on how to run with release locally and in K8s

Billy99:
Update the documentation to include a section on how to pull and use binary images from released versions.

Fixes: #508

### DIFF
--- a/docs/getting-started/running-release.md
+++ b/docs/getting-started/running-release.md
@@ -1,0 +1,143 @@
+# Run bpfd From Release Image
+
+This section describes how to deploy `bpfd` from a given release.
+See [Releases](https://github.com/bpfd-dev/bpfd/releases) for the set of bpfd
+releases.
+
+Jump to the [Setup and Building bpfd](./building-bpfd.md) section
+for help building from the latest code or building from a release branch.
+
+[Tutorial](./tutorial.md) contains more details on the different
+modes to run `bpfd` in on the host and how to test.
+Use [Privileged Mode](#privileged-mode) or [Systemd Service](#systemd-service)
+below for deploying released version of `bpfd` and then use [Tutorial](./tutorial.md)
+for further information on how to test and interact with `bpfd`. 
+
+[Deploying the bpfd-operator](../developer-guide/operator-quick-start.md) contains
+more details on deploying `bpfd` in a Kubernetes deployment and
+[Deploying Example eBPF Programs On Kubernetes](./example-bpf-k8s.md) contains
+more details on interacting with `bpfd` running in a Kubernetes deployment.
+Use [Deploying Release Version of the bpfd-operator](#deploying-release-version-of-the-bpfd-operator)
+below for deploying released version of `bpfd` in Kubernetes and then use the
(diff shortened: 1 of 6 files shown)


Billy99:
@astoycos PTAL

docs/getting-started/running-release.md
```diff
@@ -0,0 +1,143 @@
...
+
+```console
+kind create cluster --name=test-bpfd
+```
+
+Next, deploy the bpfd CRDs:
+
+```console
+export BPFD_REL=0.3.0
+kubectl apply -f  https://github.com/bpfd-dev/bpfd/releases/download/v${BPFD_REL}/bpfd-crds-install-v0.2.0.yaml
```
anfredette, line 124: Did you intend to use ${BPFD_REL} instead of "0.2.0" here and in the next two apply commands?

Billy99: Good catch. Fixed.

docs/getting-started/running-release.md
```diff
@@ -0,0 +1,143 @@
...
+ Program ID  Name       Type  Load Time                
+```
+
+Continue in [Tutorial](./tutorial.md) if desired.
+
+## Systemd Service
+
+To run `bpfd` as a systemd service, the binaries will be placed in a well known location
+(`/usr/sbin/.`) and a service configuration file will be added
+(`/usr/lib/systemd/system/bpfd.service`).
```
astoycos, line 72: Do we need to add this systemd service file to our release payload?

astoycos: Hrm maybe not your instructions following this seem to cover it by pulling from source code...

But if we want to make it easier to run via systemd we can definitely try to bundle the payload a bit differently.

docs/getting-started/running-release.md
```diff
@@ -0,0 +1,143 @@
...
+## Deploying Release Version of the bpfd-operator
+
+The quickest solution for running `bpfd` in a Kubernetes deployment is to run a
+Kubernetes KIND Cluster:
+
+```console
+kind create cluster --name=test-bpfd
+```
+
+Next, deploy the bpfd CRDs:
```
astoycos, line 120:
```suggestion
First, deploy the bpfd CRDs:
```

astoycos: Not worth holding this up :)

astoycos (approved):
LGTM

docs/getting-started/running-release.md
Billy99: I assumed the `kind create cluster --name=test-bpfd` was the "First" step so the "Next" was deploying CRDs.

docs/getting-started/running-release.md
Billy99: Just to document, we would need (this is without testing, just visual inspection):
./scripts/setup.sh
./scripts/certificates.sh
./scripts/install.sh
./scripts/user.sh
./scripts/bpfd.service

The script also copies the TC, Tracepoint and XDP example bytecode to /run/bpfd/examples/. This would fail and print error, but would complete the install.

Issue #693, opened 2023-09-07
Investigate a Daemonless Architecture


dave-tucker:
We should consider maybe not being a daemon.
I have some notes about what this would look like - I'll add them here shortly,

PR #777, opened 2023-10-19
bpfd: Unify the "run as root" and "run as bpfd user" codepaths

Billy99:
* Remove the bpfd user and user group
  * Remove the User/Group from bpfd.service
  * Remove the directory management from bpfd.service
  * Add directory mode setting to bpfd main.rs with directory creation.
  * Update docs now that bpfctl requires sudo for systemd deployment.
  * Simplify install scripts since user is no longer created.
* Document the current Linux capabilities and why they are needed.
* Add helptext to "cargo xtask".
* Add "cargo xtask copy" to copy binaries so `sudo bpfctl` can be used instead of `sudo ./target/debug/bpfctl`.
* Removed CAP_DAC_OVERRIDE and CAP_PERFMON from list of capabilities required by bpfd. 

Resolves: #742

PR #778, opened 2023-10-20
docs/design: Add daemonless design doc

dave-tucker:
Closes: #693

### DIFF
--- a/docs/design/daemonless.md
+++ b/docs/design/daemonless.md
@@ -1,0 +1,150 @@
+# Daemonless bpfd
+
+## Introduction
+
+The bpfd daemon is a userspace daemon that runs on the host and responds to
+gRPC API requests over a unix socket, to load, unload and list the eBPF
+programs on a host.
+
+The rationale behind running as a daemon was because something needs to be
+listening on the unix socket for API requests, and that we also maintain
+some state in-memory about the programs that have been loaded.
+However, since this daemon requires root privileges to load and unload eBPF
+programs it is a security risk for this to be a long-running - even with the
+mitigations we have in place to drop privileges and run as a non-root user.
+This risk is equivalent to that of something like [Docker].
+
+[Docker]: https://docs.docker.com/engine/security/#docker-daemon-attack-surface
+
+This document describes the design of a daemonless bpfd, which is a
+bpfd that runs only runs when required, for example, to load
+or unload an eBPF program.
+
+## Design
+
+The daemonless bpfd is a single binary that collects some of the functionality
+from both bpfd and bpfctl.
+
+> :note: Daemonless, not rootless. Since CAP_BPF (and more) is required to
+> load and unload eBPF programs, we will still need to run as root. But at
+> least we can run as root for a shorter period of time.
+
+### Command: bpfd system service
+
+This command will run the bpfd gRPC API server - for one or more of the
+gRPC API services we support.
+
+It will listen on a unix socket (or tcp socket) for API requests - provided via
+a positional argument, defaulting to `unix:///var/run/bpfd.sock`.
+It will shutdown after a timeout of inactivity - provided by a `--timeout` flag
+defaulting to 5 seconds.
+
+It will support being run as a systemd service, via socket activation, which
+will allow it to be started on demand when a request is made to the unix socket.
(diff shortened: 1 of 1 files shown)


dave-tucker:
@Billy99 having worked through the details of the design I think this has some relation to #777.
Happy to chat later today (or next week), but TL;DR the directory creation logic will stay in bpfd now and be removed from the unit files - which is the reverse of what we were talking about in the scripts.

docs/design/daemonless.md
```diff
@@ -0,0 +1,139 @@
...
+Currently bpfd maintains state in-memory about the programs that have been
+loaded (by bpfd, and the kernel).
+Some of this state is flushed to disk, so if bpfd is restarted it can
+reconstruct it.
+
+Flushing to disk and state reconstruction is cumbersome at present and having
+to move all state management out of in-memory stores is a forcing function
+to improve this.
+We will replace the existing state management with [sled], which gives us
+a familiar API to work with while also being fast, reliable and persistent.
```
astoycos, line 98: 🥳

docs/design/daemonless.md
```diff
@@ -0,0 +1,139 @@
...
+### Metrics and Monitoring
+
+While adding metrics and monitoring is not a goal of this design, it should
+nevertheless be a consideration. In order to provide metrics to Prometheus
+or OpenTelemetry we will require an additional exporter process.
+
+We can either:
+
+1. Use the bpfd socket and retrieve metrics via the gRPC API
+1. Place state access + metrics gathering functions in a library, such that
```
astoycos, line 111:
```suggestion
2. Place state access + metrics gathering functions in a library, such that
```

docs/design/daemonless.md
```diff
@@ -0,0 +1,139 @@
...
+## Benefits
+
+The benefits of this design are:
+
+- No long-running daemon with root privileges
+- No need to run as a non-root user, this is important since the number of
+  capabilities required is only getting larger.
+- We only need to ship a single binary
+- We can use systemd socket activation to start the daemon on demand
+- Better state management
```
astoycos, line 130: We could also introduce this new state management method with the daemon....right?

docs/design/daemonless.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Drawbacks
```
astoycos, line 132: If we ever want to pivot to being a "runtime enforcement tool" Moving away from a daemon seems like it would be an issue 🤔

astoycos (commented):
Thanks for opening sorry for the delay 

Just a few questions!

docs/design/daemonless.md
dave-tucker: Markdown numbered lists allow you to re-use `1.` so that the numbering is sequential.
That way when you add another numbered bullet the lists re-adjusts.
I prefer this as I suck at remembering to re-number lists 😆

docs/design/daemonless.md
dave-tucker: You could, but the onus is on us to keep state out of memory.
In the daemonless case we literally have no choice so it can't creep back in.

docs/design/daemonless.md
dave-tucker: Nope. One of the benefits here is that bpfd becomes a lot more modular.
We're relying on systemd - or something else - to orchestrate the necessary pieces.
You want to do runtime enforcement? Fine. That can be another binary. Whether or not that piece runs as a daemon isn't our problem ;)

docs/design/daemonless.md
```diff
@@ -0,0 +1,139 @@
...
+programs it is a security risk for this to be a long-running - even with the
+mitigations we have in place to drop privileges and run as a non-root user.
+
+This document describes the design of a daemonless bpfd, which is a
+bpfd that runs as a non-root user, and only runs when it is needed to load
+or unload an eBPF program.
+
+## Design
+
+The daemonless bpfd is a single binary that collects some of the functionality
```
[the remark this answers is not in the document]
dave-tucker: Agreed, but we can address that later

docs/design/daemonless.md
astoycos: Noice I actually didn't know that!

Issue #740, opened 2023-09-28
Investigate Options For Sitting Outside of The Container Load Path


dave-tucker:
There are at least 2 things we can try here:

1. Using the recent [BPF_TOKEN](https://lore.kernel.org/bpf/CAEiveUeDLr00SjyU=SMSc4XbHSA6LTn4U2DHr12760rbo5WqSw@mail.gmail.com/T/) stuff upstream. In this case we'd help prepare the token so it gets placed into the user namespace.
1. Using secomp_notify - so we'd perform the syscalls outside of the user namespace.

Note: This may mean we want bpfd to be more closely attached to container runtimes.

Somewhat related to: #693

astoycos:
Is this relevant anymore? I figure this maps to https://github.com/bpfman/bpfman/pull/779

PR #779, opened 2023-10-20
docs/design: Options for being outside the loadpath

dave-tucker:
Closes: #740

### DIFF
--- a/docs/design/loadpath.md
+++ b/docs/design/loadpath.md
@@ -1,0 +1,284 @@
+# bpfman outside of the load-path
+
+## Introduction
+
+One of the drawbacks of bpfman in it's current incarnation is that it requires
+users to "opt in" to our opinionated loading pattern for eBPF programs.
+
+From a "linux only" standpoint the benefits of our scheme are as follows:
+
+1. Offering containers as a lightweight packaging alternative to RPMs/DEBs
+   for eBPF programs.
+1. Provides signing of eBPF Container Images through cosign.
+1. Provides a way to load eBPF programs from a container image without
+   requiring the container runtime to be installed on the host.
+1. Allows for long-running user-space processes (that work with eBPF maps) to
+   be run with no privileges, since most map read interactions are permissable
+   without CAP_BPF. This is a security benefit, since it reduces the attack
+   surface of the user-space process. There are a few exceptions to this
+   rule (e.g perf_event_array maps), but we're planning on finding a way to
+   work around this.
+1. Simplified (and consistent) lifecycle management of eBPF programs regardless
+   of the program type or attachment point.
+
+Given that a lot of the software using eBPF is also Cloud-Native, we have the
+following additional benefits:
+
+1. A Cloud-Native, declarative way to load eBPF programs and have them
+   orchestrated by Kubernetes. This is backed by packaging eBPF programs as
+   container images.
+1. For daemonsets/deployments that use these eBPF programs to not require
+   privileged containers, host mounts or any capabilities.
+
+These benefits however are proving to be insufficient to sway prospective users
+away from loading their own eBPF programs for 2 reasons:
+
+1. They are an established project and has code that works and don't see a
+   reason to change it
+1. They believe that separating the user-space code and kernel-space code is
+   a bad idea, citing the following reasons:
+   - The versions of the kernel and user-space must be tightly coupled.
+   - They are wedded to BPF Skeletons (bpftool generated C code) since they
+     require complex .bss/.rodata section manipulation before loading, and they
+     argue that this is too complex to model via an API.
+
+This design document explores a way to address these concerns, while still
+providing the benefits of the current bpfman design.
+
+## Design
+
+### Requirements
+
+We would like for bpfman to still be valuable to users who don't want to use
+our opinionated loading pattern. To address the concerns raised above, we
+must allow users to package and load their eBPF bytecode as they see fit.
+However we are left with the following gaps:
+
+1. Signing of eBPF programs - previously enabled via cosign.
+1. Removing CAP_BPF from long running processes or daemonsets/deployments
+   that use eBPF programs.
+1. Simplified lifecycle management of eBPF programs through pinning.
+
+We can immediately discard "simplified lifecycle management" as a requirement
+since that becomes the responsibility of the user. However, we will evaluate
+potential solutions to the other 2 requirements.
+
+### Signing
+
+Therefore, in order for bpfman to NOT be in the load path, we need to find a way
+to preserve the security benefits of our current design.
+
+#### fsverity and LSM Gatekeeper
+
+[BPF Signing using fsverity and LSM Gatekeepers] is a proposal that was put
+forward by Lorenz Bauer. It proposes a way to sign programs that contain
+embedded eBPF bytecode (or eBPF bytecode itself), the kernel will then verify
+the signature when the program is loaded. LSM hooks can the be used to further
+apply policy to the program that's being loaded.
+
+This prevents a bad actor from modifying the eBPF bytecode in a program, since
+the signature will no longer be valid.
+
+This proposal is still a work in progress, but it's a promising development.
+In particular, the integration with the IMA subsystem is very interesting as it
+opens up the possibility of remote attestations using Keylime. Not only that,
+the more I learn about IMAs capabilities, the more I think it's a good fit for
+eBPF.
+
+There are however some drawbacks to this approach:
+
+- Fsverity is limited to ext4, btrfs and f2fs filesystems.
+- Support for fsverity on overlayfs is enabled via mount option - additional work
+  may be required to support this in container runtimes.
+- This approach is less effective on static binaries, since if they were to,
+  for example, link to libbpf, then you would also need libbpf to be signed
+  with fsverity to avoid any potential tampering. This could be mitigated by
+  having relocations performed by the kernel instead.
+- Some investigation is needed to see whether the fsverity xattrs are preserved
+  when a file is copied into a container image, or signed there as part of the
+  build process.
+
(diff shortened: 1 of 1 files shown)


docs/design/loadpath.md
```diff
@@ -0,0 +1,164 @@
...
+
+1. bpfd might be able to provide information to Seitan that can be used to
+   generate the seccomp filters that are required to proxy syscalls.
+1. bpfd might do nothing, except for preparing a bpffs filesystem for the
+   container runtime to use.
+
+[The Seccomp Notifier: New Frontiers in Unprivileged Container Development]: https://people.kernel.org/brauner/the-seccomp-notifier-new-frontiers-in-unprivileged-container-development
+[Seitan](https://seitan.rocks/seitan/about/)
+
+### Authoritative Policy (via LSM or otherwise)
```
astoycos, line 126: We should probably add a note on the approach from 

 http://vger.kernel.org/bpfconf2023_material/Lorenz_Bauer_-_BPF_signing_using_fsverity_and_LSM_gatekeeper.pdf 

and 

https://github.com/isovalent/bpf-verity

astoycos (commented):
Thanks for opening this @dave-tucker! 

I think It lays out the permission model use cases very succinctly, I do wonder should we be mentioning our bpf subsystem observability value adds  as well or is that for a different doc?

docs/design/loadpath.md
```diff
@@ -0,0 +1,164 @@
...
+users to "opt in" to our opinionated loading pattern for eBPF programs.
+
+From a "linux only" standpoint the benefits of our scheme are as follows:
+
+1. Offering containers as a lightweight packaging alternative to RPMs/DEBs
+   for eBPF programs.
+1. Provides signing of eBPF Container Images through cosign.
+1. Provides a way to load eBPF programs from a container image without
+   requiring the container runtime to be installed on the host.
+1. Allows for long-running user-space processes (that work with eBPF maps) to
```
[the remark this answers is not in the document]
dave-tucker: We're just missing a command in the bpfd CLI to populate a bpffs with maps etc...
But the design will absolutely work for this use-case.

docs/design/loadpath.md
```diff
@@ -0,0 +1,164 @@
...
+## Introduction
+
+One of the drawbacks of bpfd in it's current incarnation is that it requires
+users to "opt in" to our opinionated loading pattern for eBPF programs.
+
+From a "linux only" standpoint the benefits of our scheme are as follows:
+
+1. Offering containers as a lightweight packaging alternative to RPMs/DEBs
+   for eBPF programs.
+1. Provides signing of eBPF Container Images through cosign.
```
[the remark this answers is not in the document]
dave-tucker: True, but who's verifying the signature? It's all well and good signing something, but someone somewhere needs to actuall verify that signature when the process is started - we're targetting Linux, not macOS/iOS/Android/Windows that already do signature checking by default.

docs/design/loadpath.md
```diff
@@ -0,0 +1,164 @@
...
+   requiring the container runtime to be installed on the host.
+1. Allows for long-running user-space processes (that work with eBPF maps) to
+   be run with no privileges, since most map read interactions are permissable
+   without CAP_BPF. This is a security benefit, since it reduces the attack
+   surface of the user-space process. There are a few exceptions to this
+   rule (e.g perf_event_array maps), but we're planning on finding a way to
+   work around this.
+1. Simplified (and consistent) lifecycle management of eBPF programs regardless
+   of the program type or attachment point.
+
```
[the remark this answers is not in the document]
dave-tucker: We can always add one later, right now we don't really have any policy to speak of - oustide of kube RBAC.

docs/design/loadpath.md
```diff
@@ -0,0 +1,164 @@
...
+1. They are an established project and has code that works and don't see a
+   reason to change it
+1. They believe that separating the user-space code and kernel-space code is
+   a bad idea, citing the following reasons:
+   - The versions of the kernel and user-space must be tightly coupled.
+   - They are wedded to BPF Skeletons (bpftool generated C code) since they
+     require complex .bss/.rodata section manipulation before loading, and they
+     argue that this is too complex to model via an API. 
+
```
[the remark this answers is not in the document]
dave-tucker: I'm sorry but I don't buy "it's too complex" as a reason.
There are absolutely options at their disposal, for example, the complex .bss/.rodata manipulations as described above, or pre-populating maps with config at program load time so you don't need to recompile as often. Even if you needed to recompile often, you might want to consider having a trusted bytecode factory that spits out signed bytecode images into a registry etc...

docs/design/loadpath.md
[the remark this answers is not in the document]
astoycos: good point!

docs/design/loadpath.md
```diff
@@ -0,0 +1,164 @@
...
+This design document explores a way to address these concerns, while still
+providing the benefits of the current bpfd design.
+
+## Design
+
+### Requirements
+
+In summary, the main things we want to accomplish with bpfd are as follows:
```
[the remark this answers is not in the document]
astoycos: Ok I should we specify somewhere that this design doc focuses on security? It just wan't apparent to me

docs/design/loadpath.md
```diff
@@ -0,0 +1,284 @@
...
+away from loading their own eBPF programs for 2 reasons:
+
+1. They are an established project and has code that works and don't see a
+   reason to change it
+1. They believe that separating the user-space code and kernel-space code is
+   a bad idea, citing the following reasons:
+   - The versions of the kernel and user-space must be tightly coupled.
+   - They are wedded to BPF Skeletons (bpftool generated C code) since they
+     require complex .bss/.rodata section manipulation before loading, and they
+     argue that this is too complex to model via an API.
```
tohojo, line 43: There_ are applications of dynamically generating BPF bytecode based on user input: https://github.com/xdp-project/xdp-tools/pull/356 - so I believe "dynamically generated code" should go on this list as well (BCC comes to mind as another user of this)...

docs/design/loadpath.md
```diff
@@ -0,0 +1,284 @@
...
+[BPF Signatures]: https://lpc.events/event/16/contributions/1357/attachments/1045/1999/BPF%20Signatures.pdf
+
+#### Signing Conclusion
+
+Both of these approaches have merits and some drawbacks.
+In an ideal world, we would have both of these approaches available to us in the
+following way:
+
+- All CO:RE relocations occur in the kernel - making libbpf and other libraries
+  that use CO:RE much simpler.
```
tohojo, line 136: ...by moving that complexity into the kernel so it runs with ring 0 privileges? Not sure that's an improvement...

docs/design/loadpath.md
```diff
@@ -0,0 +1,284 @@
...
+apply policy to the program that's being loaded.
+
+This prevents a bad actor from modifying the eBPF bytecode in a program, since
+the signature will no longer be valid.
+
+This proposal is still a work in progress, but it's a promising development.
+In particular, the integration with the IMA subsystem is very interesting as it
+opens up the possibility of remote attestations using Keylime. Not only that,
+the more I learn about IMAs capabilities, the more I think it's a good fit for
+eBPF.
```
tohojo, line 86: But AFAICT, the proposal is not to sign the actual BPF bytecode, it's just to verify that the binary trying to load a BPF program is trusted to do so? Which is basically the same that BPF token is doing, the fsverity thing is just a way of establishing who gets that trust. There is no restriction on *which* bytecode is being loaded into the kernel, so the application still gets the same blank check as if does with bpf token...

docs/design/loadpath.md
```diff
@@ -0,0 +1,284 @@
...
+   and how much room we have to define policy there.
+1. See what value, if any, bpfman could add on top of this.
+
+Longer term we might consider working with the Tetragon team on policy
+profiling and enforcement for eBPF programs.
+
+## In Summary
+
+In summary, we have a path forward to remove bpfman from the load path,
+while still providing some of the benefits of our current design.
```
tohojo, line 279: From reading the above it seems that the proposal is that bpfman can operate in one of two modes:
- The current "load on behalf of the container" mode
- A mode where bpfman becomes the entity that does the user namespace hole punching required to grant an entity CAP_BPF inside a container

As noted above, I'm not really sure the fsverity stuff adds anything that just signing the container image (combined with granting it CAP_BPF directly or by token) doesn't already provide. So, like, if the proposal is "we'll try to make people happy by making bpfman the system level daemon responsible for the privilege hole punching", that's fine, but a lot of the details could be omitted in that case? :)

docs/design/loadpath.md
dave-tucker: Ah yes I'd assumed that the signed binary contained bytecode embedded in an ELF section - which is the most common way for doing it. It could load the bytecode from disk, but unless you also protect that file with fs-verity you could be loading anything. I'll add some more wording here to clarify.

docs/design/loadpath.md
dave-tucker: Yes agreed, this did get pretty wordy in the end.
And your summary is pretty accurate.

I felt that we still needed to mention signing as that's one of the features of bpfman "load-on-behalf-of-container" mode. While we can hole punch, we'd rather not lose a security feature, so having another signing solution would be great.

I'll re-organize this and move a lot of the wordier bits into appendicies.

Thanks 🙏

docs/design/loadpath.md
tohojo: Well, in either case, if you can modify the binary you can also just modify it to completely rewrite the bytecode before loading it into the kernel, so whether the BPF bytecode itself is covered by the signature doesn't really matter as it's only the integrity of the userspace binary that is being checked, not the bytecode itself...

shaneutt:
Checking in: what's the current state / summary here?

dave-tucker:
Closing as stale

PR #783, opened 2023-10-25
workspace-ified the netlink dependencies

anfredette:
Also added `features = ["tokio_socket"]` to `rtnetlink` in the bpfd `Cargo.toml` which became necessary after the title change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,12 +29,14 @@ clap = { version = "4", default-features = false }
 comfy-table = { version = "7.1.0", default-features = false }
 env_logger = { version = "0.10", default-features = false }
 flate2 = { version = "1.0", default-features = false }
+futures = { version = "0.3.28", default-features = false }
 hex = { version = "0.4.3", default-features = false }
 integration-test-macros = { path = "./tests/integration-test-macros" }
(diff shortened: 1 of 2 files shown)


anfredette:
I'm not sure if this is the right solution.  We didn't see this build error before.  None of our related code changed, but it started failing at some point.

Cargo.toml
```diff
@@ -12,6 +12,7 @@ resolver = "2"
 
 [patch.crates-io]
 aya = { git = "https://github.com/aya-rs/aya", branch = "main" }
+rtnetlink = { git = "https://github.com/rust-netlink/rtnetlink", branch = "main" }
```
astoycos, line 15: Is there an upstream issue number we can link here to make sure we know when to revert back to a release?

anfredette: It was a local cargo cache issues, so I deleted this change.

anfredette:
It turned out that my original issue was a local one.  After clearing my cargo cache, the builds worked.  However, we still want to use workspaces for all the dependencies, so I left that change in this pr.

Billy99 (commented):
LGTM

Billy99 (approved):

PR #784, opened 2023-10-25
Support attaching uprobes to targets inside containers

anfredette:
This pr adds support for attaching uprobes to targets inside containers.

`namespace` has been renamed to `namespace_pid` in bpfd proto and changed to an int32. When the user provides a `namespace_pid`, bpfd will attempt to attach the uprobe inside the container with the given PID.

We use a new executable called `bpfd-ns` which can be called to attach uprobes and uretprobes inside containers.  `bpfd-ns` is spawned from bpfd when a `namespace_pid` parameter is provided when loading a uprobe. `bpfd-ns` changes to the mount namespace of the container, does the attach, and then changes back to the original mount namespace.


anfredette:
This is working and ready for review -- I'd be glad to get some feedback (even if it's general).

I marked it as a draft because I want to add an integration test before merging it.

tests/integration-test/src/tests/utils.rs
```diff
@@ -939,3 +956,87 @@ pub fn bpfman_output_map_ids(stdout: &str) -> Vec<String> {
 
     map_ids
 }
+
+#[derive(Debug)]
+pub struct DockerContainer {
```
astoycos, line 961: Do we want to be adding a dependency on docker in our core integration tests? Could we just do a 

```
[astoycos@nfvsdn-02-oot bpfman]$ unshare -Urfp --mount-proc

[astoycos@nfvsdn-02-oot bpfman]$ lsns | grep bash
4026533388 pid         1 3416159 astoycos └─-bash

[astoycos@nfvsdn-02-oot bpfman]$ sudo nsenter -t 3416159 -a ls
```

To basically mimic the same thing without docker?

Cargo.lock
```diff
@@ -420,6 +420,19 @@ dependencies = [
  "tonic 0.10.2",
 ]
 
+[[package]]
+name = "bpfman-ns"
+version = "0.3.0"
```
astoycos, line 425:
```suggestion
version = "0.3.1"
```

astoycos (changes requested):
Looking good a bunch of questions really! Also I just am looking to make sure this is generic enough to also use for exec-ing into network namespaces as well for TC and XDP. 

Additionally with the move to damonless, do we for-see all of this functionality moving back into core rather than it's own binary?

anfredette:
> Looking good a bunch of questions really! Also I just am looking to make sure this is generic enough to also use for exec-ing into network namespaces as well for TC and XDP. 

Additionally with the move to damonless, do we for-see all of this functionality moving back into core rather than it's own binary?   

I belive it can be extended for other types as well, but I wasn't planning to do that unless there's a specific demand due to our change in priorities.

We should be able to move this functionality into the core code if it becomes single threaded.

bpfman-ns/src/main.rs
```diff
@@ -0,0 +1,158 @@
...
+#[clap(author, version, about, long_about = None)]
+struct Cli {
+    #[clap(subcommand)]
+    command: Commands,
+}
+
+#[derive(Debug, Subcommand)]
+enum Commands {
+    /// Attach a uprobe program in the given namespace.
+    Uprobe(UprobeArgs),
```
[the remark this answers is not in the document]
anfredette: pid is optional.  It means "only execute uprobe if the function is called by the given pid".  namespace_pid is optional for the user api, but is required for this executable.  I think that when we extend this to other program types, program_pin_path and namespace_pid (or whatever we call it per comment below) would be common required parameters.

Regarding your second question, "namespace_pid" may not be a good name for the parameter.  This feature was originally billed as attaching uprobes in namespaces, so we started with "namespace" and I stuck with the theme. However, this is really the host PID for the container in which we want to attach the eBPF program.  Namespace isn't required except possibly as part of a selector in k8s.  My inclination is to change this parameter name to "container_pid" unless someone has better idea.

bpfman-ns/src/main.rs
```diff
@@ -0,0 +1,158 @@
...
+struct UprobeArgs {
+    /// Required: path to pinned entry for bpf program on a bpffs.
+    #[clap(short, long, verbatim_doc_comment)]
+    program_pin_path: String,
+
+    /// Optional: Function to attach the uprobe to.
+    #[clap(short, long)]
+    fn_name: Option<String>,
+
+    /// Required: Offset added to the address of the target function (or
```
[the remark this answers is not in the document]
anfredette: Good question.  It's optional for the user api, but I made it required for the api of this executable.  I set it to zero if the user didn't give an offset.  I'll update the comment to make this more clear.

bpfman-ns/src/main.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    // Set namespace to target namespace
+    let setns_result = setns(target_mnt_file, CloneFlags::CLONE_NEWNS);
+    match setns_result {
+        Ok(_) => debug!(
+            "set ns to target PID {} mnt namespace succeeded",
+            args.namespace_pid
+        ),
+        Err(e) => {
+            debug!(
```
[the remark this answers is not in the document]
anfredette: It was useful for debugging.  And, if this fails in real life, I think we'll want to know why.  I'll see what it looks like if I do a wrapper function like you suggest below.

bpfman-ns/src/main.rs
```diff
@@ -0,0 +1,158 @@
...
+                "failed to set ns to target PID {} mnt namespace",
+                args.namespace_pid
+            );
+            bail!(
+                "failed to set ns to target PID {} mnt namespace. Error: {}",
+                args.namespace_pid,
+                e
+            );
+        }
+    }
```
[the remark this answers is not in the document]
anfredette: I moved the setns out and cleaned up some debug logs.

bpfman/src/bpf.rs
```diff
@@ -559,28 +559,79 @@ impl BpfManager {
...
+                        prog_args.push("--retprobe");
+                    }
+
+                    let mut pid = String::new();
+                    if program.pid.is_some() {
+                        pid = program.pid.unwrap().to_string()
+                    }
+                    if !pid.is_empty() {
+                        prog_args.extend(["--pid", pid.as_str()])
+                    }
```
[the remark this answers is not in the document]
anfredette: I changed it a little to use `if let`, but if I try to do everything inside the `if let`, rust complains about the lifetime of `pid`.  So, I addressed that with the initial `let mut pid`.  If I try to do the `prog_args.extend` inside the `if let` rust complains that pid is assigned to but never read.  That doesn't make sense to me, but ¯\_(ツ)_/¯.  I don't know if there is a better way.

bpfman/src/bpf.rs
```diff
@@ -559,28 +559,79 @@ impl BpfManager {
...
-                    .pin(format!("{RTDIR_FS}/prog_{}_link", id))
-                    .map_err(BpfmanError::UnableToPinLink)?;
+                let program_pin_path = format!("{RTDIR_FS}/prog_{id}");
 
                 uprobe
-                    .pin(format!("{RTDIR_FS}/prog_{id}"))
+                    .pin(program_pin_path.clone())
                     .map_err(BpfmanError::UnableToPinProgram)?;
 
+                if program.namespace_pid.is_none() {
```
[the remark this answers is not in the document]
anfredette: That's better.

tests/integration-test/src/tests/utils.rs
anfredette: To be honest, I don't know.  It would be simpler and probably less overhead to not use it.  I asked in one of the team meetings, and people seemed to be okay with using Docker.  I saw https://www.redhat.com/sysadmin/pid-namespace where they discussed using the command you suggest to "build containers by hand".  Is this exactly equivalent to a Docker container and/or the ones used in k8s, or is it safer to just using Docker?  Personally, I'd have to do some more research to be sure.

anfredette:
@astoycos, Thanks for the review.  I think I've addressed most of your comments, but I didn't do them all exactly as suggested, so please take a look.

However, there are two things that I'd like to get consensus on before moving forward:

1. I think the parameter `namespace_pid` is confusing and should be changed.  I'm thinking `container_pid`.  We originally talked about attaching these programs in "namespaces", but it's not really a "namespace" in the k8s sense but rather a container (which does use linux namespaces++ and may be in a different k8s namespace from the user, but may not).  This change will ripple through the code, so I'd like to get some consensus before I do it.

1. Should we use Docker to create containers for integration test, or is it better to create our own "containers" using the Linux command line?

bpfman/src/bpf.rs
```diff
@@ -559,28 +559,82 @@ impl BpfManager {
...
+                            "uprobe",
+                            "--program-pin-path",
+                            program_pin_path.as_str(),
+                            "--offset",
+                            offset.as_str(),
+                            "--target",
+                            program.target.as_str(),
+                            "--namespace-pid",
+                            namespace_pid.as_str(),
+                        ];
```
dave-tucker, line 603:
```suggestion
                        let mut prog_args = vec![
                            "uprobe".to_string(),
                            "--program-pin-path".to_string(),
                            program_pin_path,
                            "--offset".to_string(),
                            offset,
                            "--target".to_string(),
                            program.target,
                            "--namespace-pid".to_string(),
                            namespace_pid,
                        ];
```

bpfman/src/bpf.rs
```diff
@@ -559,28 +559,82 @@ impl BpfManager {
...
+
+                        let fn_name = program.fn_name.clone().unwrap_or_default();
+                        if !fn_name.is_empty() {
+                            prog_args.extend(["--fn-name", fn_name.as_str()])
+                        }
```
dave-tucker, line 608:
```suggestion
                        if let Some(fn_name) = program.fn_name {
                            prog_args.extend(["--fn-name".to_string(), fn_name])
                        }
```

bpfman/src/bpf.rs
```diff
@@ -559,28 +559,82 @@ impl BpfManager {
...
+
+                        if program.retprobe {
+                            prog_args.push("--retprobe");
+                        }
```
dave-tucker, line 612:
```suggestion
                        if program.retprobe {
                            prog_args.push("--retprobe".to_string());
                        }
```

bpfman/src/bpf.rs
```diff
@@ -559,28 +559,82 @@ impl BpfManager {
...
+
+                        let mut pid = String::new();
+                        if let Some(p) = program.pid {
+                            pid = p.to_string();
+                        }
+                        if !pid.is_empty() {
+                            prog_args.extend(["--pid", &pid])
+                        }
```
dave-tucker, line 620:
```suggestion
                        if let Some(pid) = program.pid {
                            prog_args.extend(["--pid".to_string(), pid.to_string()])
                        }
```

dave-tucker (changes requested):

dave-tucker (commented):
Some nits around the way you build the args to the bpfman-ns binary but otherwise LGTM once Andrew's comments are addressed.

dave-tucker:
> @astoycos, Thanks for the review. I think I've addressed most of your comments, but I didn't do them all exactly as suggested, so please take a look.
> 
> However, there are two things that I'd like to get consensus on before moving forward:
> 
> 1. I think the parameter `namespace_pid` is confusing and should be changed.  I'm thinking `container_pid`.  We originally talked about attaching these programs in "namespaces", but it's not really a "namespace" in the k8s sense but rather a container (which does use linux namespaces++ and may be in a different k8s namespace from the user, but may not).  I'd also change the executable from `bpfman-ns` to something like `bpfman-container`.  This change will ripple through the code, so I'd like to get some consensus before I do it.

`container_pid` makes sense. Stay with `bpfman-ns` - it's not user-facing so we can change it whenever.
Once this has merged we can look at folding this into a dedicated (hidden) command like `bpfman attach-in-container` or something similar.

> 2. Should we use Docker to create containers for integration test, or is it better to create our own "containers" using the Linux command line?

No. We should use https://github.com/testcontainers/testcontainers-rs
IIRC testcontainers had some support for podman - and we can document both in the integration-test docs.
It will be a lot easier than manually creating fake containers ourselves.

astoycos:
So talking to @anfredette offline, I wrote previously 

> From an external(API) POV I think the container ID/Name is what should be provided by the user. So then a user can easily express that the uprobe should be attached to the process in the specified container and bpfman internally can just map the container name to the actual container PID. Having a user have to specify containerPID seems ultra low level.

Due to the fact that using a container ID/name would require us to interface more closely with a container runtime (podman, Docker) I'm actually alright with just using container_pid as the API for now.

proto/bpfman.proto
```diff
@@ -110,7 +110,7 @@ message KprobeAttachInfo {
     string fn_name = 1;
     uint64 offset = 2;
     bool retprobe = 3;
-    optional string namespace = 4;
+    optional int32 namespace_pid = 4;
```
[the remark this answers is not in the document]
anfredette: Good point.  I don't mind changing this, but it may not be necessary yet.  First, the namespace option isn't supported yet.  There's no way to use it from kubernetes, and if someone tries it from bpfctl or bpfman, they'll get an error.  Second, we're going to be making major changes in 0.4.0, so I don't know if we need to worry about backward compatibiilty yet. Let's discuss during the meeting tomorrow.

bpfman/src/bpf.rs
anfredette: Done

bpfman/src/bpf.rs
anfredette: Done.

bpfman/src/bpf.rs
anfredette: Done

astoycos (approved):
LGTM Thanks @anfredette 

Please remember to make any follow up issues we discussed

dave-tucker (approved):
LGTM 🌴

PR #794, opened 2023-11-01
Don't try and pin .data maps

astoycos:
Bpfd should skip pinning .data section maps.


dave-tucker (changes requested):
```rust
fn should_be_pinned(name: &str) -> bool {
    return name.contains(".rodata") || name.contains(".bss") || name.contains(".data")
}
```

Nit: Would be neater to contain this in a util function vs. repeating it 3 times.

astoycos:
Done!

dave-tucker (approved):

PR #796, opened 2023-11-03
Fix BPF Licensing

dave-tucker:
- Update README description of BPF licensing
- Correct SPDX headers for XDP dispatchers
- Correct/Add SPDX headers for bpfd's own bpf code
- clang-formats C files, this should be a lint in CI


anfredette (commented):
LGTM

PR #817, opened 2023-11-13
bpfd: Use tc dispatcher from container image

dave-tucker:


.github/workflows/image-build.yaml
```diff
@@ -354,6 +354,16 @@ jobs:
...
               type=raw,value=v2,enable=true
 
+          - registry: quay.io
+            bpf_build_wrapper: rust
+            repository: bpfd
+            image: tc-dispatcher
+            context: .
+            dockerfile: ./Containerfile.tc_dispatcher
+            tags: |
+              type=sha,format=long
```
astoycos, line 364: Only one nit that already existed should we also push dispatcher images whenever we release In case we need to make breaking changes? i.e for the xdp_dispatcher_v2 and the tc_dispatcher just add 

```
type=ref,event=tag
```

dave-tucker: I'm not sure. The dispatcher images change less often. I'm not sure what impact if any there would be on upgrade from 0.3 > 0.4 if we change the tags - guessing it would still fetch from quay again? Either way, we can always fix this later.

astoycos: ACK

astoycos (approved):
LGTM

Issue #741, opened 2023-09-28
Remove HTTPS/TLS from all the things


dave-tucker:
- Unix socket is more practical
- We get to lose the OpenSSL deps (and perhaps rustls)

PR #819, opened 2023-11-15
bpfd, bpfctl, operator: Remove support for TCP/TLS

dave-tucker:
This removes support for the use of TCP/TLS endpoints.

bpfd/bpfctl will default to using the unix socket. Permissions there may be controlled using users/groups.

The bpfd-operator will also use the unix socket but through a volume mount that is shared with bpfd.

TCP/TLS isn't required for any practical reason - we only want to support on-the-same-box communication - and the dependencies required to implement this support were vast. We may chose to revisit this in future.

Fixes: #741


bpfd-operator/cmd/bpfd-agent/main.go
```diff
@@ -111,17 +111,8 @@ func main() {
...
-		if err != nil {
-			setupLog.Error(err, "Failed to generate credentials for new client")
-			os.Exit(1)
-		}
-	} else {
-		setupLog.Info("No MTLS credentials provided, using insecure connection")
-		creds = insecure.NewCredentials()
-	}
+	setupLog.Info("Connecting over UNIX socket to bpfd")
+	creds = insecure.NewCredentials()
```
astoycos, line 115: nit
```suggestion
	setupLog.Info("Connecting over UNIX socket to bpfd")
	creds := insecure.NewCredentials()
```

And remove

astoycos (approved):
LGTM

PR #820, opened 2023-11-15
Fix some panics + add testing and fix for map sharing 

astoycos:
See commit messages for details.

### DIFF
--- a/bpfd/src/oci_utils/image_manager.rs
+++ b/bpfd/src/oci_utils/image_manager.rs
@@ -309,7 +309,8 @@ impl ImageManager {
             .open(bytecode_path.clone())
             .map_err(|e| ImageError::ByteCodeImageProcessFailure(e.into()))?;
 
-        let image_config: Value = serde_json::from_str(&config_contents).unwrap();
+        let image_config: Value = serde_json::from_str(&config_contents)
+            .map_err(|e| ImageError::ByteCodeImageProcessFailure(e.into()))?;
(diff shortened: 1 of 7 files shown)


astoycos:
CI will fail here until I manually push images OR we merge broken

bpfd/src/oci_utils/image_manager.rs
```diff
@@ -493,7 +494,9 @@ fn load_image_meta(image_content_path: PathBuf) -> Result<ContainerImageMetadata
     let file_content =
         read_to_string(image_config_path).context("failed to read image config file")?;
 
-    let image_config: Value = serde_json::from_str(&file_content).unwrap();
+    // This should panic since it means that the on disk memory format has changed during runtime.
```
dave-tucker, line 497: on disk memory? you mean on disk storage format?

dave-tucker (changes requested):
One nit, and were you also planning to add some documentation around the LIBBPF_PIN_BY_NAME flag?

bpfd/src/oci_utils/image_manager.rs
astoycos: Lol yerp thanks!

astoycos:
> One nit, and were you also planning to add some documentation around the LIBBPF_PIN_BY_NAME flag?

Ah yeah I'll do this, we definitely need an interim note in the docs

dave-tucker (approved):

PR #827, opened 2023-11-21
Document valid kernel versions

Billy99:


anfredette (commented):
This LGTM, but I'll let someone else do that approval since I was out for most of the discussion.

As a side note, I think it would be good to test and document the results for some Ubuntu releases in the future.  I'll open an issue for this.

astoycos (approved):
Really good work here @Billy99 Thanks!

PR #846, opened 2023-11-27
Running `cargo xtask build-proto` resulted in some changes

anfredette:
Perhaps it's because I have a newer version of `protoc`.


astoycos (approved):
LGTM

PR #861, opened 2023-12-08
bpfman: in memory db setup + oci_utils conversion

astoycos:
Prep for using sled as the in memory db for all state persistence so that we no longer have to write directly to disk.

Add some error handling fixes.

Convert to using sled for the OCI image store. This ensures that no data is written to the filesystem directly and that everything goes through our sled db instance.

Works towards finishing https://github.com/bpfman/bpfman/issues/860

TODO convert all `BpfManager` state management to use sled as well.


astoycos:
@dave-tucker This should be mostly there (I think I still have an int test flake or two I'm tracking down) 

Decided to break this up a bit over two PRs one for the image manager ^^ and the other for the BpfManager

bpfman/src/oci_utils/image_manager.rs
```diff
@@ -347,60 +320,50 @@ impl ImageManager {
...
         &self,
-        content_dir: String,
-    ) -> Result<Vec<u8>, anyhow::Error> {
-        debug!("bytecode is stored as tar+gzip file at {}", content_dir);
-        let image_dir = Path::new(&content_dir);
-        // Get image manifest from disk
-        let manifest = load_image_manifest(image_dir.to_path_buf())?;
+        base_key: String,
+    ) -> Result<Vec<u8>, ImageError> {
+        // Get image manifest from database convert back to a &str with the as_ref() implementation on IVec
```
dave-tucker, line 337: Nit: The comment is the code

bpfman/src/oci_utils/image_manager.rs
```diff
@@ -435,79 +397,66 @@ impl ImageManager {
...
-            _ => match image.digest() {
-                Some(d) => d,
-                _ => "latest",
-            },
-        };
+    fn load_image_meta(
+        &self,
+        image_content_key: &str,
+    ) -> Result<ContainerImageMetadata, anyhow::Error> {
+        // Get image manifest from database convert back to a &str with the as_ref() implementation on IVec
```
dave-tucker, line 404: Nit: As above

bpfman/src/oci_utils/image_manager.rs
```diff
@@ -435,79 +397,66 @@ impl ImageManager {
...
+            .database
+            .get(image_config_key)
+            .map_err(ImageError::DatabaseReadFailure)?
+            .expect("Image manifest is empty");
 
-fn load_image_manifest(image_dir: PathBuf) -> Result<OciImageManifest, anyhow::Error> {
-    let manifest_path = image_dir.join("manifest.json");
+        let file_content = std::str::from_utf8(db_content)?;
+
+        // This should panic since it means that the on disk storage format has changed during runtime.
```
dave-tucker, line 429: Is this comment still valid?

bpfman/src/oci_utils/image_manager.rs
```diff
@@ -611,33 +579,27 @@ mod tests {
...
-            Case{input:"test.com/repo:tag", output: tmpdir.as_ref().to_path_buf().join("test.com/repo/tag")},
+            Case{input: "busybox", output: "docker.io_library_busybox_latest"},
+            Case{input:"quay.io/busybox", output: "quay.io_busybox_latest"},
+            Case{input:"docker.io/test:tag", output: "docker.io_library_test_tag"},
+            Case{input:"quay.io/test:5000", output: "quay.io_test_5000"},
+            Case{input:"test.com/repo:tag", output: "test.com_repo_tag"},
             Case{
                 input:"test.com/repo@sha256:ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
-                output: tmpdir.as_ref().to_path_buf().join("test.com/repo/sha256:ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"),
+                output: "test.com_repo_sha256:ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
```
dave-tucker, line 595: Is this separator safe to use?

What happens if my repository is called `dave_tucker`...
Also what happens if the digest is not a SHA256 - can we reconstruct it in the same way?

bpfman/src/oci_utils/mod.rs
```diff
@@ -20,4 +21,14 @@ pub enum ImageError {
     ByteCodeImageProcessFailure(#[from] anyhow::Error),
     #[error("BytecodeImage not found: {0}")]
     ByteCodeImageNotfound(String),
+    #[error("Failed to flush database to disk: {0}")]
```
dave-tucker, line 24: I'm not happy with all these error types 😉 since I don't think we ever handle the invariants. Might as well just have `DatabaseError(String)` or get fancy and have both String and optionally the source.

bpfman/src/serve.rs
```diff
@@ -60,13 +61,18 @@ pub async fn serve(
     let allow_unsigned = config.signing.as_ref().map_or(true, |s| s.allow_unsigned);
     let (itx, irx) = mpsc::channel(32);
 
-    let mut image_manager =
-        ImageManager::new(STDIR_BYTECODE_IMAGE_CONTENT_STORE, allow_unsigned, irx).await?;
+    // Before dropping this make sure to manually call flush on the database.
```
dave-tucker, line 64: This comment would be better near the call to flush

tests/integration-test/src/tests/basic.rs
```diff
@@ -413,18 +411,12 @@ fn test_load_unload_kretprobe() {
...
 fn test_pull_bytecode() {
-    if std::path::PathBuf::from(STDIR_BYTECODE_IMAGE_CONTENT_STORE).exists() {
-        std::fs::remove_dir_all(STDIR_BYTECODE_IMAGE_CONTENT_STORE).unwrap();
-    }
-
     let _bpfman_guard = start_bpfman().unwrap();
 
     debug!("Pull bytecode image");
 
+    // Just ensure this doesn't panic
```
dave-tucker, line 418: That's `assert!(bpfman_pull_bytecode().is_ok())`

bpfman/src/oci_utils/image_manager.rs
astoycos: Yeah it's still safe to use, in that case it would just look like `dave_tucker_repo_sha256:fffff`

> Also what happens if the digest is not a SHA256 - can we reconstruct it in the same way?

This function just ensure we can go from IMAGE_URL to database key, as long as the digest is represented as a string we should always be fine.

bpfman/src/oci_utils/mod.rs
astoycos: Fixed up made this into `    #[error("{0}: {1}")]
    DatabaseError(String, String),`  where the first string is a context and the second is the actual error since sled::Error implements `to_string()` 👍

dave-tucker (approved):

Issue #703, opened 2023-09-13
Generate a man page for bpfman


anfredette:
Can we use https://crates.io/crates/clap_mangen?

dave-tucker:
Yep. We would use an xtask instead of build.rs.

dave-tucker:
Hi @weiyuhang2011!
Yes it would be awesome to have manpages.
Since we only have one binary called `bpfman` now, we'd just need the manpage for that one. I'll update the issue title appropriately.

weiyuhang2011:
> Hi @weiyuhang2011! Yes it would be awesome to have manpages. Since we only have one binary called `bpfman` now, we'd just need the manpage for that one. I'll update the issue title appropriately.

Hi @dave-tucker , the problem I found is that it is hard to import bpfman cli struct dependency in xtask because it is a bin crate. Including the whole mod.rs file from cli doesn't look right because the dependency issue of that file need to be solved. Shall we refactor the cli part into a lib crate?

dave-tucker:
> Hi @dave-tucker , the problem I found is that it is hard to import bpfman cli struct dependency in xtask because it is a bin crate. Including the whole mod.rs file from cli doesn't look right because the dependency issue of that file need to be solved. Shall we refactor the cli part into a lib crate?

I don't think we need to go that far. I'll try moving all the structs that clap uses into `args.rs` and then you should be able to simply `include!("../../bpfman/src/cli/args.rs)` and then generate the manpages using clap_mangen. It's not an easy refactor, so leave it with me and I'll get a PR done in the next few hours - I'll tag you here when it's done so you can carry on with the manpage generation.

dave-tucker:
@weiyuhang2011 ok now #870 has merged you should be unblocked 🎉 

I had a quick try with `include!(../../bpfman/src/cli/args.rs)` from an xtask and I was able to get manpage generation to work.
A couple of pointers:
- You'll want to use the module attribute `!#[allow_dead_code]` since some of the included code isn't used and this is the easiest way to suppress those warnings.
- We'll need to generate, `bpfman.1`, `bpfman-load.1` etc.. - one manpage for each command/subcommand
- `let cli = Cli::command();` is what you need to pass into `clap_mangen`, which requires using the `clap::CommandFactory` trait.

I'll also go ahead and generate an issue for completions since these should also be easy enough to add.

weiyuhang2011:
@dave-tucker Thank you for your quick response! I'm working on it.

PR #870, opened 2023-12-12
bpfman: cli: Refactor command handling

dave-tucker:
This moves all clap code into args.rs to allow it to be used via the include! macro in an xtask to generate manpages.

Updates: #703


bpfman/src/cli/args.rs
```diff
@@ -0,0 +1,344 @@
...
+    pub(crate) csi_support: bool,
+}
+
+#[derive(Subcommand, Debug)]
+pub(crate) enum SystemSubcommand {
+    /// Load an eBPF program from a local .o file.
+    Service(ServiceArgs),
+}
+
+/// Parse a single key-value pair
```
astoycos, line 323: nit these should probably live in a `util.rs`

astoycos (approved):
LGTM

One small nit which I'll make a good-first-issue for I like this organization much better.

bpfman/src/cli/args.rs
anfredette: They could, but they are only useed by the functions in this file, so I think it's okay.

anfredette (commented):
LGTM too.

Issue #463, opened 2023-05-02
Set up socket activation for the unix socket


dave-tucker:

astoycos:
linked to #90

astoycos:
@dave-tucker Please add more info

PR #872, opened 2023-12-12
bpfman: add support for socket activation

Billy99:
When loading bpfman as a systemd service, implement socket activation. Socket activation is a systemd feature in which a given service is not started until it is needed, in this case, a request is received on a socket. To reduce the attack surface, it is preferred that bpfman not run forever, but instead stop after a period of inactivity. The current
implementation has a hard stop of bpfman after 15 seconds. Trying to restart timers over multiple async threads was not worth the effort since async nature of bpfman is going to be ripped out over the nextcouple of weeks.

NOTE: Systemd can support stopping a service after a specified time of inactivity. But it requires the builtin systemd proxy, where systemd listens on the primary socket and when a request comes in the request is sent to bpfman over a private socket. It was decided that this was not the correct solution for bpfman.

Resolves: #463

### DIFF
--- a/bpfman/src/serve.rs
+++ b/bpfman/src/serve.rs
@@ -1,13 +1,19 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Authors of bpfman
 
-use std::{fs::remove_file, path::Path};
+use std::{
+    fs::remove_file,
+    os::unix::prelude::{FromRawFd, IntoRawFd},
+    path::Path,
+};
 
+use anyhow::anyhow;
 use bpfman_api::{
     config::Config,
     util::directories::{RTPATH_BPFMAN_SOCKET, STDIR_DB},
     v1::bpfman_server::BpfmanServer,
 };
+use libsystemd::activation::IsType;
 use log::{debug, info};
 use sled::Config as DbConfig;
 use tokio::{
@@ -43,10 +49,8 @@ pub async fn serve(
 
     let mut listeners: Vec<_> = Vec::new();
 
-    match serve_unix(path.clone(), service.clone()).await {
-        Ok(handle) => listeners.push(handle),
-        Err(e) => eprintln!("Error = {e:?}"),
-    }
+    let (handle, use_activity_timer) = serve_unix(path.clone(), service.clone()).await?;
+    listeners.push(handle);
 
     let allow_unsigned = config.signing.as_ref().map_or(true, |s| s.allow_unsigned);
     let (itx, irx) = mpsc::channel(32);
@@ -58,7 +62,7 @@ pub async fn serve(
 
     let mut image_manager = ImageManager::new(database.clone(), allow_unsigned, irx).await?;
     let image_manager_handle = tokio::spawn(async move {
-        image_manager.run().await;
+        image_manager.run(use_activity_timer).await;
     });
 
     let mut bpf_manager = BpfManager::new(config.clone(), rx, itx, database);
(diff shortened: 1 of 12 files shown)


bpfman/src/serve.rs
```diff
@@ -42,6 +48,8 @@ pub async fn serve(
 
     let mut listeners: Vec<_> = Vec::new();
 
+    let mut use_activity_timer = false;
```
astoycos, line 51: why is this mut?

bpfman/src/serve.rs
```diff
@@ -117,12 +126,25 @@ pub async fn serve(
     Ok(())
 }
 
-pub(crate) async fn shutdown_handler() {
+pub(crate) async fn shutdown_handler(use_activity_timer: bool) {
+    const TIMEOUT: std::time::Duration = std::time::Duration::from_secs(15);
```
astoycos, line 130: I think you can use https://blog.logrocket.com/rust-lazy-static-pattern/ for this since then it won't actually get initialized unless it's used

bpfman/src/serve.rs
```diff
@@ -117,12 +126,25 @@ pub async fn serve(
     Ok(())
 }
 
-pub(crate) async fn shutdown_handler() {
+pub(crate) async fn shutdown_handler(use_activity_timer: bool) {
```
astoycos, line 129: Why not just have two methods `shutdown_handler_timer()` `shutdown_handler()` rather than all the stuff with `&mut bool` which is kind of a weird way to send signals between threads

bpfman/src/serve.rs
```diff
@@ -162,3 +181,38 @@ async fn serve_unix(
...
+        remove_file(&path)?;
+    }
+
+    let uds = UnixListener::bind(&path)?;
+    let stream = UnixListenerStream::new(uds);
+    // Always set the file permissions of our listening socket.
+    set_file_permissions(&path.clone(), SOCK_MODE).await;
+
+    info!("Using default Unix socket");
+    Ok(stream)
```
astoycos, line 217:
```suggestion
    // Always set the file permissions of our listening socket.
    set_file_permissions(&path.clone(), SOCK_MODE).await;

    info!("Using default Unix socket");
    Ok(UnixListenerStream::new(uds));
```

Can you just do this?

bpfman/src/serve.rs
```diff
@@ -138,21 +160,18 @@ async fn join_listeners(listeners: Vec<JoinHandle<()>>) {
...
-    let uds = UnixListener::bind(&path)?;
-    let uds_stream = UnixListenerStream::new(uds);
-    // Always set the file permissions of our listening socket.
-    set_file_permissions(&path.clone(), SOCK_MODE).await;
+    let uds_stream = if let Ok(stream) = systemd_unix_stream(path.clone()) {
+        *use_activity_timer = true;
+        stream
+    } else {
+        std_unix_stream(path.clone()).await?
+    };
```
astoycos, line 170: You could just assign a shutdown handler var here here OR just call serve with the different shutdown methods

bpfman/src/serve.rs
```diff
@@ -50,7 +58,7 @@ pub async fn serve(
                     continue;
                 }
 
-                match serve_unix(path.clone(), service.clone()).await {
+                match serve_unix(path.clone(), service.clone(), &mut use_activity_timer).await {
```
astoycos, line 61: Or you could just return a boolean here and pass it onwards

bpfman/src/storage.rs
```diff
@@ -361,7 +361,7 @@ impl StorageManager {
         }
     }
 
-    pub async fn run(self) {
+    pub async fn run(self, use_activity_timer: bool) {
         let path: &Path = Path::new(RTPATH_BPFMAN_CSI_SOCKET);
```
astoycos, line 365: OR we could pass around the shutdown handler as a function arg

astoycos (commented):
Looks good! The only part I don't like is passing around the `&mut bool` everywhere I left some opinions on how we could change this feel free to ignore

Additionally do we need an integration test for this?

dave-tucker:
@Billy99 this needs to rebased on top of your other patch

dave-tucker (approved):
As discussed, this is green in CI so LGTM, considering that we're about to go nuts on rewriting parts of the codebase in preparation for the 0.4.0 release

Issue #860, opened 2023-12-08
Convert all state management to go through sled


astoycos:
As described in the [bpfman damonless design doc](https://github.com/bpfman/bpfman/blob/main/docs/design/daemonless.md#state-management)
we would like to move all our state persistence to an in memory db, for that we've chose to use [sled](https://github.com/spacejam/sled).

I'll break this up into three big chunks:

- [x] Setup + Image Manager -> https://github.com/bpfman/bpfman/pull/861
- [x] Core Program Type -> https://github.com/bpfman/bpfman/pull/874
- [x] XdpDispatcher -> https://github.com/bpfman/bpfman/pull/901
- [x] TCDispatcher

astoycos:
This is done!!! 🥳

PR #874, opened 2023-12-15
Sled integration for the core program type

astoycos:
Works towards #860 

This commit is huge and it re-writes the memory layout for all Program
related information. More specifically it removed the KernelInfo struct
and collapses all information for bpfman programs AND other programs
into a single Program enum. This program Enum is essentially a wrapper
for a sled database tree. Each program will get it's own sled database tree
and anytime a getter/setter is called on a Program it will interact with
the database.

Special considerations: 

- On load we have to use a random number generator to generate a unique id
for the program tree prior to load.  Once the program is successfully
loaded we can then swap the root index of the database tree to the
correct id.  From then on it's always safe to assume the tree name will
correlate to a given program id.

- On list and get we will write the kernel information for programs not loaded via bpfman to the database in ONLY TEMPORARILY, specifically they will be written in bpf.rs's list_programs and deleted from the db in rpc.rs.  This will most likey be removed when removing our GRPC service, and we can instead write a function to just return the necessary information from aya::program::loaded_programs.

- The ROOT DB last static handle is the global "root" for our sled db instance, currently we use it in the BPFMANAGER and IMAGEMANAGER and we make sure to flush before exiting either of those threads.

- TODO The next PR will try and move even more of the Dispatcher related state to the in memory-db with the goal of being able to completely remove 90% of our "rebuild" logic.

**Please focus review on our internal API (command.rs) which contains many of the core bits.**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -82,9 +82,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.4"
+version = "0.6.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2ab91ebe16eb252986481c5b62f6098f3b698a45e34b5b98200cf20dd2484a44"
+checksum = "d664a92ecae85fd0a7392615844904654d1d5f5514837f471ddef4a057aba1b6"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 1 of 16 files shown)


astoycos:
This is almost there just need a couple more rounds of proof-reading + debugging

Billy99 (commented):
I spent some time today looking over the patch and running it locally. I didn't find anything glaring, but would need more time to go through it completely. LGTM if Dave wants to merge it, I can always continue playing with it after the break.

dave-tucker (approved):
Like with Billy's PR, approving since CI is green and the general design LGTM.
There will be nits, sure, but we'll address those when rewriting everything for the 0.4.0 release.

PR #885, opened 2023-12-20
bpfman: Remove grpc section from config

Billy99:
At one point, the ConfigFile (/etc/bpfman/bpfman.toml) could be used to configure the gRPC endpoint as TCP/TLS or Unix socket. TCP/TLS was removed and now only Unix Sockets are valid, so this configuration is now useless. So removing the gRPC Endpoint from the ConfigFile.

bpfman-agent was only using this field from the ConfigFile, so it no longer needs to read the ConfigFile. So all ConfigFile logic was removed from bpfman-operator. The only gotcha is that bpfman-agent was using the gRPC Endpoint to change the socket path from the default path of /run/bpfman/bpfman.sock to /bpfman-sock/bpfman.sock. This was because the bpfman agent doesn't have permissions to access the host. To work around this fact, the default path was moved to a subdirectory of /run/bpfman/sock/bpfman.sock.


dave-tucker (approved):

bpfman/src/storage.rs
```diff
@@ -400,11 +400,7 @@ impl StorageManager {
     }
 
     #[allow(dead_code)] // TODO: Remove this when the storage manager is fully implemented
-    fn pin_map_to_bpffs(
-        &self,
-        source_object: &mut MapData,
-        dest_bpffs: &Path,
-    ) -> anyhow::Result<()> {
+    fn pin_map_to_bpffs(&self, source_object: &MapData, dest_bpffs: &Path) -> anyhow::Result<()> {
```
astoycos, line 403: Did the compiler prompt this change?

astoycos (commented):
I saw this already merged just a few comments and nits we can cleanup later

PR #901, opened 2024-01-04
Sled fixes, convert xdpdispatcher to use sled, fix loaded_programs race

astoycos:
Continues to work towards #860

### DIFF
--- a/bpfman-api/src/config.rs
+++ b/bpfman-api/src/config.rs
@@ -7,6 +7,8 @@ use aya::programs::XdpFlags;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
+use crate::ParseError;
+
 #[derive(Debug, Deserialize, Default, Clone)]
 pub struct Config {
     pub interfaces: Option<HashMap<String, InterfaceConfig>>,
@@ -61,6 +63,21 @@ impl XdpMode {
             XdpMode::Skb => XdpFlags::SKB_MODE,
             XdpMode::Drv => XdpFlags::DRV_MODE,
             XdpMode::Hw => XdpFlags::HW_MODE,
+        }
+    }
+}
+
+impl TryFrom<u32> for XdpMode {
+    type Error = ParseError;
+
+    fn try_from(mode: u32) -> Result<Self, Self::Error> {
+        match mode {
+            0 => Ok(XdpMode::Skb),
+            1 => Ok(XdpMode::Drv),
+            2 => Ok(XdpMode::Hw),
+            _ => Err(ParseError::InvalidXdpMode {
+                mode: mode.to_string(),
+            }),
         }
     }
 }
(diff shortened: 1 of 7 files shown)


bpfman-api/src/config.rs
```diff
@@ -63,6 +63,15 @@ impl XdpMode {
             XdpMode::Hw => XdpFlags::HW_MODE,
         }
     }
+
+    pub fn from_u32(val: u32) -> Self {
```
dave-tucker, line 67: This should either be `impl TryFrom<u32>` or `impl From<u32>` - I'd lean towards the former to avoid the panic.

bpfman/src/bpf.rs
```diff
@@ -859,8 +873,14 @@ impl BpfManager {
 
         // Call Aya to get ALL the loaded eBPF programs, and loop through each one.
         loaded_programs()
-            .map(|p| {
-                let prog = p.map_err(BpfmanError::BpfProgramError)?;
+            .filter_map(|p| match p {
```
dave-tucker, line 876:
```suggestion
            .flat_map(|p| p.ok())
```
Which will silently ignore errors - there is no point in logging warn messages since there is nothing the user can do about it. This also avoids using option types in the map closure that immediately follows this.

bpfman/src/multiprog/xdp.rs
```diff
@@ -285,59 +310,83 @@ impl XdpDispatcher {
...
     }
 
-    pub(crate) fn if_name(&self) -> String {
-        self.if_name.clone()
+    pub(crate) fn set_revision(&mut self, revision: u32) -> Result<(), BpfmanError> {
+        sled_insert(&self.db_tree, "revision", &revision.to_ne_bytes())
+    }
+
+    pub(crate) fn get_revision(&self) -> Result<u32, BpfmanError> {
+        sled_get(&self.db_tree, "revision").map(bytes_to_u32)
```
dave-tucker, line 346: nit: I guess these DB keys should probably be `const` to avoid typos in set vs get. Not going to block this PR on it though.

dave-tucker (changes requested):
Great work! A couple of small comments.

bpfman/src/multiprog/xdp.rs
astoycos: Great first issue! https://github.com/bpfman/bpfman/issues/912

bpfman/src/bpf.rs
astoycos: Done!

bpfman-api/src/config.rs
astoycos: Done!

astoycos: (I think XdpMode should maybe be moved to lib.rs in the future though)

dave-tucker (approved):

PR #903, opened 2024-01-05
Tidy up async code and shutdown handling

dave-tucker:
1. Removes custom Tokio runtime construction
2. Uses a broadcast channel for the shutdown handler so each of the "managers" may recieve the message and shutdown.

### DIFF
--- a/bpfman/src/serve.rs
+++ b/bpfman/src/serve.rs
@@ -12,14 +12,13 @@ use bpfman_api::{
     config::Config, util::directories::RTPATH_BPFMAN_SOCKET, v1::bpfman_server::BpfmanServer,
 };
 use libsystemd::activation::IsType;
-use log::{debug, info};
+use log::{debug, error, info};
 use tokio::{
     join,
     net::UnixListener,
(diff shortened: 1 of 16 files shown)


bpfman/src/serve.rs
```diff
@@ -33,7 +32,13 @@ use crate::{
     ROOT_DB,
 };
 
-pub async fn serve(config: &Config, csi_support: bool) -> anyhow::Result<()> {
+pub async fn serve(config: &Config, csi_support: bool, timeout: u64) -> anyhow::Result<()> {
+    let (shutdown_tx, shutdown_rx1) = broadcast::channel(32);
+    let shutdown_rx2 = shutdown_tx.subscribe();
```
astoycos, line 37: nit these names aptly would me nice,  `storage_shutdown_rx`, `bpf_manager_shutdown_rx`

astoycos (approved):
LGTM 

Not going to hold this up over one small nit, the broadcast queue is much better suited for the shutdown usecase 👏

PR #910, opened 2024-01-08
Sled tc dispatcher

astoycos:
Finishes https://github.com/bpfman/bpfman/issues/860 depends on the changes in #901  (I'll leave as a draft until that work merges)

### DIFF
--- a/bpfman-api/src/config.rs
+++ b/bpfman-api/src/config.rs
@@ -65,15 +65,6 @@ impl XdpMode {
             XdpMode::Hw => XdpFlags::HW_MODE,
         }
     }
-
-    pub fn from_u32(val: u32) -> Self {
-        match val {
-            0 => XdpMode::Skb,
-            1 => XdpMode::Drv,
-            2 => XdpMode::Hw,
-            _ => panic!("unable to marshall u32 to XdpMode"),
-        }
-    }
 }
 
 impl TryFrom<u32> for XdpMode {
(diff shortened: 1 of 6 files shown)


bpfman-api/src/config.rs
```diff
@@ -63,6 +63,15 @@ impl XdpMode {
             XdpMode::Hw => XdpFlags::HW_MODE,
         }
     }
+
+    pub fn from_u32(val: u32) -> Self {
```
astoycos, line 67: Convert to tryfrom

dave-tucker: Thanks! Saved me adding that comment here lol

bpfman/src/utils.rs
```diff
@@ -81,6 +81,14 @@ pub(crate) fn bytes_to_u32(bytes: Vec<u8>) -> u32 {
     )
 }
 
+pub(crate) fn bytes_to_u16(bytes: Vec<u8>) -> u16 {
+    u16::from_ne_bytes(
+        bytes
+            .try_into()
+            .expect("unable to martial &[u8] to &[u8; 4]"),
```
dave-tucker, line 88: Just saw this one, it should be `Marshall` not `Martial`

dave-tucker (changes requested):

dave-tucker (changes requested):
Couple of nits and the `TryFrom` to do, but otherwise LGTM

bpfman/src/utils.rs
astoycos: Good catch :)

astoycos:
Done!

dave-tucker (approved):

PR #923, opened 2024-01-19
refactor: make all sled-db keys into const

### LINKED ISSUE #912: Make all sled-db keys `const`'s
Now that we're focusing on sled database there's a number of get and set operations we maintain similar to the following:

```
    pub(crate) fn set_name(&mut self, name: &str) -> Result<(), BpfmanError> {
        sled_insert(&self.db_tree, "name", name.as_bytes())
    }

    pub(crate) fn get_name(&self) -> Result<String, BpfmanError> {
        sled_get(&self.db_tree, "name").map(|v| bytes_to_string(&v))
    }
```

All of these keys should be turned into rust `const` variables to avoid typos in set/get functions!

shawnh2:
close #912


astoycos:
Hiya @shawnh2!! Thanks for opening a PR, I'll check it out, in the meantime looks like you need to run `cargo fmt` 👍

Billy99:
@shawnh2 Thank you for your PR. Looks like you just have a few formatting errors. See https://bpfman.io/main/governance/contributing/#pull-request-checklist

shawnh2:
@astoycos @Billy99 hi, the ci formatting issue has been resolved, PTAL

bpfman/src/command.rs
```diff
@@ -36,6 +36,62 @@ use crate::{
     ROOT_DB,
 };
 
+const SLED_DB_KEY_KIND: &str = "kind";
```
astoycos, line 39: so this looks alright but can you please just remove the `SLED_DB_KEY` prefixes from all these variable names in favor of just a code comment above the block where the keys are defined?

astoycos (commented):
Thanks @shawnh2 generally looks good! One nit and if you could also do the same conversion in https://github.com/bpfman/bpfman/tree/main/bpfman/src/multiprog that'd be great!

bpfman/src/command.rs
shawnh2: done

shawnh2:
> Thanks @shawnh2 generally looks good! One nit and if you could also do the same conversion in https://github.com/bpfman/bpfman/tree/main/bpfman/src/multiprog that'd be great!

sure, the conversion has been made. PATL

astoycos (approved):
LGTM 

@shawnh2 Thanks for the first contribution 🥳 Can you please squash your commits (adding hold until then) and then this should be good to go!

PR #935, opened 2024-01-22
cli: fix nits in manpage and tab-completion

Billy99:
There are a couple of nits in the CLI manpage and tab-completion generation. This PR addresses most of them.

* Document how to build and install. It's not obvious how to install the generated files.
* We tend not to use environment variables to build bpfman. Moved OUT_DIR to a fixed location.
* Only top level manpages are generated. `man bpfman get` and `man bpfman list` are implemented, but those with subcommands, primarily `man bpfman load`, don't document any of the subcommand parameter or usage.
* Update the install scripts/setup.sh to copy the generated files, if they have been generated. Remove them on uninstall.
* Call the new xtask commands in CI

Not completed:
* Add the generated files to the RPM build
* To show the load subcommands, hyphens are required, this should be fixed: `man bpfman load-file-xdp`

### DIFF
--- a/bpfman/src/cli/args.rs
+++ b/bpfman/src/cli/args.rs
@@ -6,7 +6,11 @@ use clap::{Args, Parser, Subcommand};
 use hex::FromHex;
 
 #[derive(Parser, Debug)]
-#[command(author, version, about, long_about = None)]
+#[command(
+    long_about = "An eBPF manager focusing on simplifying the deployment and administration of eBPF programs."
+)]
+#[command(name = "bpfman")]
+#[command(disable_version_flag = true)]
 pub(crate) struct Cli {
     #[command(subcommand)]
     pub(crate) command: Commands,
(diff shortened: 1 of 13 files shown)


bpfman/src/cli/args.rs
```diff
@@ -6,7 +6,11 @@ use clap::{Args, Parser, Subcommand};
 use hex::FromHex;
 
 #[derive(Parser, Debug)]
-#[command(author, version, about, long_about = None)]
+#[command(
+    long_about = "CLI for interacting with bpfman. Used to load, unload and view eBPF programs on a system."
```
dave-tucker, line 10: This should say something like `An eBPF Program Manager built for Linux and Kubernetes` or whatever we use in our README and or Cargo.toml as our description.

dave-tucker (commented):
LGTM other than one little nit on the long about.

bpfman/src/cli/args.rs
Billy99: Done

dave-tucker (approved):
LGTM

PR #956, opened 2024-01-25
sled fixups and prefixes for final conversion bits

astoycos:
Add prefixes to the program tree names in sled and remove the erroneous id structure field which was no longer needed.

Still need to 

- [x] Add a `new_from_db` method for the `Program` enumeration
- [x] Flatten the ProgramMap and DispatcherMap into BpfManager
- [x] Convert all bpfMap bits into sled

### DIFF
--- a/bpfman/src/command.rs
+++ b/bpfman/src/command.rs
@@ -1380,7 +1380,7 @@ impl Program {
 
     pub(crate) fn delete(&self) -> Result<(), anyhow::Error> {
         let id = self.get_data().get_id()?;
-        ROOT_DB.drop_tree(id.to_string())?;
+        ROOT_DB.drop_tree(PROGRAM_PREFIX.to_string() + id.to_string().as_str())?;
 
         let path = format!("{RTDIR_FS}/prog_{id}");
         if PathBuf::from(&path).exists() {
@@ -1498,5 +1498,5 @@ impl Program {
 // it.
 #[derive(Debug, Clone)]
 pub(crate) struct BpfMap {
-    pub(crate) used_by: Vec<u32>,
+    pub(crate) _used_by: Vec<u32>,
 }
--- a/bpfman/src/command.rs
+++ b/bpfman/src/command.rs
@@ -38,6 +38,7 @@ use crate::{
 
 /// These constants define the key of SLED DB
 pub(crate) const PROGRAM_PREFIX: &str = "program_";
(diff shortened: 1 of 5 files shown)


astoycos:
OK @dave-tucker This should do it! Sorry for the delay it took me a while to unravel all of the delicate data lifetime management we had baked in over the past year 💀

bpfman/src/command.rs
```diff
@@ -375,15 +365,12 @@ impl ProgramData {
         let id_rand = rng.gen::<u32>();
```
dave-tucker, line 365: Looking at this again I've realized we'll have issues here.
There's a `N in u32::max` chance  (where N is the number of loaded BPF progams in the system) that you'll generate an ID that's already in use, and then start overwriting keys in the DB 🤔 
It might be better to use `program_tmp_` as the prefix just in case.

bpfman/src/command.rs
```diff
@@ -37,6 +37,7 @@ use crate::{
 };
 
 /// These constants define the key of SLED DB
+pub(crate) const PROGRAM_PREFIX: &str = "program_";
```
dave-tucker, line 40: Since you're changing the schema I'm guessing we'll need to delete the entire DB and re-create it - unless there are DB migrations for sled. We'll need to be mindful of this after we've done a release and perhaps we'll need to codify this somehow.

bpfman/src/command.rs
```diff
@@ -1520,5 +1498,5 @@ impl Program {
 // it.
 #[derive(Debug, Clone)]
 pub(crate) struct BpfMap {
-    pub(crate) used_by: Vec<u32>,
+    pub(crate) _used_by: Vec<u32>,
```
dave-tucker, line 1501: If this is the only field in the struct, and it's unused... does that mean the whole struct can be removed?

dave-tucker (changes requested):

bpfman/src/command.rs
astoycos: Good point.. 🤔  Let's use program_tmp_ for now

bpfman/src/command.rs
astoycos: Hrm yeah I'm thinking for right now we're alright... but once we release we'll need to have to have cleanup logic if the schema changes at all 🤔

bpfman/src/command.rs
astoycos: Good catch meant to remove this 👍

dave-tucker (approved):

PR #963, opened 2024-02-01
docs: Minor nits

Billy99:
Fix some minor nits in docs. Primarily, when running as a long running process, use `bpfman system service --timeout=0`. ﻿﻿Also add docker and kind to the list of packages that may need to be installed.

### DIFF
--- a/docs/getting-started/building-bpfman.md
+++ b/docs/getting-started/building-bpfman.md
@@ -241,6 +241,45 @@ sudo dnf install perl
 sudo apt install perl
 ```
 
+### Install docker
+
+To build the `bpfman-agent` and `bpfman-operator` using the provided Makefile and the
+`make build-images` command, `docker` needs to be installed.
+There are several existing guides:
+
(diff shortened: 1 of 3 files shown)


docs/getting-started/building-bpfman.md
```diff
@@ -241,6 +241,20 @@ sudo dnf install perl
 sudo apt install perl
 ```
 
+### Install docker
+
+To build the `bpfman-agent` and `bpfman-operator`, `docker` needs to be installed.
```
astoycos, line 246: I think we need to make sure builds work with podman as well just to be good net citizens :) I thought they did!

astoycos: For KIND we'll always need docker though

astoycos: But at the very least le't just say "container build tooling" or something like that

Billy99: So I agree that should make sure that builds work with podman, but `make build-images` specifically calls `docker build`. I just hit an issue on my server that took an hour to debug. I had `docker` installed but not `docker buildx`, so `make build-images` was failing. So we definitely have a dependency on docker.

astoycos (approved):
Looks good to me thanks @Billy99!

PR #962, opened 2024-02-01
bpfman: move bpfman.sock to standalone directory

Billy99:
On a redeploy of bpfman in a Kubernetes deployment (for example, edit the bpfman ConfigMap and change one of the log level values), bpfman fails to come back up. `/run/bpfman/sock/` is defined in the bpfman-deployment as an empty-dir, but on a restart, `/run/bpfman/` is remounted, but contains a non-empty `/run/bpfman/sock/` directory. This causes the container not to come up.

The fix is to rename `/run/bpfman/sock/bpfman.sock` to `/run/bpfman-sock/bpfman.sock`.

To reproduce:

```
kubectl edit configmap -n bpfman bpfman-config
apiVersion: v1
data:
  bpfman.agent.image: quay.io/bpfman/bpfman-agent:latest
  bpfman.agent.log.level: info
  bpfman.image: quay.io/bpfman/bpfman:latest
  bpfman.log.level: info -> debug
:

kubectl describe pod -n bpfman bpfman-daemon-sb95t
:
Events:
  Type     Reason     Age                From               Message
  ----     ------     ----               ----               -------
  Warning  Failed     23s (x3 over 36s)  kubelet            Error: failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/var/lib/kubelet/pods/102843fe-2d3d-42d5-8c2d-c258d846b81e/volumes/kubernetes.io~empty-dir/bpfman-sock" to rootfs at "/run/bpfman/sock": possibly malicious path detected -- refusing to operate on /run/containerd/io.containerd.runtime.v2.task/k8s.io/bpfman/rootfs/run/bpfman/sock (deleted): unknown
```

See Issue #915, specifically https://github.com/bpfman/bpfman/issues/915#issuecomment-1894472712

PR #964, opened 2024-02-02
bpfman: Make the listening socket configurable

astoycos:
Make the socket bpfman listen's on configurable. This isn't currently useful for the CLI since the socket can't be configured there, but for external users like the bpfman-agent it is.

Follow up to #962

### DIFF
--- a/bpfman/src/cli/args.rs
+++ b/bpfman/src/cli/args.rs
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Authors of bpfman
 
+use std::path::PathBuf;
+
 use bpfman_api::ProgramType;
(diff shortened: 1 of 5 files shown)


bpfman/src/cli/args.rs
```diff
@@ -332,6 +334,9 @@ pub(crate) struct ServiceArgs {
     /// Shutdown after N seconds of inactivity. Use 0 to disable.
     #[clap(long, default_value = "15")]
     pub(crate) timeout: u64,
+    #[clap(long, default_value = "/run/bpfman-sock/bpfman.sock")]
```
Billy99, line 337: Can we use the existing const string for "/run/bpfman-sock/bpfman.sock"?

dave-tucker (approved):
LGTM

Issue #914, opened 2024-01-12
Health and Metrics port numbers clash


maryamtahhan:
The issue is that in some scenarios port 8080 and 8081 are already in use by other services on the system. When this happens the bpfman-agent container fails to deploy

```
# kubectl describe pods -n bpfman bpfman-daemon-4cs8d

Events:
  Type     Reason     Age                From               Message
  ----     ------     ----               ----               -------
  Normal   Scheduled  21s                default-scheduler  Successfully assigned bpfman/bpfman-daemon-4cs8d to wsfd-advnetlab47.anl.lab.eng.bos.redhat.com
  Normal   Pulling    21s                kubelet            Pulling image "quay.io/fedora/fedora-minimal:latest"
  Normal   Pulled     20s                kubelet            Successfully pulled image "quay.io/fedora/fedora-minimal:latest" in 207ms (207ms including waiting)
  Normal   Created    20s                kubelet            Created container mount-bpffs
  Normal   Started    20s                kubelet            Started container mount-bpffs
  Normal   Pulling    19s                kubelet            Pulling image "quay.io/bpfman/bpfman:latest"
  Normal   Started    17s                kubelet            Started container bpfman
  Normal   Created    17s                kubelet            Created container bpfman
  Normal   Pulled     17s                kubelet            Successfully pulled image "quay.io/bpfman/bpfman:latest" in 1.908s (1.908s including waiting)
  Normal   Pulling    17s                kubelet            Pulling image "quay.io/bpfman/bpfman-agent:latest"
  Normal   Pulled     14s                kubelet            Successfully pulled image "quay.io/bpfman/bpfman-agent:latest" in 2.977s (2.977s including waiting)
  Normal   Pulling    14s                kubelet            Pulling image "quay.io/bpfman/csi-node-driver-registrar:v2.9.0"
  Normal   Pulled     14s                kubelet            Successfully pulled image "quay.io/bpfman/csi-node-driver-registrar:v2.9.0" in 246ms (246ms including waiting)
  Normal   Created    14s                kubelet            Created container node-driver-registrar
  Normal   Started    14s                kubelet            Started container node-driver-registrar
  Warning  BackOff    11s (x2 over 12s)  kubelet            Back-off restarting failed container bpfman-agent in pod bpfman-daemon-4cs8d_bpfman(b0e1ffb4-811e-4843-9c6a-5f050a39a8f3)
  Normal   Created    0s (x3 over 14s)   kubelet            Created container bpfman-agent
[...]

PR #965, opened 2024-02-02
operator: make health and metrics ports configurable

Billy99:
In some scenarios, the health and metric ports may are already in use by other services on the system. When this happens the bpfman-agent container fails to deploy. The current default is 8080 and 8081. First, move the port numbers to something less common. Then make the values configurable.
    
The ports are passed in through the yaml files. However, the bpfd-agent releated yaml files are built into the bpfman-operator image, so the values cannot be easily changed without rebuilding the image. So add fields to the ConfigMap so they can be overwritten without rebuilding the bpfman-operator image.

Resolves: #914


astoycos (approved):
LGTM

Thanks @Billy99 !!!

PR #976, opened 2024-02-05
build: make bundle failing on upstream main

Billy99:
Upstream main builds have been failing:

```
Run cd bpfman-operator && make bundle
:
go: downloading golang.org/x/text v0.5.0
/home/runner/work/bpfman/bpfman/bpfman-operator/bin/controller-gen crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
controllers/bpfman-agent/common.go:25:2: package slices is not in GOROOT (/opt/hostedtoolcache/go/1.20.13/x64/src/slices)
Error: not all generators ran successfully
run `controller-gen crd webhook paths=./... output:crd:artifacts:config=config/crd/bases -w` to see all available markers, or `controller-gen crd webhook paths=./... output:crd:artifacts:config=config/crd/bases -h` for usage
make: *** [Makefile:192: manifests] Error 1
```

The error mentions `/opt/hostedtoolcache/go/1.20.13/x64/src/slices` but I believe slices package wasn't introduced until go 1.21.

While working on the above issue, clippy errors started. This also includes a commit to pin the rust toolchain to nightly-2024-02-03 until the clippy issue gets fixed. Backing out this change is tracked by #977

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,7 +54,7 @@ jobs:
 
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly
+          toolchain: nightly-2024-02-03
           components: rustfmt, clippy, rust-src
           override: false
 
@@ -83,11 +83,11 @@ jobs:
 
       - name: Check formatting
         run: |
-          cargo +nightly fmt --all -- --check
+          cargo +nightly-2024-02-03 fmt --all -- --check
 
       - name: Run clippy
         run: |
-          cargo +nightly clippy --all -- --deny warnings
+          cargo +nightly-2024-02-03 clippy --all -- --deny warnings
 
       - name: Build
         run: cargo build --verbose
@@ -207,7 +207,7 @@ jobs:
 
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly
+          toolchain: nightly-2024-02-03
           components: rust-src
           override: true
 
--- a/.github/workflows/image-build.yaml
+++ b/.github/workflows/image-build.yaml
@@ -22,6 +22,7 @@ jobs:
       matrix:
         image:
           - registry: quay.io
+            build_language: rust
             repository: bpfman
             image: bpfman
             dockerfile: ./Containerfile.bpfman
@@ -35,6 +36,7 @@ jobs:
               type=raw,value=latest,enable={{is_default_branch}}
 
           - registry: quay.io
+            build_language: go
             repository: bpfman
             image: bpfman-agent
             dockerfile: ./bpfman-operator/Containerfile.bpfman-agent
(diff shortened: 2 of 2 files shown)


astoycos:
Looks like the `prost-derive` crate is giving some issues?

Billy99:
> Looks like the `prost-derive` crate is giving some issues?

Actually, `curve25519-dalek`:
https://github.com/dalek-cryptography/curve25519-dalek/issues/618

.github/workflows/build.yml
```diff
@@ -54,7 +54,7 @@ jobs:
 
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly
+          toolchain: nightly-2024-02-03
```
astoycos, line 57: Do we want this pinned to a specific version rather than the latest nightly

.github/workflows/image-build.yaml
```diff
@@ -61,6 +64,8 @@ jobs:
               type=raw,value=latest,enable={{is_default_branch}}
 
           - registry: quay.io
+            build_language: go
+            bpf_build_wrapper: go
```
astoycos, line 68: This only needs the `build_language` I think

.github/workflows/image-build.yaml
```diff
@@ -356,20 +378,20 @@ jobs:
...
-        if: ${{ matrix.image.bpf_build_wrapper == 'go' }}
+        if: ${{ matrix.image.build_language == 'go' }}
         with:
           go-version: "1.21"
 
       - uses: sigstore/cosign-installer@v3.3.0
 
       - uses: actions/checkout@v4
-        if: ${{ matrix.image.bpf_build_wrapper == 'rust' }}
+        if: ${{ matrix.image.build_language == 'rust' }}
```
astoycos, line 388: Wait we only need libbpf if there's bpf code being built! 

I think we want to leave this as bpf_build_wrapper

.github/workflows/build.yml
Billy99: Yes, until the issue gets fixed. I'll write a bpfman issue to track the curve25519-dalek issue and unwind once it gets fixed.

astoycos: Sounds good thanks!

.github/workflows/image-build.yaml
Billy99: You are right. That was how I had the original patch, but the `bpf_build_wrapper` line sneaked in when I was making sure I didn't introduce the clippy errors. I'll remove it.

.github/workflows/image-build.yaml
astoycos: And I'm pretty sure we only need it for rust/aya based bpf builds 🤔

.github/workflows/image-build.yaml
astoycos: Sounds good thanks!

.github/workflows/image-build.yaml
Billy99: Good ctach. I thought I saw it elsewhere, but just re-looked and only needed for BPF builds.

astoycos:
LGTM

thanks @Billy99 !

PR #875, opened 2023-12-16
Kubernetes Support for attaching uprobes in containers

anfredette:
Add Kubernetes support for attaching uprobes in containers

Fixes: #724

PR #978, opened 2024-02-06
fixups for pr #875

anfredette:
bpfman-operator fixups to pr #875
 
Address comments from astoycos plus a few additional cleanups.

This doesn't contain any functional changes from pr #875, but rather optimizations, better organization, and general cleanup.


bpfman-operator/apis/v1alpha1/uprobeProgram_types.go
```diff
@@ -77,7 +77,7 @@ type UprobeProgramSpec struct {
 	Pid int32 `json:"pid"`
 
 	// Containers identifes the set of containers in which to attach the uprobe.
-	// If Containers are not specified, the uprobe will be attached in the
+	// If Containers is not specified, the uprobe will be attached in the
 	// bpfman-agent container.  The ContainerSelector is very flexible and even
```
anfredette, line 81: Andrew:
I think being able to easily specify "all containers in a cluster" Is a footgun we shouldn't expose

Andre:
I started out making it more opinionated to avoid the footguns.  However, I found this approach less user friendly.  There are times when you might want to include (or not) each of the parameters.  In the end I came to the opinion that this is pretty low level functionality, and you should know what you're doing to use it.  So, I'd rather not make it harder to use by arbitrarily limiting how you can choose the containers.

bpfman-operator/config/bpfman-deployment/daemonset.yaml
```diff
@@ -122,10 +122,6 @@ spec:
           volumeMounts:
             - name: bpfman-sock
               mountPath: /run/bpfman-sock
```
anfredette, line 124: Andrew: Why are we mounting the bpfman config in the agent?

Andre: Good catch.  I think that was a mistake during a rebase conflict fixup.  Billy had removed it as part of a recent pr, but I guess I accidentally added it back.  Removed.

bpfman-operator/controllers/bpfman-agent/common.go
```diff
@@ -447,8 +436,75 @@ func reconcileProgram(ctx context.Context,
...
+}
+
+// reconcileProgram is called by ALL *Program controllers, and contains much of
+// the core logic for taking *Program objects, turning them into bpfProgram
+// object(s), and ultimately telling the custom controller types to load real
+// bpf programs on the node via bpfman. Additionally it acts as a central point for
+// interacting with the K8s API. This function will exit if any action is taken
+// against the K8s API. If the function returns a retry boolean and error, the
+// reconcile will be retried based on a default 5 second interval if the retry
+// boolean is set to `true`.
```
anfredette, line 451: @astoycos, you made some good comments on the function and I it to try to address them.  As part of this, I broke out the handling of the case were we are deleting programs and the case where we are not.  I also reused the delete function to delete unexpected programs.  I'm sure it can be broken into more functional components, but this is a start.

bpfman-operator/controllers/bpfman-agent/common.go
```diff
@@ -533,6 +589,7 @@ func ProcessMapOwnerParam(ctx context.Context,
 	}
 }
 
+// get Clientset returns a kubernetes clientset.
```
anfredette, line 592: Andrew:
We can't rely on this always being set there's a different process for generating a k8s clientset for in-cluster vs out of cluster see https://github.com/bpfman/bpfman/blob/main/bpfman-operator/pkg/helpers/helpers.go#L109

I think we can actually just reuse getk8sConfigOrDie here

Andre:
I tried using GetClientOrDie(), but wasn't able to get it working.  My code is using a *kubernetes.Clientset, but GetClientOrDie() returns a *bpfmanclientset.Clientset.  Maybe there's a way to convert between the two or use *bpfmanclientset.Clientset to get the pod list with the filters needed, but I'll need to work on it more.  The other thought is to maybe use the pattern in getk8sConfigOrDie() to handle the in-cluster vs out of cluster case but get a *kubernetes.Clientset.  I'll need to investigate more.

bpfman-operator/controllers/bpfman-agent/common.go
```diff
@@ -109,6 +106,29 @@ func nodePredicate(nodeName string) predicate.Funcs {
 	}
 }
 
+// Predicate to watch for Pod events on a specific node which checks if the
+// event's Pod is scheduled on the given node.
+func podOnNodePredicate(nodeName string) predicate.Funcs {
```
anfredette, line 111: Andrew:
Why not just mimic what was done here https://github.com/bpfman/bpfman/blob/main/bpfman-operator/controllers/bpfman-agent/common.go#L87 Instead of making a new struct?

Andre:
I followed a different pattern which works fine too.  I don't know which is preferable, but I changed it to use the pattern we already have in nodePredicate.

bpfman/src/bpf.rs
```diff
@@ -616,22 +616,21 @@ impl BpfManager {
                             "bpfman-ns"
                         };
```
anfredette, line 617: Andrew:
I really don't think we should be bringing k8s specific things into core bpfman... Maybe we make this path configurable?

Andre:
I agree in general, but it's more user friendly if you don't have to configure it.

bpfman/src/utils.rs
```diff
@@ -185,3 +185,25 @@ pub(crate) fn sled_insert(db_tree: &Tree, key: &str, value: &[u8]) -> Result<(),
         )
     })
 }
+
+// Helper function to get the error message from stderr
```
anfredette, line 189: Andrew:
Made multiple comments about logging errors and then bailing with basically the same error.

Andre:
I agree.  This is not ideal.  I was doing it because the error wasn't getting back to the calling program.  However, I did see that stderr was being returned, which contained the error message, so I implemented this function to pull the error out so it could be logged by the calling function.  I'm not sure if this is the best way to do this, but I'm open to other ideas.

anfredette:
@astoycos, Thanks for your detailed review!  I think I addressed your comments one way or another in this pr.  In cases where I didn't, or where I thought my change was questionable, I added your comment to this pr and responded to it rather than trying to respond in the old pr.

bpfman-operator/apis/v1alpha1/uprobeProgram_types.go
astoycos: That's a fair enough argument for me! Let's just make sure to have documentation/ comments that say as much.

bpfman-operator/config/bpfman-deployment/daemonset.yaml
astoycos: Thanks!

bpfman-operator/controllers/bpfman-agent/common.go
astoycos: Ah ok so yeah that code in helpers is for our API specifically, however we still need to follow the same logic for generating the raw K8s client i.e if we're in-cluster see https://github.com/kubernetes/client-go/tree/master/examples/in-cluster-client-configuration for a good example

bpfman-operator/controllers/bpfman-agent/common.go
astoycos: Looks good to me, thanks for keeping it consistent!

bpfman/src/bpf.rs
```diff
@@ -616,22 +616,21 @@ impl BpfManager {
...
-                        match status {
-                            Ok(s) => {
-                                if !s.success() {
-                                    debug!("bpfman-ns returned non-successful result: {:?}", s);
+                            .output();
+
+                        match output {
+                            Ok(o) => {
+                                if !o.status.success() {
+                                    debug!(
```
astoycos, line 626: Shouldn't this just be `info!`? If there's an error we should always know about it

bpfman/src/bpf.rs
astoycos: I mean yes but for a true separation of concerns bpfman as the core process shouldn't be aware of K8s unless it really needs to be, that's the whole point of the agent IMO

Anyways It's not a hill I'm willing to die on at the moment.

bpfman/src/utils.rs
astoycos: This looks good and is what I was thinking 👍

bpfman-operator/controllers/bpfman-agent/common.go
```diff
@@ -109,6 +106,29 @@ func nodePredicate(nodeName string) predicate.Funcs {
...
+	return predicate.Funcs{
+		GenericFunc: func(e event.GenericEvent) bool {
+			pod, ok := e.Object.(*v1.Pod)
+			return ok && pod.Spec.NodeName == nodeName
+		},
+		CreateFunc: func(e event.CreateEvent) bool {
+			pod, ok := e.Object.(*v1.Pod)
+			return ok && pod.Spec.NodeName == nodeName && pod.Status.Phase != v1.PodPending
```
astoycos, line 119: hrmm will this always work? Just because the pod isn't pending doesn't mean it will be running right? Are we trying to say, "return true if the pod is located on my node AND wait to make sure the it's running without entering Error or CrashLoopBackoff"

bpfman-operator/controllers/bpfman-agent/common.go
astoycos: +1 One nit is that `handleProgramCreate` is also handling updates right? Maybe change it to just `handleProgram`?

bpfman-operator/controllers/bpfman-agent/common.go
```diff
@@ -533,6 +589,7 @@ func ProcessMapOwnerParam(ctx context.Context,
 	}
 }
 
+// get Clientset returns a kubernetes clientset.
 func getClientset() (*kubernetes.Clientset, error) {
 	kubeconfig := os.Getenv("KUBECONFIG")
```
astoycos, line 594: Yeah this still isn't the right way to do this when creating a clientset inside a k8s cluster, see my other comment!

astoycos (changes requested):
Pretty much there thanks for going back and doing this!! I still want to consolidate and spend more time analyzing the core reconciler code but this is a start 👍

bpfman-operator/controllers/bpfman-agent/common.go
anfredette: It returns true if the pod is located on my node AND the pod is not in the pending phase.

Every time a pod is started, it spends some time in the pending phase before moving to the PodRunning phase. While in this state, crictl doesn't know anything about the pod, so the calls fail.  By ignoring events for pods in this state, we avoid these errors.  I have wondered about other states, though.  We could filter some of these out too, but I thought it was safer to allow the reconcile to proceed in these cases unless we know for sure it's a problem. I think I can test the PodFailed and PodSucceeded phases.

astoycos: Just to be thorough can you make it 

!= PodPending, != PodFailed, != PodUnknown

see https://github.com/kubernetes/api/blob/v0.29.1/core/v1/types.go#L2966

bpfman-operator/controllers/bpfman-agent/common.go
anfredette: Naming is hard.  I thought about this a bit when I was doing it, but couldn't left it as handleProgramCreate.  It's really handleProgCreateAndUpdate, or handleProgNotBeingDeleted.  I'll think about it.

bpfman-operator/controllers/bpfman-agent/common.go
anfredette: I'm thinking it should be: `return ok && pod.Spec.NodeName == nodeName && pod.Status.Phase == v1.PodRunning`
I don't think we need to be concerned about `PodSucceeded` either because it means that "all containers in the pod have voluntarily terminated".  So, there's nothing to install an eBPF program in.  That just leaves `PodRunning`.

astoycos: Thats fine with me as well 👍

anfredette: Done.

anfredette:
This should be ready.

astoycos (approved):
LGTM

Thanks @anfredette!!!

PR #980, opened 2024-02-07
bpfman: add retries to sled access

Billy99:
To support removing RPC from the bpfman CLI, add a retry to the sled DB access so whien multiple simultaneous CLI calls are made, the additional calls won't panic immediately but will retry after a delay.

### DIFF
--- a/bpfman/src/main.rs
+++ b/bpfman/src/main.rs
@@ -1,9 +1,12 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Authors of bpfman
+use std::{thread, time::Duration};
+
+use bpfman_api::config::Config as BpfmanConfig;
 #[cfg(not(test))]
 use bpfman_api::util::directories::STDIR_DB;
 use clap::Parser;
-use lazy_static::lazy_static;
+use once_cell::sync::OnceCell;
 use sled::{Config, Db};
 
 mod bpf;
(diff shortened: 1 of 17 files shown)


Billy99:
Example:

Terminal 1 - First call into DB:

```
$ sudo RUST_LOG=info /home/bmcfall/src/bpfman/target/debug/bpfman system service --timeout=0
[INFO  bpfman::cli::system] Log using env_logger
[INFO  bpfman::cli::system] Has CAP_BPF: true
[INFO  bpfman::cli::system] Has CAP_SYS_ADMIN: true
[INFO  bpfman::serve] Using no inactivity timer
[INFO  bpfman::serve] Using default Unix socket
[INFO  bpfman::serve] Listening on /run/bpfman-sock/bpfman.sock
[INFO  bpfman] BILLY: db opened.                                                 <-- HAS DB LOCK
[INFO  bpfman::oci_utils::cosign] Starting Cosign Verifier, downloading data from Sigstore TUF repository
```

Terminal 2 - Second call into DB:

```
$ sudo RUST_LOG=info /home/bmcfall/src/bpfman/target/debug/bpfman system service --timeout=0
[INFO  bpfman::cli::system] Log using env_logger
[INFO  bpfman::cli::system] Has CAP_BPF: true
[INFO  bpfman::cli::system] Has CAP_SYS_ADMIN: true
[INFO  bpfman::serve] Using no inactivity timer
[INFO  bpfman::serve] Using default Unix socket
[INFO  bpfman::serve] Listening on /run/bpfman-sock/bpfman.sock
[INFO  bpfman] BILLY: db locked 1                                                <-- RETRIES
[INFO  bpfman] BILLY: db locked 2
[INFO  bpfman] BILLY: db locked 3
thread 'main' panicked at bpfman/src/main.rs:40:9:
Unable to open temporary root database
```

bpfman/src/main.rs
```diff
@@ -23,10 +26,18 @@ const BPFMAN_ENV_LOG_LEVEL: &str = "RUST_LOG";
 
 #[cfg(not(test))]
 lazy_static! {
-    pub static ref ROOT_DB: Db = Config::default()
-        .path(STDIR_DB)
-        .open()
-        .expect("Unable to open root database");
+    pub static ref ROOT_DB: Db = {
```
astoycos, line 29: This looks like a super simple fix and I think it will work, however I think we'd want to be able to easily configure the maximum number of retries and duration between retries

astoycos: It's definitely going to be a bit annoying to tune those parameters in some scenarios

PR #982, opened 2024-02-07
ci: Group dependabot updates

dave-tucker:
This should reduce PR churn and also solve for issues when tonic/prost and other crates need to be simultaneously updated.


Billy99 (approved):
LGTM

PR #984, opened 2024-02-08
chore: Fix yamllint/vscode-yaml formatting discrepencies

dave-tucker:
Tell yamllint to prefer double quotes + nag on unnecessary quotes. Tell prettier (used in vscode-yaml) to ignore the golang versions, that are in single quotes due to "1.20" being shortened to 1.2, which has strange results.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,21 +3,21 @@
 
 version: 2
 updates:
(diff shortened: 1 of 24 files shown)


astoycos (approved):
Thanks @dave-tucker!

astoycos (approved):
LGTM

PR #989, opened 2024-02-12
remove to-string clippy failures

astoycos:
Implement std::fmt::Display rather than to_string
for Location and XDPMode to fix clippy failures.


anfredette (approved):
lgtm

PR #990, opened 2024-02-12
fix startup race

astoycos:
Make sure bpfman state is rebuilt before starting
the unix server, since data is intially written
in rpc.rs it could race with the rebuilding state
in bpfmanager::rebuild_state().

In this particular scenario, when running with
systemd, a `ProgramData` temporary pre-load
database tree would get created before the rebuild_state() was called, and then the rebuild logic would
delete the `ProgramData` tree mid load causing
all of the loading errors.


Billy99 (approved):
LGTM

PR #1000, opened 2024-02-20
bpfman: Cache TUF Metadata

dave-tucker:
This writes the TUF metadata to disk instead of caching in-memory. This should allow for multiple invocations of bpfman to be faster.


astoycos (approved):
LGTM

Thanks @dave-tucker

PR #1003, opened 2024-02-27
docs: Use the include-markdown plugin

dave-tucker:
This plugin includes markdown files and by default rewrites relative links. This allows us to preserve the governance docs in the repository root (and for them to use relative links) while also allowing those same docs to appear on the website.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -191,7 +191,7 @@ cargo +nightly clippy --all -- --deny warnings
 * Verify that Yaml files have been formatted (see
   [Install Yaml Formatter](https://bpfman.io/main/getting-started/building-bpfman/#install-yaml-formatter))
 * Verify that unit tests are passing locally (see
-  [Unit Testing](https://bpfman.io/main/developer-guide/testing/#unit-testing)):
+  [Unit Testing](./docs/developer-guide/testing.md#unit-testing)):
 
 ```console
 cd src/bpfman/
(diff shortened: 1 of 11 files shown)


anfredette (changes requested):
The include-markdown plugin seems to work fine, but I think this has the same problem as the previous method of including the files.  For, example, using your pr, I added a relative link in CONTRIBUTING.md.  I got the following error, and the link in the webpage doesn't work:
```
nfvsdn-04:~/go/src/github.com/bpfman/bpfman$ mkdocs build
INFO    -  Cleaning site directory
INFO    -  Building documentation to directory: /home/afredette/go/src/github.com/bpfman/bpfman/site
WARNING -  Doc file 'governance/CONTRIBUTING.md' contains a relative link './docs/developer-guide/testing.md#unit-testing',
           but the target 'governance/docs/developer-guide/testing.md' is not found among documentation files.
INFO    -  Documentation built in 2.53 seconds
```

anfredette:
One thing that would work (if it was possible) is if the file could include a file that contained the link definitions, and it could be rendered properly by both mkdocs and github.  This way, for example, we could have one set of links in bpfman/CONTRIBUTING_LINKS.md, and another set in bpfman/docs/governance/CONTRIBUTING_LINKS.md.

Even better -- why can't mkdocs just process .md files outside of the docs directory if you give an explicit link?

PR #1004, opened 2024-02-28
bpfman-agent: list filters not working

Billy99:
The XDP map-sharing example stopped working. The main XDP program loads fine but the program using the main XDP programs maps was not loading. After some investigation, the request to retrieve a list of programs filtered by a given UUID was returning multiple programs, not just the program that matched the UUID filter. All the non-bpfman loaded programs were being returned as well as the bpfman loaded program that matched the filter.


astoycos (approved):
Thanks @Billy99!

PR #1011, opened 2024-03-01
Image manager refactor

astoycos:
Refactor the image manager
- Remove all channels and commands
- only initialize image_manager when needed and do this in the core codepath (bpf.rs)
- Cleanup a bunch of extra async functions which don't need to be async any longer

### DIFF
--- a/bpfman/src/bpf.rs
+++ b/bpfman/src/bpf.rs
@@ -3,6 +3,7 @@
 
 use std::{
     collections::HashMap,
+    fs::{create_dir_all, remove_dir_all},
     path::{Path, PathBuf},
 };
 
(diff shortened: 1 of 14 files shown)


bpfman/src/utils.rs
```diff
@@ -20,16 +21,14 @@ use crate::errors::BpfmanError;
 pub(crate) const SOCK_MODE: u32 = 0o0660;
 
 // Like tokio::fs::read, but with O_NOCTTY set
-pub(crate) async fn read<P: AsRef<Path>>(path: P) -> Result<Vec<u8>, BpfmanError> {
+pub(crate) fn read<P: AsRef<Path>>(path: P) -> Result<Vec<u8>, BpfmanError> {
     let mut data = vec![];
-    tokio::fs::OpenOptions::new()
+    std::fs::OpenOptions::new()
```
dave-tucker, line 34: No need to fully qualify this. May as well `use std::fs::OpenOptions;`

bpfman/src/utils.rs
```diff
@@ -47,22 +46,22 @@ pub(crate) fn get_ifindex(iface: &str) -> Result<u32, BpfmanError> {
     }
 }
 
-pub(crate) async fn set_file_permissions(path: &Path, mode: u32) {
+pub(crate) fn set_file_permissions(path: &Path, mode: u32) {
     // Set the permissions on the file based on input
-    if (tokio::fs::set_permissions(path, std::fs::Permissions::from_mode(mode)).await).is_err() {
+    if (std::fs::set_permissions(path, std::fs::Permissions::from_mode(mode))).is_err() {
```
dave-tucker, line 59: As above re: qualification. Just import instead.

bpfman/src/bpf.rs
```diff
@@ -54,22 +50,31 @@ const MAPS_USED_BY_PREFIX: &str = "map_used_by_";
...
         }
     }
 
+    // Make sure to call init_image_manger if the command requires interation with
+    // an OCI based container registry. It should ONLY be used where needed, to
+    // explicitly control when bpfman blocks for network calls to both sigstore's
+    // cosign tuf registries and container registries.
+    pub(crate) async fn init_image_manager(&mut self) {
+        let config = open_config_file();
+        self.image_manager = Some(
```
dave-tucker, line 71: This should only happen if `self.image_manager.is_none()`.
I know you've put comments here, but humans make mistakes.

dave-tucker (changes requested):
A couple of nits but overall LGTM.

bpfman/src/utils.rs
astoycos: Done

bpfman/src/bpf.rs
astoycos: Fair enough, done!

dave-tucker (approved):

dave-tucker:
@astoycos needs a rebase!

PR #1014, opened 2024-03-04
Turn bpfman into a library crate

### LINKED ISSUE #1009: CLI: Remap types passed to CLI print
Originally, CLI code sent an RPC request to bpfman. For Get/List/Create commands, bpfman returns a `command::Program` (see [command.rs](https://github.com/bpfman/bpfman/blob/main/bpfman/src/command.rs#L201)), which then gets mapped into the API structure used by RPC, `bpfman_api::v1::ProgramInfo` and `bpfman_api::v1::KernelProgramInfo` (see [bpfman.v1.rs](https://github.com/bpfman/bpfman/blob/main/bpfman-api/src/bpfman.v1.rs#L35)) and passed to `ProgTable`(see [table.rs](https://github.com/bpfman/bpfman/blob/main/bpfman/src/cli/table.rs#L19)) to be printed.

`ProgTable` can be rewritten to take `command::Program` instead, skipping the mapping from `command::Program` to `bpfman_api::v1::ProgramInfo` and `bpfman_api::v1::KernelProgramInfo`.

astoycos:
This commit involves a bunch of re-organization.
It mostly moves the cli portion of the codebase
to be it's own binary, and turns the core
functionality into a library both of which
live within the bpfman package.

This mainly involved making a bunch of functions
public and moving bits around.

~~This will need to be rebased on #1011~~
This is rebased on #1011  

- Move the GRPC server out into it's own binary that
simply calls into the core bpfman library. This
allows us to fully stop acting like a long
running daemon even for the kubernetes and non-rust
language use cases. This binary lives in the
existing bpfman-api crate, making this work required
removing a bunch of circular dependencies where
there were internal types defined in bpfman-api
that were being used in bpfman.

- remove some unused deps.

- fix log levels in k8s integration tests Make sure
we are logging at the debug level in both
the daemon and operator when running our kubernetes
integration tests.

- Now that bpfman is a library and the rpc server
has been extracted we need to make sure
that the sled database lock is released whenever
it's not being used so other processes can call the
library on the same host. This wan't possible with
the root database reference as a lazy static variable
since once it was initialized by a process it would
never be dropped until the calling process exited.
This is an issue when the calling process is our
long lived rpc shim. To fix this simply add the
root database instance as a member of the BpfManager
and then pass it via arguments where ever it's needed.
Additionally fixup the retry mechanism for acquiring
the lock on the filesystem backed sled db instance
to default accordingly and tweak it to retry a bit
more when running in kubernetes. 

- [x] Make bpfman a consumable library crate
- [x] Remove RPC server from the bpfman library and put it in it's own binary (now in the bpfman-api crate)
- [x] Make the root_db instance not a lazy static variable
- [x] Fix #1009


bpfman/Cargo.toml
```diff
@@ -6,9 +6,13 @@ name = "bpfman"
 repository.workspace = true
 version.workspace = true
 
+[lib]
+name = "bpfman"
+path = "src/lib/bpf.rs"
```
dave-tucker, line 11: This should be `src/lib.rs`

bpfman/Cargo.toml
```diff
@@ -6,9 +6,13 @@ name = "bpfman"
...
+
 [[bin]]
 name = "bpfman"
-path = "src/main.rs"
+path = "src/cli/main.rs"
```
dave-tucker, line 15: By convention, this is usually `src/bin/bpfman.rs` or `src/bin/main.rs`

bpfman-rpc/Cargo.toml
```diff
@@ -0,0 +1,77 @@
+[package]
+description = "An RPC server shim for the bpfman library"
```
dave-tucker, line 2:
```suggestion
description = "A gRPC API for bpfman"
```

bpfman-rpc/Cargo.toml
```diff
@@ -0,0 +1,77 @@
+[package]
```
dave-tucker, line 1: I'm not convinced we need to add yet-another-crate here.
Given the only consumer of `bpfman-api` is going to be `bpfman-rpc`... why not just add `bpfman-api/src/bin/main.rs`?

dave-tucker (changes requested):
Just a few minor things relating to code layout to fix before this is good to go.

bpfman/Cargo.toml
astoycos: Hrm based on the convention here (https://doc.rust-lang.org/cargo/guide/project-layout.html) I'd like to make it /src/bin/cli/main.rs since this is a "multi-file-executable)

bpfman-rpc/Cargo.toml
astoycos: Sounds good to me!

bpfman/Cargo.toml
astoycos: Additionally `/src/bin/main.rs` `/src/bin/used_by_main.rs` just won't work

bpfman-rpc/Cargo.toml
astoycos: uhh cyclic dependency hell, trying to clean this up

astoycos:
I added a hold to make sure I know when this is going to merge 🤣

bpfman/src/serve.rs
```diff
@@ -8,49 +8,41 @@ use std::{
...
-    let shutdown_rx4 = shutdown_tx.subscribe();
     let shutdown_handle = tokio::spawn(shutdown_handler(timeout, shutdown_tx));
 
-    let (tx, rx) = mpsc::channel(32);
-
-    let loader = BpfmanLoader::new(tx.clone());
+    let loader = BpfmanLoader::new(config);
     let service = BpfmanServer::new(loader);
 
+    //let mut bpf_manager = BpfManager::new(config.clone());
```
astoycos, line 43: nit remove

bpfman/src/storage.rs
```diff
@@ -45,7 +44,7 @@ const OPERATOR_PROGRAM_KEY: &str = "bpfman.io/ProgramName";
 const NPV_NOT_FOUND: i32 = 5;
 const OWNER_READ_WRITE: u32 = 0o0750;
 
-pub(crate) struct StorageManager {
+pub struct StorageManager {
```
astoycos, line 47: pub?

dave-tucker (approved):

astoycos:
Pushed some small CLI fixes around defaulting Proceed on and making sure program type was displaying as a string not an int.

astoycos:
Should be good to go!

PR #1017, opened 2024-03-05
Add kprobe and uprobe example programs

anfredette:
Added example programs for kprobes and uprobes along with support for deploying them in a Kubernetes cluster.  To provide a reliable target for uprobes in Kubernetes, this pr also includes a simple program called go-target that can be deployed in a pod on Kubernetes.

I also refactored the Makefile so that the instructions that were repeated for each program are shared by common deploy_prog and undeploy_prog targets.

Fixes: #1020 #1021


Billy99 (commented):
Like we discussed in the meeting yesterday, you don't need kprobe/uprobe in `examples/config/v0.2.0/`, `examples/config/v0.3.0/` or `examples/config/v0.3.1/`. However, you should probably create a `examples/config/v0.4.0/` and copy what you did there.

examples/go-kprobe-counter/bpf/kprobe_counter.c
```diff
@@ -0,0 +1,37 @@
...
+struct datarec {
+  __u64 counter;
+} datarec;
+
+struct {
+  __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+  __type(key, __u32);
+  __type(value, datarec);
+  __uint(max_entries, 1);
+  __uint(pinning, LIBBPF_PIN_BY_NAME);
```
astoycos, line 20: Hrm I need to revisit the current requirement we have for specifying pin by name, I don't want our users to have to do this.

astoycos: i.e https://github.com/aya-rs/aya/issues/837

examples/Makefile
```diff
@@ -155,116 +166,151 @@ IMAGE_XDP_BC ?= quay.io/bpfman-bytecode/go-xdp-counter:latest
...
-	$(KUSTOMIZE) build config/$(KUST_DIR)/go-tc-counter | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
+	$(KUSTOMIZE) build config/$(KUST_DIR)/$(CONFIG_DIR) | kubectl delete --ignore-not-found=$(ignore-not-found) -f -; \
+
+
+.PHONY: deploy-tc
+deploy-tc: PROG_NAME=go-tc-counter
+deploy-tc: CONFIG_DIR=$(PROG_NAME)
+deploy-tc: IMAGE_BC=$(IMAGE_TC_BC)
+deploy-tc: IMAGE_US=$(IMAGE_TC_US)
+deploy-tc: deploy-prog ## Deploy go-tc-counter to the cluster specified in ~/.kube/config.
```
astoycos, line 212: Don't duplicate this target just put the env vars at the top of the makefile

OR just make it one

deploy-tc:
    PROG_NAME=go-tc-counter
    CONFIG_DIR=prog_name

examples/Makefile
```diff
@@ -155,116 +166,151 @@ IMAGE_XDP_BC ?= quay.io/bpfman-bytecode/go-xdp-counter:latest
...
 .PHONY: deploy-tracepoint
-deploy-tracepoint: kustomize ## Deploy go-tracepoint-counter to the cluster specified in ~/.kube/config.
-ifndef TAG
-	sed 's@URL_BC@$(IMAGE_TP_BC)@' config/default/go-tracepoint-counter/patch.yaml.env > config/default/go-tracepoint-counter/patch.yaml
-	cd config/default/go-tracepoint-counter && $(KUSTOMIZE) edit set image quay.io/bpfman-userspace/go-tracepoint-counter=${IMAGE_TP_US}
-else
-	$(eval KUST_DIR=$(TAG))
-endif
-	$(KUSTOMIZE) build config/$(KUST_DIR)/go-tracepoint-counter | kubectl apply -f -
+deploy-tracepoint: PROG_NAME=go-tracepoint-counter
```
astoycos, line 223: same as above

astoycos: And please fix everywhere

examples/config/base/go-kprobe-counter/bytecode.yaml
```diff
@@ -15,4 +15,4 @@ spec:
   bytecode:
     image:
       url: quay.io/bpfman-bytecode/go-kprobe-counter:latest
-      imagepullpolicy: Always
+      imagepullpolicy: IfNotPresent
```
astoycos, line 18: Why the change here? For CI at least Always doesn't seem like a bad thing

examples/go-target/main.go
```diff
@@ -0,0 +1,26 @@
+// This is the target program that we will attach the uprobe to. It is a simple
+// program that runs the `ls` command every 5 seconds and logs the output.
```
astoycos, line 2: This doesn't seem to be running ls anymore

astoycos (commented):
Some small things biggest one was the makefile, otherwise looks good, thanks @anfredette!

examples/Makefile
anfredette: Unless I'm misunderstanding, I don't think what you are suggesting will work.  I tried it.  I hate Makefiles :-).

To separate the instructions from the variables, I needed to use [target-specific variables](https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html).  This allows me to define a separate set of variables for each target.  In theory, we could use the same set of variables for the deploy and undeploy versions for a given program, but then we'd need a separate variable to be passed in to describe the action.  E.g., make `tc-prog ACTION=DEPLOY`.  I think this is more clumsy for the user.

anfredette: Also, the multiple lines prefixed by `deploy-tc:` aren't duplicating the target.  That's how you defined target-specific variables in a Makefile.  It tells `make` that these variables only apply to the given target.

examples/config/base/go-kprobe-counter/bytecode.yaml
anfredette: Good question.  I changed it because it's not specified on the other existing programs, and I thought that the default was IfNotPresent.  IfNotPresent is also faster.  However, I just checked and the default is actually [it depends on several factors](https://kubernetes.io/docs/concepts/containers/images/).  So, I think we should explicitly specify it the same way for all the program types, and I agree with IfNotPresent so you always get the latest image with the given tag.

examples/go-kprobe-counter/bpf/kprobe_counter.c
anfredette: I copied this from the tracepoint example.  What is the preferred way to pin?

examples/go-target/main.go
anfredette: Good catch.  I started with ls, but changed it to call a local function.  The time also changed.  I'll correct the comment.

anfredette:
@astoycos thanks for the review!  Let me know what you think about the Makefile changes and if you have a better way to do it.  It was really irritating me that we had a whole bunch of instructions repeated for every target, and I was going to have to add several more lines to every target to check whether a given TAG was supported for that target.  I think it's a lot cleaner this way and a lot easier to add new program types.

astoycos (approved):
LGTM 

thanks @anfredette!!!

PR #1028, opened 2024-03-12
feat(bpf-log-exporter): Initial Commit

dave-tucker:
Add the bpf-log-exporter which uses netlink to communicate with the audit subsytem and stream eBPF events to stdout. Eventually, this will operate similar to bpf-metric-exporter and instead forward these events as logs to OpenTelemetry.


bpf-log-exporter/src/main.rs
```diff
@@ -0,0 +1,198 @@
...
+        let is_hex = Regex::new(r"^[0-9A-F]+$").unwrap();
+        let mut log_paylod = LogMessage::new();
+        loop {
+            let n = socket.recv(&mut &mut receive_buffer[..], 0)?;
+            let bytes = &receive_buffer[..n];
+            let rx_packet = <NetlinkMessage<AuditMessage>>::deserialize(bytes).unwrap();
+            match rx_packet.payload {
+                NetlinkPayload::InnerMessage(AuditMessage::Event((event_id, message))) => {
+                    match event_id {
+                        1300 => {
```
astoycos, line 81: Where do these come from? should the event id's be constants?

astoycos (approved):

Issue #636, opened 2023-08-03
Implement k8s integration tests for probes


anfredette:

astoycos:
(Uprobes AND Kprobes)

PR #1041, opened 2024-03-21
Add kprobe and uprobe k8s integration tests

anfredette:
Add kubernetes integration tests for uprobes and kprobes.

Fixes: #636

### DIFF
--- a/bpfman-operator/test/integration/kprobe_test.go
+++ b/bpfman-operator/test/integration/kprobe_test.go
@@ -1,0 +1,70 @@
+//go:build integration_tests
+// +build integration_tests
+
+package integration
+
+import (
+	"bytes"
+	"context"
+	"io"
+	"regexp"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/kong/kubernetes-testing-framework/pkg/clusters"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	kprobeGoCounterKustomize       = "../../../examples/config/default/go-kprobe-counter"
+	kprobeGoCounterUserspaceNs     = "go-kprobe-counter"
+	kprobeGoCounterUserspaceDsName = "go-kprobe-counter-ds"
+)
+
+func TestKprobeGoCounter(t *testing.T) {
+	t.Log("deploying kprobe counter program")
+	require.NoError(t, clusters.KustomizeDeployForCluster(ctx, env.Cluster(), kprobeGoCounterKustomize))
+	addCleanup(func(context.Context) error {
+		cleanupLog("cleaning up kprobe counter program")
+		return clusters.KustomizeDeleteForCluster(ctx, env.Cluster(), kprobeGoCounterKustomize)
+	})
+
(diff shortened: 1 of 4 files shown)


anfredette:
I'm leaving this as draft because I can't test it yet due to the sigstore issue.

anfredette:
This should be good to go now.

bpfman-operator/test/integration/kprobe_test.go
```diff
@@ -0,0 +1,68 @@
...
+		cleanupLog("cleaning up kprobe counter program")
+		return clusters.KustomizeDeleteForCluster(ctx, env.Cluster(), kprobeGoCounterKustomize)
+	})
+
+	t.Log("waiting for go kprobe counter userspace daemon to be available")
+	require.Eventually(t, func() bool {
+		daemon, err := env.Cluster().Client().AppsV1().DaemonSets(kprobeGoCounterUserspaceNs).Get(ctx, kprobeGoCounterUserspaceDsName, metav1.GetOptions{})
+		require.NoError(t, err)
+		return daemon.Status.DesiredNumberScheduled == daemon.Status.NumberAvailable
+	}, time.Minute, time.Second)
```
Billy99, line 40:
```suggestion
	},
	// Wait 5 minutes since cosign is slow, https://github.com/bpfman/bpfman/issues/1043
	5 * time.Minute, 10 * time.Second)
```
Add Andrew's temp work around.

bpfman-operator/test/integration/uprobe_test.go
```diff
@@ -0,0 +1,85 @@
...
+		cleanupLog("cleaning up target program")
+		return clusters.KustomizeDeleteForCluster(ctx, env.Cluster(), targetKustomize)
+	})
+
+	t.Log("waiting for go target userspace daemon to be available")
+	require.Eventually(t, func() bool {
+		daemon, err := env.Cluster().Client().AppsV1().DaemonSets(targetUserspaceNs).Get(ctx, targetUserspaceDsName, metav1.GetOptions{})
+		require.NoError(t, err)
+		return daemon.Status.DesiredNumberScheduled == daemon.Status.NumberAvailable
+	}, time.Minute, time.Second)
```
Billy99, line 43:
```suggestion
	},
	// Wait 5 minutes since cosign is slow, https://github.com/bpfman/bpfman/issues/1043
	5 * time.Minute, 10 * time.Second)
```
Ditto

bpfman-operator/test/integration/kprobe_test.go
anfredette: Is that still needed?  These have passed a few times without it.

Billy99: I would say just add it to keep them all consistent. If everything works, then no impact. If the network issue comes back, then all the tests will be covered.

anfredette: I copied this code before the workaround was added.  I agree with keeping them consistent.  However, 5 minutes seems like a long time.  I can add it to keep them consistent, but we should think about whether we want to keep the long wait.

Billy99: +1

astoycos:
@anfredette this just needs a `cargo +nightly  xtask public-api --bless`

Billy99 (approved):
LGTM

PR #1042, opened 2024-03-22
DROP pull in cosign fixes

astoycos:
This should fix our main!


astoycos:
For some reason Triangulation seems to be taking longer  🤔 This is causing the operator to time out in the k8s integration tests

Billy99 (approved):
LGTM

PR #1047, opened 2024-03-25
bpfman-operator: Add K8s support for fentry and fexit

Billy99:
Add K8s support for fentry and fexit


anfredette (approved):
Thanks for fixing a few of my kprobe/uprobe typos.

This looks good to me!

PR #1061, opened 2024-04-08
stop checking bpfman-api and bpfman-csi public_api

astoycos:
Stop checking the public-api for the bpfman-api
and bpfman-csi crates since they're autogenerated
and cause constant churn.


dave-tucker (approved):
LGTM

PR #1068, opened 2024-04-09
De-duplicate code in the bpfman-agent

anfredette:
- Created a common loop for reconciling *Programs
- Created a common reconcileBpfmanProgram() function
- Added some more getters to the bpfmanReconciler interface so we don't have to pass as many parameters around
- Made all of the functions in common.go that use ReconcilerCommon methods of ReconcilerCommon

Some additional changes:
- Used a different Type in the BpfProgram CRDs for fentry and fexit. With this change, the fentry and fexit controllers will only get triggered for the their own programs, and you can see the type when you get the bpfprogram CRD.
- Modified Makefile in examples
  - Made building go userspace code and bytecode dependencies of their respective build-*-image targets
  - Added a target to build-all-images
  - This combines 4 steps into one if you want to build both images.

Fixes: #1065

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,6 +23,32 @@ jobs:
       - name: Check License Header
         uses: apache/skywalking-eyes@cd7b195c51fd3d6ad52afceb760719ddc6b3ee91
 
+  get-more-space:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check disk space
+        run: df -h
+
+      - name: Free up space
+        run: |
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /opt/ghc
+          sudo rm -rf "/usr/local/share/boost"
+          sudo rm -rf "$AGENT_TOOLSDIRECTORY"
+          sudo rm -rf /usr/local/lib/android
+          sudo apt-get update
+          sudo eatmydata apt-get purge --auto-remove -y \
+            azure-cli aspnetcore-* dotnet-* ghc-* firefox \
+            google-chrome-stable \
+            llvm-* microsoft-edge-stable mono-* \
+            msbuild mysql-server-core-* php-* php7* \
+            powershell temurin-* zulu-*
+          sudo apt-get autoremove -y
+          sudo apt-get autoclean -y
+
+      - name: Check disk space again
+        run: df -h
+
   build:
     runs-on: ubuntu-latest
     steps:
@@ -279,7 +305,7 @@ jobs:
         run: cargo xtask integration-test
 
   kubernetes-integration-tests:
-    needs: [build, build-go, build-docs]
+    needs: [get-more-space, build, build-go, build-docs, basic-integration-tests]
     runs-on: ubuntu-latest
     env:
       BPFMAN_IMG: quay.io/bpfman/bpfman:int-test
@@ -287,6 +313,15 @@ jobs:
       BPFMAN_OPERATOR_IMG: quay.io/bpfman/bpfman-operator:int-test
       XDP_PASS_PRIVATE_IMAGE_CREDS: ${{ secrets.XDP_PASS_PRIVATE_IMAGE_CREDS }}
     steps:
+      - name: Check disk space
+        run: df -h
+
+      - name: Prune Docker
+        run: sudo docker system prune -a -f --volumes
+
+      - name: Check disk space
+        run: df -h
+
       - name: Install dependencies
         run: |
           sudo apt-get update
@@ -322,16 +357,28 @@ jobs:
         run: |
           go install github.com/cilium/ebpf/cmd/bpf2go@v0.14.0
           cd examples
-          make build-bc-images
+          make build-all-images
 
-      - name: build example userspace images
-        run: cd examples && make build-us-images
+      - name: Check disk space
+        run: df -h
 
       - name: build images
         run: cd bpfman-operator && make build-images
 
+      - name: Check disk space
+        run: df -h
+
+      - name: run cargo clean
+        run: cargo clean
+
+      - name: Check disk space
+        run: df -h
+
       - name: run integration tests
         run: cd bpfman-operator && make test-integration
+
+      - name: Check disk space
+        run: df -h
 
       ## Upload diagnostics if integration test step failed.
       - name: upload diagnostics
--- a/bpfman-operator/controllers/bpfman-agent/common.go
+++ b/bpfman-operator/controllers/bpfman-agent/common.go
@@ -74,19 +74,281 @@ type ReconcilerCommon struct {
 }
 
 // bpfmanReconciler defines a generic bpfProgram K8s object reconciler which can
-// program bpfman from user intent in K8s CRDs.
+// program bpfman from user intent in the K8s CRDs.
 type bpfmanReconciler interface {
-	getRecCommon() *ReconcilerCommon
-	reconcileBpfmanProgram(context.Context,
-		map[string]*gobpfman.ListResponse_ListResult,
-		*bpfmaniov1alpha1.BytecodeSelector,
-		*bpfmaniov1alpha1.BpfProgram,
-		bool,
-		bool,
-		*MapOwnerParamStatus) (bpfmaniov1alpha1.BpfProgramConditionType, error)
+	// SetupWithManager registers the reconciler with the manager and defines
+	// which kubernetes events will trigger a reconcile.
+	SetupWithManager(mgr ctrl.Manager) error
+	// Reconcile is the main entry point to the reconciler. It will be called by
+	// the controller runtime when something happens that the reconciler is
+	// interested in. When Reconcile is invoked, it initializes some state in
+	// the given bpfmanReconciler, retrieves a list of all programs of the given
+	// type, and then calls reconcileCommon.
+	Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error)
+	// getFinalizer returns the string used for the finalizer to prevent the
+	// BpfProgram object from deletion until cleanup can be performed
 	getFinalizer() string
+	// getRecType returns the type of the reconciler.  This is often the string
+	// representation of the ProgramType, but in cases where there are multiple
+	// reconcilers for a single ProgramType, it may be different (e.g., uprobe,
+	// fentry, and fexit)
 	getRecType() string
-	expectedBpfPrograms(ctx context.Context) (*bpfmaniov1alpha1.BpfProgramList, error)
+	// getProgType returns the ProgramType used by bpfman for the bpfPrograms
+	// the reconciler manages.
+	getProgType() internal.ProgramType
+	// getName returns the name of the current program being reconciled.
+	getName() string
+	// getExpectedBpfPrograms returns the list of BpfPrograms that are expected
+	// to be loaded on the current node.
+	getExpectedBpfPrograms(ctx context.Context) (*bpfmaniov1alpha1.BpfProgramList, error)
+	// getLoadRequest returns the LoadRequest that should be sent to bpfman to
+	// load the given BpfProgram.
+	getLoadRequest(bpfProgram *bpfmaniov1alpha1.BpfProgram, mapOwnerId *uint32) (*gobpfman.LoadRequest, error)
+	// getNode returns node object for the current node.
+	getNode() *v1.Node
+	// getBpfProgramCommon returns the BpfProgramCommon object for the current
+	// Program being reconciled.
+	getBpfProgramCommon() *bpfmaniov1alpha1.BpfProgramCommon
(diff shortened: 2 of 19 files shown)


bpfman-operator/controllers/bpfman-agent/common.go
```diff
@@ -74,19 +74,234 @@ type ReconcilerCommon struct {
...
+func (r *ReconcilerCommon) reconcileCommon(ctx context.Context, rec bpfmanReconciler,
+	programList []client.Object) (ctrl.Result, error) {
+
+	r.Logger.V(1).Info("Start reconcileCommon(), programList:")
+	for _, program := range programList {
+		r.Logger.V(1).Info("    Program", "Name", program.GetName())
+	}
+
+	// Get existing ebpf state from bpfman.
+	programMap, err := bpfmanagentinternal.ListBpfmanPrograms(ctx, r.BpfmanClient, rec.getProgType())
```
astoycos, line 132: nit can we make this var name more self describing? above we have `programList` referring to a list of K8s BpfProgram CRDs and here we have `programMap` referring to a map of loaded programs.  Maybe it should be `bpfProgramCrds` and `loadedBpfPrograms` respectively?

bpfman-operator/controllers/bpfman-agent/common.go
```diff
@@ -74,19 +74,234 @@ type ReconcilerCommon struct {
...
+
+	requeue := false // initialize requeue to false
+	for _, program := range programList {
+		r.Logger.V(1).Info("Reconciling program", "Name", program.GetName())
+
+		err := rec.setCurrentProgram(program)
+		if err != nil {
+			r.Logger.Error(err, "Failed to set current program")
+			return ctrl.Result{Requeue: true, RequeueAfter: retryDurationAgent}, nil
+		}
```
astoycos, line 146: Can you add a comment here, what `Program`? Is is a K8s Crd, an actual ebpf program?

bpfman-operator/controllers/bpfman-agent/common.go
```diff
@@ -74,19 +74,234 @@ type ReconcilerCommon struct {
...
+
+		// Make sure if we're not selected just exit
+		if !isNodeSelected {
+			return bpfmaniov1alpha1.BpfProgCondNotSelected, nil
+		}
+
+		// If a container selector is present but there were no matching
+		// containers on this node, just exit.
+		if noContainers {
```
astoycos, line 207: hrm this comment is confusing me, as we're saying that it's checking "if container selector is present" but in reality it's only checking if there are "noContainers" maybe just update the `noContainersOnNode(` functions description?

bpfman-operator/controllers/bpfman-agent/common.go
```diff
@@ -74,19 +74,234 @@ type ReconcilerCommon struct {
...
+	// prog ID should already have been set if program exists
+	id, err := bpfmanagentinternal.GetID(bpfProgram)
+	if err != nil {
+		r.Logger.Error(err, "Failed to get program ID")
+		return bpfmaniov1alpha1.BpfProgCondNotLoaded, nil
+	}
+
+	// BpfProgram exists but either Program is being deleted, node is no
+	// longer selected, or map is not available....unload program
+	if isBeingDeleted || !isNodeSelected || noContainers ||
```
astoycos, line 247: Should we break this up into multiple if statements for readability? It seems like we end up doing it anyways deeper in this if statement

bpfman-operator/controllers/bpfman-agent/fentry-program.go
```diff
@@ -130,54 +153,23 @@ func (r *FentryProgramReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 	}
 
-	// Get existing ebpf state from bpfman.
-	programMap, err := bpfmanagentinternal.ListBpfmanPrograms(ctx, r.BpfmanClient, internal.Tracing)
-	if err != nil {
-		r.Logger.Error(err, "failed to list loaded bpfman programs")
-		return ctrl.Result{Requeue: true, RequeueAfter: retryDurationAgent}, nil
+	// Create a list of fentry programs to pass into reconcileCommon()
+	var fentryObjects []client.Object = make([]client.Object, len(fentryPrograms.Items))
+	for i := range fentryPrograms.Items {
```
astoycos, line 158: I searched for a way to do this that was more direct with no luck :/ this will work

astoycos (commented):
Looking good, thanks for this. I also pulled locally and ran, things look pretty good when running the xdp example.

The only thing I noticed is the uprobe reconciler seems to run even when not needed at times.

```
{"level":"info","ts":"2024-04-24T19:43:42Z","msg":"Reconcile XDP: Enter","controller":"xdpprogram","controllerGroup":"bpfman.io","controllerKind":"XdpProgram","XdpProgram":{"name":"go-xdp-counter-example"},"namespace":"","name":"go-xdp-counter-example","reconcileID":"2b4f205a-78b4-415e-9989-47a3c9889bab","ReconcileKey":"/go-xdp-counter-example"}
{"level":"info","ts":"2024-04-24T19:43:42Z","logger":"xdp","msg":"calling bpfman to load Program on Node","bpfProgram Name":"go-xdp-counter-example-bpfman-deployment-control-plane-eth0"}
{"level":"info","ts":"2024-04-24T19:43:50Z","logger":"xdp","msg":"Updating bpfProgram Object","Id":12321,"bpfProgram":"go-xdp-counter-example-bpfman-deployment-control-plane-eth0"}
{"level":"info","ts":"2024-04-24T19:43:50Z","msg":"Reconcile XDP: Enter","controller":"xdpprogram","controllerGroup":"bpfman.io","controllerKind":"XdpProgram","XdpProgram":{"name":"go-xdp-counter-example"},"namespace":"","name":"go-xdp-counter-example","reconcileID":"25f8680c-261e-4775-8fef-5d819d390b65","ReconcileKey":"/go-xdp-counter-example"}
{"level":"info","ts":"2024-04-24T19:43:50Z","msg":"Reconcile XDP: Enter","controller":"xdpprogram","controllerGroup":"bpfman.io","controllerKind":"XdpProgram","XdpProgram":{"name":"go-xdp-counter-example"},"namespace":"","name":"go-xdp-counter-example","reconcileID":"4289808a-81b8-4f84-9917-2abf3cac7036","ReconcileKey":"/go-xdp-counter-example"}
{"level":"info","ts":"2024-04-24T19:43:50Z","logger":"xdp","msg":"Nothing to do for this program"}
{"level":"info","ts":"2024-04-24T19:43:50Z","msg":"Reconcile XDP: Enter","controller":"xdpprogram","controllerGroup":"bpfman.io","controllerKind":"XdpProgram","XdpProgram":{"name":"go-xdp-counter-example"},"namespace":"","name":"go-xdp-counter-example","reconcileID":"1da80820-935c-4ffa-b6cf-0a0d6993f1df","ReconcileKey":"/go-xdp-counter-example"}
{"level":"info","ts":"2024-04-24T19:43:50Z","logger":"xdp","msg":"Nothing to do for this program"}
{"level":"info","ts":"2024-04-24T19:43:57Z","msg":"Reconcile Uprobe: Enter","controller":"uprobeprogram","controllerGroup":"bpfman.io","controllerKind":"UprobeProgram","UprobeProgram":{"name":"go-xdp-counter-ds-bcs98","namespace":"go-xdp-counter"},"namespace":"go-xdp-counter","name":"go-xdp-counter-ds-bcs98","reconcileID":"f01b8bf5-28ea-4f45-b686-a621144371ea","ReconcileKey":"go-xdp-counter/go-xdp-counter-ds-bcs98"}
{"level":"info","ts":"2024-04-24T19:43:57Z","logger":"uprobe","msg":"UprobeProgramController found no Uprobe Programs"}
```

bpfman-operator/controllers/bpfman-agent/fentry-program.go
anfredette: I also searched some more after we talked, but same thing here.

bpfman-operator/controllers/bpfman-agent/common.go
anfredette: This is the way it was.  However, I'll see what I can do.

bpfman-operator/controllers/bpfman-agent/common.go
anfredette: In getExpectedBpfPrograms(), if there is a container selector present, I see if I can find any matching containers on the given node.  If I can't find any matching containers, I set the UprobeNoContainersOnNode annotation. The noContainersOnNode() function returns true if that annotation exists and is set to true.  Does that make sense?  So, in effect, that's what it's telling us, but indirectly.  I can see if I can make the comments more clear.

bpfman-operator/controllers/bpfman-agent/common.go
anfredette: Dond

bpfman-operator/controllers/bpfman-agent/common.go
anfredette: I belive this is the way it was, but I agree, your suggestions are more clear. Done

bpfman-operator/controllers/bpfman-agent/common.go
```diff
@@ -74,19 +74,234 @@ type ReconcilerCommon struct {
...
+// reconciler.  It reconciles each program in the list.  reconcileCommon should
+// not return error because it will trigger an infinite reconcile loop.
+// Instead, it should report the error to user and retry if specified. For some
+// errors the controller may decide not to retry. Note: This only results in
+// calls to bpfman if we need to change something
+func (r *ReconcilerCommon) reconcileCommon(ctx context.Context, rec bpfmanReconciler,
+	programList []client.Object) (ctrl.Result, error) {
+
+	r.Logger.V(1).Info("Start reconcileCommon(), programList:")
+	for _, program := range programList {
```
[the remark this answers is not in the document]
anfredette: I'm going to delete the loop. The names get printed out below, but I added this for debugging because it wasn't working the way I expected.  Now it works right, so I don't think it's needed.

anfredette:
> The only thing I noticed is the uprobe reconciler seems to run even when not needed at times.

Because uprobes can be attached inside containers, we need to attach them if new pods are created and delete them when pods are deleted.  So, we watch for pod events and check for relevant changes.

bpfman-operator/controllers/bpfman-agent/common.go
anfredette: I agree that this looks funky.  I think that the reason it was done this way is that for all these cases, the program needs to be unloaded, however, we return a different a different condition for each one.  The way it's written, we only have one call to UnloadBpfmanProgram().  If we break it up, we'd need to make the UnloadBpfmanProgram() call for each case.  Still looking for a way to make this whole function less confusing...

bpfman-operator/controllers/bpfman-agent/common.go
anfredette: I ended up doing the following and tried to make it consistent throughout the file:
program = *Program CRD, programs = program list
bpfProgram = BpfProgram CRD, bpfPrograms = bpfProgram list
loadedBpfProgram = list result from bpfman, loadedBpfPrograms = loadedBpfProgram list

bpfman-operator/controllers/bpfman-agent/common.go
anfredette: I completely rewrote reconcileBpfProgram() into somethign that I think is more logical and readable.  Please let me know what you think.

anfredette:
I'm done addressing comments.  This is ready for review again.

.github/workflows/build.yml
```diff
@@ -279,14 +305,23 @@ jobs:
...
-    needs: [build, build-go, build-docs]
+    needs: [get-more-space, build, build-go, build-docs, basic-integration-tests]
     runs-on: ubuntu-latest
     env:
       BPFMAN_IMG: quay.io/bpfman/bpfman:int-test
       BPFMAN_AGENT_IMG: quay.io/bpfman/bpfman-agent:int-test
       BPFMAN_OPERATOR_IMG: quay.io/bpfman/bpfman-operator:int-test
       XDP_PASS_PRIVATE_IMAGE_CREDS: ${{ secrets.XDP_PASS_PRIVATE_IMAGE_CREDS }}
     steps:
+      - name: Check disk space
```
astoycos, line 316: I don't think we need to do this before building the images

anfredette: I wanted to see how much disk space we were starting `kubernetes-integration-tests` with, and how much the `docker system prune` recovered.  I also wanted to see how much each step was consuming.  I don't think running `df -h` is very expensive and it's good to keep them in there so we can keep track of the usage over time in case we hit the wall again.

astoycos (approved):
Let's get this in and then go back and fixup things 👍

PR #1083, opened 2024-04-11
update packit config

astoycos:
Start building for stable on PRs so that we
have some coverage.

Propose downsream RPM releases whenever we cut
a release upstream

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -28,22 +28,16 @@ jobs:
     owner: "@ebpf-sig"
     project: bpfman-next
     targets:
-      - fedora-all
+      - fedora-development
+      - fedora-stable
   - job: tests
     trigger: commit
     targets:
-      - fedora-all
+      - fedora-development
+      - fedora-stable
     specfile_path: bpfman.spec
-  - job: copr_build
+  - job: propose_downstream
     trigger: release
-    branch: main
-    specfile_path: bpfman.spec
-    owner: "@ebpf-sig"
-    project: bpfman
-    targets:
-      - fedora-all
-  - job: tests
-    trigger: release
-    targets:
-      - fedora-all
-    specfile_path: bpfman.spec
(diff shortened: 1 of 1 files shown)


astoycos:
So I probably don't want to build-on each PR here since it's pretty slow, but the release bits will still be needed.

.packit.yml
```diff
@@ -35,15 +35,14 @@ jobs:
       - fedora-all
     specfile_path: bpfman.spec
   - job: copr_build
-    trigger: release
-    branch: main
+    trigger: pull_request
+    targets: [fedora-stable]
```
danielmellado, line 39: nit, I'd format this in a different way. IMHO it enhances readability:
```suggestion
    targets:
      - fedora-stable
```

.packit.yml
```diff
@@ -35,15 +35,14 @@ jobs:
       - fedora-all
     specfile_path: bpfman.spec
   - job: copr_build
-    trigger: release
-    branch: main
+    trigger: pull_request
```
danielmellado, line 38: if not specified, this would trigger builds for pull request with any target branch.

.packit.yml
danielmellado: Besides that, I'd be interested in getting to know if we're targeting any specific arch.

.packit.yml
astoycos: Right I just did that for some testing 
I'll remove that bit.

astoycos:
> Besides the comment, I was thinking that we may want to add a new CI validation for the packit config itself, calling `packit validate-config`

Sounds good what sort of validation are you thinking? Just to make sure no one changes it?

danielmellado (approved):
lgtm now, thanks!

Billy99 (approved):
LGTM

Issue #1080, opened 2024-04-11
Use "//go:generate go run github.com/cilium/ebpf/cmd/bpf2go ..." instead of just "//go:generate bpf2go ..."


thediveo:
In its examples, [`@cilium/ebpf`](https://github.com/cilium/ebpf) uses

```go
//go:generate go run github.com/cilium/ebpf/cmd/bpf2go ...
```

instead of just 

```go
//go:generate bpf2go ...
```

The rationale is that this does automatically pull and build bpf2go without the user having to explicitly install the tool first.

This allows for the Go-idiomatic way to get the example up and running without much ado:

```bash
cd examples/go-xdp-counter
go generate . # generate eBPF .o
go run -exec sudo . -iface ens33
```

thediveo:
Ah, the nudge 😁 ... can work on a PR. I could tackle both examples and the associated description for the bpfman.io, if that's okay?

Juwt to clarify one aspect here: based on the system.d configuration I assume access rights `0660` for the `bpfman.sock`: if this assumption is correct, then I'll update the docs in the PR to correctly reflect the need of `-exec sudo` when runningnthe Go examples.

Off tangent: could you please rename `bpfman`, as I'm constantly mistyping it as `pdfman`, please?

astoycos:
> Ah, the nudge 😁 ... can work on a PR. I could tackle both examples and the associated description for the bpfman.io, if that's okay?

Ahaha all good I was just triaging :) If you want to snag it I wouldn't stop you though! 

> Just to clarify one aspect here: based on the system.d configuration I assume access rights 0660 for the bpfman.sock: if this assumption is correct, then I'll update the docs in the PR to correctly reflect the need of -exec sudo when running the Go examples.

Correct! That brings up a good point though....and maybe another issue? I think it would be cool to make it possible to run the example programs on the host without root privs... @Billy99 We used to have the bpfman user group (but don't anymore) do we bring that back? 

> Off tangent: could you please rename bpfman, as I'm constantly mistyping it as pdfman, please?

LOL We're still reeling from the [most recent name change](https://bpfman.io/v0.4.0/blog/2023/11/23/bpfd-becomes-bpfman/)  😉

thediveo:
> Correct! That brings up a good point though....and maybe another issue? I think it would be cool to make it possible to run the example programs on the host without root privs... @Billy99 We used to have the bpfman user group (but don't anymore) do we bring that back?

[...]

PR #1053, opened 2024-03-28
docs: Scrub documentation for Release 0.4.0

Billy99:
Update all the documentation in preparation for the v0.4.0 release.

Fixes: #1038

PR #1084, opened 2024-04-13
Examples/Docs: idiomatic Go usage and update installation instructions for apt-based OS

thediveo:
- fixes issue #1080 
- update apt-based OS instructions to install missing packages for bpfman development, as per Ubuntu 23.10.
- as per [cilium/ebpf examples](https://github.com/cilium/ebpf/tree/main/examples) use "go run" and the full import path of the bpf2go tool in order to avoid developers having to explicitly install bpf2go; this way, the tool is automatically pulled and build if not present in Go's cache.
- document the more Go idiomatic use of `go run -exec sudo . ...` instead of `sudo ./go-... ...`, as users can immediately run an example without having to explicitly build everything first.

### DIFF
--- a/docs/getting-started/building-bpfman.md
+++ b/docs/getting-started/building-bpfman.md
@@ -200,6 +200,22 @@ sudo dnf install llvm-devel clang-devel elfutils-libelf-devel
 sudo apt install clang lldb lld libelf-dev gcc-multilib
 ```
 
+### Install libssl Development Library
+
+`apt` based OS:
+
+```console
+sudo apt install libssl-dev
+```
+
+### Install bpf Helper Header Files
+
+`apt` based OS:
+
+```console
+sudo apt install libbpf-dev
+```
+
 ### Install Protobuf Compiler
 
 For further detailed instructions, see [protoc](https://grpc.io/docs/protoc-installation/).
--- a/docs/getting-started/example-bpf-local.md
+++ b/docs/getting-started/example-bpf-local.md
@@ -8,7 +8,7 @@ Assume the following command is run:
 
 ```console
 cd bpfman/examples/go-xdp-counter/
-sudo ./go-xdp-counter -iface eno3
+go run -exec sudo . -iface eno3
 ```
 
 The diagram below shows `go-xdp-counter` example, but the other examples operate in
@@ -50,7 +50,7 @@ The output should show the count and total bytes of packets as they pass through
 interface as shown below:
 
 ```console
-sudo ./go-xdp-counter --iface eno3
+go run -exec sudo . --iface eno3
 2023/07/17 17:43:58 Using Input: Interface=eno3 Priority=50 Source=/home/<$USER>/src/bpfman/examples/go-xdp-counter/bpf_bpfel.o
 2023/07/17 17:43:58 Program registered with id 6211
 2023/07/17 17:44:01 4 packets received
--- a/examples/go-kprobe-counter/main.go
+++ b/examples/go-kprobe-counter/main.go
@@ -31,7 +31,7 @@ type Stats struct {
 	Counter uint64
 }
 
-//go:generate bpf2go -cc clang -no-strip -cflags "-O2 -g -Wall" bpf ./bpf/kprobe_counter.c -- -I.:/usr/include/bpf:/usr/include/linux
+//go:generate go run github.com/cilium/ebpf/cmd/bpf2go -cc clang -no-strip -cflags "-O2 -g -Wall" bpf ./bpf/kprobe_counter.c -- -I.:/usr/include/bpf:/usr/include/linux
 func main() {
 	stop := make(chan os.Signal, 1)
 	signal.Notify(stop, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
--- a/docs/getting-started/example-bpf.md
+++ b/docs/getting-started/example-bpf.md
@@ -149,10 +149,6 @@ If not, see [Setup and Building bpfman](./building-bpfman.md)._
 
     **Ubuntu:** `sudo apt-get install libbpf-dev`
 
-3. Cilium's `bpf2go` binary
-
-    `go install github.com/cilium/ebpf/cmd/bpf2go@v0.14.0`
-
 #### Build
 
 To build all the C based eBPF counter bytecode, run:
--- a/examples/go-tc-counter/main.go
+++ b/examples/go-tc-counter/main.go
@@ -36,7 +36,7 @@ const (
 	TC_ACT_OK = 0
 )
 
-//go:generate bpf2go -cc clang -no-strip -cflags "-O2 -g -Wall" bpf ./bpf/tc_counter.c -- -I.:/usr/include/bpf:/usr/include/linux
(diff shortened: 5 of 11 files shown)


docs/getting-started/building-bpfman.md
```diff
@@ -200,6 +200,22 @@ sudo dnf install llvm-devel clang-devel elfutils-libelf-devel
...
 
+### Install libssl Development Library
+
+`apt` based OS:
+
+```console
+sudo apt install libssl-dev
+```
+
+### Install bpf Helper Header Files
```
Billy99, line 211: This install is described in https://bpfman.io/main/getting-started/example-bpf/#building-locally, which is linked from https://bpfman.io/main/getting-started/example-bpf-local/#running-example-programs

Do you think we should move the examples specific install packages to this file? Was trying to keep the examples stuff separate, but may not be the best practice.

docs/getting-started/building-bpfman.md
```diff
@@ -200,6 +200,22 @@ sudo dnf install llvm-devel clang-devel elfutils-libelf-devel
 sudo apt install clang lldb lld libelf-dev gcc-multilib
 ```
 
+### Install libssl Development Library
```
Billy99, line 203: We were trying to remove the SSL dependency. Do you know if this was needed for the examples or bpfman?

docs/getting-started/example-bpf-local.md
```diff
@@ -8,7 +8,7 @@ Assume the following command is run:
 
 ```console
 cd bpfman/examples/go-xdp-counter/
-sudo ./go-xdp-counter -iface eno3
+go run -exec sudo . -iface eno3
```
Billy99, line 11: I'm not a huge fan of this format, but I'm not a Go purest. There are 4 places in https://github.com/bpfman/bpfman/blob/main/docs/getting-started/example-bpf.md#running-examples that we show `sudo ./go-xxx-counter ...` examples. I guess we should keep them consistent.

examples/go-kprobe-counter/main.go
```diff
@@ -31,7 +31,7 @@ type Stats struct {
 	Counter uint64
 }
 
-//go:generate bpf2go -cc clang -no-strip -cflags "-O2 -g -Wall" bpf ./bpf/kprobe_counter.c -- -I.:/usr/include/bpf:/usr/include/linux
+//go:generate go run github.com/cilium/ebpf/cmd/bpf2go -cc clang -no-strip -cflags "-O2 -g -Wall" bpf ./bpf/kprobe_counter.c -- -I.:/usr/include/bpf:/usr/include/linux
```
Billy99, line 34: Since we don't need to install bpf2go any more, please remove it from https://bpfman.io/main/getting-started/example-bpf/#building-locally

Billy99 (changes requested):

Billy99:
Thanks for the PR. Just some documentation nits.

docs/getting-started/building-bpfman.md
thediveo: I have to admit that I completely missed the "Building Locally" for building the examples.

left sidebar

In its current form, the documentation first describes how to deploy the example programs, but not how to build them. This is only shown in the last-but-one item, and then it's not even clear from the sidebar that it includes instructions for the needed building.

I'm open to discussing this further, as my gut feeling is that this part of the documentation needs refactoring: at least the example building instructions should come before the deployment examples. Maybe we should split the building instructions from https://bpfman.io/main/getting-started/example-bpf/#examples-in-kubernetes-deployment, keeping the major rest as is; the split part then could move into its own new "Building Examples" that comes before https://bpfman.io/main/getting-started/example-bpf-local/

docs/getting-started/example-bpf-local.md
thediveo: The reason for the `go run -exec sudo` idiom is that it
- works also for unit testing,
- allows quick turn-around for on-the-fly experiments,
- doesn't leave unwanted binaries lying around.

examples/go-kprobe-counter/main.go
thediveo: This probably needs more refactoring:
1. remove installing bpf2go installation instruction.
2. regenerating the eBPF bytecode normally isn't necessary, as it already comes with the examples. I would like to suggest moving this into its own section about "Tinkering with Examples".
3. building the examples also isn't necessary, as `go run ...` takes care of building on-the-fly.

IMHO this would simplify the examples user journey, at least for the local host. Shall I move forward on this?

docs/getting-started/building-bpfman.md
Billy99: I think I got into that habit with my personal notes. You install once, but revisit the notes multiple times to remember how to do something and the install instructions are in the way. So I add them to the bottom with a link. But for this, it makes sense to not follow that pattern.

docs/getting-started/example-bpf-local.md
Billy99: Ack

examples/go-kprobe-counter/main.go
Billy99: I agree with everything except Bullet 2. eBPF bytecode doesn't come with the examples, unless I'm missing something. We used to check it into git, but that was removed several months ago.

docs/getting-started/building-bpfman.md
Billy99: I agree with your assessment. We have already been discussing other changes to the docs (see comments in #1053). So let's punt on this for your PR and we will address it with some of the overall docs rework.

thediveo: Shall I keep this change, or drop it?

examples/go-kprobe-counter/main.go
Billy99, line 34: Sorry, one last request. While your PR was pending, a new example was created. Do mind adding your changes to https://github.com/bpfman/bpfman/blob/main/examples/go-uretprobe-counter/main.go#L29

Billy99 (commented):
Looks great. Thank you! One request.

Billy99:
@thediveo The kubernetes-integration-tests issue looks to be an issue with the last PR that merged. Give me a min to dig further.

Billy99:
@thediveo Ok, so looks like it was your PR. Took a while to hunt down. Since you are pulling bpf2go from upstream, it relies on the version in go.mod, which currently is v0.10.0. On main, we are using v0.14.0. The different version changed the generated file name, move `_x86` around in the name.

So in your PR, run `go get github.com/cilium/ebpf` and run `go tidy` and  that should fix it. It will also bump Go from 1.19 to 1.21, but that is fine.

thediveo:
> @thediveo Ok, so looks like it was your PR. Took a while to hunt down. Since you are pulling bpf2go from upstream, it relies on the version in go.mod, which currently is v0.10.0. On main, we are using v0.14.0. The different version changed the generated file name, move `_x86` around in the name.

Oh, sorry for the inconvenience and big thanks for tracking this down!

> So in your PR, run `go get github.com/cilium/ebpf` and run `go tidy` and that should fix it. It will also bump Go from 1.19 to 1.21, but that is fine.

A `go mod tidy` brings in a "lovely" `toolchain` declaration, which I deleted before committing. I'm still not friends with this `toolchain` concept introduced in Go 1.22 when it comes to deterministic CI pipelines and maintaining open source tool modules -- and I don't seem to be the only one from Go's issue tracker...

thediveo:
Many thanks to you!

PR #1087, opened 2024-04-16
bpfman: include license in crate workspace

danielmellado:
This PR includes the licenses files in the crate workspace subdirectory.
Without this, they won't be showing on crates.io and would be giving out
errors on tooling such as rust2rpm.


dave-tucker (approved):
LGTM. We'll just need to fix the license-check CI workflow to ignore symlinks

Issue #596, opened 2023-07-17
Document/verify how bpfman users can interact with bpfd on Openshift with 0 privileges


astoycos:
Currently (as of release v0.4.0) we still do not support applications running on Openshift with zero permissions, part of this was fixed in #829  However this still needs to be verified and any other bugs worked out.

astoycos:
In the OLM documentation

astoycos:
Done in #1097

PR #1089, opened 2024-04-16
add csi fsgroup support

astoycos:
Add support for our CSI driver to be able to
assign a passed fsGroup value to the mounted bpffs. see https://kubernetes-csi.github.io/docs/support-fsgroup.html for more.

This allows unprivileged applications to access
those pin points when configured correctly
with `fsGroup`, `runAsUser`, and `runAsGroup` for
a kubernetes workload.

A First step in #596 

this also adresses a bunch of small formatting things

### DIFF
--- a/examples/config/base/go-kprobe-counter/deployment.yaml
+++ b/examples/config/base/go-kprobe-counter/deployment.yaml
@@ -27,9 +27,11 @@ spec:
         name: go-kprobe-counter
     spec:
       nodeSelector: {}
-      hostNetwork: true
       dnsPolicy: ClusterFirstWithHostNet
       serviceAccountName: bpfman-app-go-kprobe-counter
+      securityContext:
+        runAsNonRoot: true
+        fsGroup: 65534
       tolerations:
         # these tolerations are to have the daemonset runnable on control plane nodes
         # remove them if your control plane nodes should not run pods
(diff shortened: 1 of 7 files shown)


examples/config/base/go-kprobe-counter/deployment.yaml
```diff
@@ -27,9 +27,11 @@ spec:
         name: go-kprobe-counter
     spec:
       nodeSelector: {}
-      hostNetwork: true
       dnsPolicy: ClusterFirstWithHostNet
       serviceAccountName: bpfman-app-go-kprobe-counter
+      securityContext:
+        runAsNonRoot: true
+        fsGroup: 65534
```
Billy99, line 34: Do we need anything in our documentation around this? If the user creates their own deployment, is 65534 a magic number, can they use any value they want?

astoycos: I was hoping you'd see that..... 65534 isn't magic but it is random :) 

I don't think we need to document anything as it's covered in upstream k8s docs https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod
and should be pretty common for any user wishing to run an unprivileged application.

Billy99 (approved):
LGTM

Issue #1095, opened 2024-04-17
Running from RPM is missing bpfman-rpc


Silvanoc:
I'm trying to run [bpfman from RPM as documented](https://bpfman.io/v0.4.0/getting-started/running-rpm/#install-rpm-from-packit-service). The current version is *0.4.0~dev-1.20240417072016942084.main.13.g5fc9011.fc39*, DNF reports that the version documented in that page doesn't exist. My system is a Fedora 39 installation.

The systemd service is failing because `/usr/sbin/bpfman-rpc` is missing. I've reinstalled the package, listed the content of the package, that file doesn't appear anywhere.

Billy99:
Yes, this is a problem. Thank you. We will add this to the list of issues for the v0.4.1 release.

PR #1096, opened 2024-04-17
rpm: Add bpfman-rpc and bpfman-ns to rpm package

Billy99:
bpfman-rpc and bpfman-ns are not currently in the bpfman RPM. Add both binaries.

Resolves: #1095


astoycos (commented):
DANG I thought I had done this in https://github.com/bpfman/bpfman/pull/1058 but it must have gotten lost :/ Thanks for picking this up @Billy99 

LGTM

astoycos (approved):

PR #1102, opened 2024-04-24
fix mounter image to fedora 39

astoycos:
Fix the image we use to mount the default bpffs
if it's not mounted (i.e kind) to fedora 39 since
the newer fedora-minimal images don't have the
mount binary by default.


anfredette (approved):
/lgtm

PR #1107, opened 2024-04-29
test: add image name test

Billy99:
Adding one additional test to integration error test cases. For loading from image, if no name is provided, use name from OCI Image. If name is provided, use that instead. So added a test to pass in an invalid name from CLI when laoding from image and make sure command fails.

Also global changed all forms of selinux to SELinux.

### DIFF
--- a/docs/developer-guide/k8s-selinux-distros.md
+++ b/docs/developer-guide/k8s-selinux-distros.md
@@ -1,14 +1,14 @@
-# Running the Examples as Non-Root on Selinux Distributions
+# Running the Examples as Non-Root on SELinux Distributions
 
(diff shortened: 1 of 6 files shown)


docs/developer-guide/k8s-selinux-distros.md
```diff
@@ -1,14 +1,14 @@
-# Running the Examples as Non-Root on Selinux Distributions
+# Running the Examples as Non-Root on SELinux Distributions
```
astoycos, line 1: Thanks sorry I missed this :/

astoycos (approved):
LGTM thanks!

PR #1110, opened 2024-04-29
Free up disk space for running k8s int tests

anfredette:
PR #1068 attempted to do this, in part, by creating a separate `get-more-space` job, but since each job runs in its own runner, the `get-more-space` job didn't help.  This PR moves some of the `get-more-space` steps into the `kubernetes-integration-tests` job.

The solution in this pr is much better than the one in #1068 because all I had to do was delete some unneeded directories in the `kubernetes-integration-tests` job.  This is much faster and still frees up plenty of disk space.


astoycos (approved):
LGTM

PR #1111, opened 2024-04-30
Remove propose_downstream packit job

danielmellado:
So far we're still proposing the package to be acceptef on Fedora, so
this job would just fail. Let's remove it until we get the package in
and it gets accepted.


astoycos (approved):
LGTM thanks!

PR #1124, opened 2024-05-06
WIP: Add a node status field to *Programs

anfredette:
This PR is a POC of a different way of maintaining the cluster-level status for each *Program CRD

With the current implementation, the operator controller determines a cluster-wide status for each *Program CRD by watching and evaluating all of the BpfProgram CRDs related to the given *Program.

This PR adds a NodeStatus field to the *Program CRD that holds the status for each node, and the bpfman agent controller for each node is responsible for updating it's own node-level status element in the NodeStatus field.  With this change, the Operator controller doesn't have to watch all bpfPrograms, but only the NodeStatus field in the *Programs. This also distributes some of the processing that the controller agent currently has to do across all the nodes, so it should scale better.

I implemented two options:
1. NodeStatus []NodeStatusEntry
2. NodeConditions []metav1.Condition

I think the NodeConditions option is a twisted use of conditions and the NodeStatus option is cleaner. So, my preference is NodeStatus.

This only includes the bpfman agent part of updating the NodeStatus/Conditions. The controller would need to watch the node level status and determine a global cluster-level status for each *Program based on the per-node status.

TODOs:
1. Investigate using Server-Side Apply.
2. Implement operator controller support.
3. Scale testing.
4. Fix samples/bpfman.io_v1alpha1_kprobe_kprobeprogram.yaml (I changed it so I had a failing test case).


anfredette:
This PR represents an evaluation and POC of a distributed way of updating the cluster-level status for each *Program CRD.  We are not planning to implement this solution at this time, so I'm closing this PR.

PR #1125, opened 2024-05-07
fixup the bpfman-csv

astoycos:
- Fixup the csv file
- Move the custom SCC into each application since the CSVs do not support holding the SCC type unfortunately 

see https://github.com/operator-framework/operator-lifecycle-manager/issues/2847

### DIFF
--- a/examples/config/selinux/go-tc-counter/user-scc.yaml
+++ b/examples/config/selinux/go-tc-counter/user-scc.yaml
@@ -1,0 +1,52 @@
+---
+apiVersion: security.openshift.io/v1
+kind: SecurityContextConstraints
+metadata:
+  name: bpfman-restricted
+allowHostDirVolumePlugin: false
+allowHostIPC: false
+allowHostNetwork: false
+allowHostPID: false
+allowHostPorts: false
+allowPrivilegeEscalation: false
+allowPrivilegedContainer: false
+allowedCapabilities: null
+defaultAddCapabilities: null
+groups: []
+priority: null
+readOnlyRootFilesystem: false
+requiredDropCapabilities:
+  - ALL
+runAsUser:
+  type: RunAsAny
+seLinuxContext:
+  type: RunAsAny
+seccompProfiles: null
+supplementalGroups:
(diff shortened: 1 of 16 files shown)


examples/config/selinux/go-tc-counter/user-scc.yaml
```diff
@@ -0,0 +1,52 @@
+---
+apiVersion: security.openshift.io/v1
+kind: SecurityContextConstraints
+metadata:
+  name: bpfman-restricted
```
Billy99, line 5: Does it matter that all the user-scc.yaml files in the examples are using the same name, `bpfman-resticted` and below, `bpfman-user`?

astoycos: No, you'd just get a message "these already exist"... It's annoying because I would rather the bpfman-restricted scc and bpfman-user Cluster role be deployed whenever we deploy bpfman... However the bundle's CSV can't deploy those types.  

there are other options besides doing this though

1. Add an extra step to the documentation telling users they need to manually create these
OR 
2. Have the bpfman-operator itself deploy the custom SCC + ClusterRole

but for now just including them in the example manifest seemed to make the most sense 

I just wanted to avoid changing the previous experience which was 

`make deploy-xdp-selinux`

to something like

`make unprivileged-bpfman-user` followed by `make deploy-xdp-selinux`

Billy99: Ok, what if I run:

```
make deploy-xdp-selinux
make deploy-tc-selinux

make undeploy-xdp-selinux
```

Does that remove `bpfman-restricted` and `bpfman-user` for TC?

astoycos: Hrmmm good point, It might just keep working since something is still referencing it, but yeah It's not Ideal....I guess we need to just have a separate command for now when deploying on selinux, I'll update this PR for that

astoycos:
Close in favor of https://github.com/bpfman/bpfman-operator/pull/19

PR #1134, opened 2024-05-10
RPM fixes 

astoycos:
A bunch of fixes based on the feedback in https://bugzilla.redhat.com/show_bug.cgi?id=2269411

- [ ] https://github.com/bpfman/bpfman/issues/1127 -> b0530dd054e00657f9af3710a85a28e303207cc4
     - I we need to patch `ring v0.17.8` and `sigstore-protobuf-specs` crates since they don't specify any license metadata. I didn't do it here since we don't maintain a version of the vendor tarball in github
- [x] https://github.com/bpfman/bpfman/issues/1128 -> a4c3a9f1a2142ce28de7159dd290ccce23fd1043
- [x] https://github.com/bpfman/bpfman/issues/1129
- [x] https://github.com/bpfman/bpfman/issues/1130
- [x] https://github.com/bpfman/bpfman/issues/1131
- [x] https://github.com/bpfman/bpfman/issues/1132
- [x] https://github.com/bpfman/bpfman/issues/1133

### DIFF
--- a/csi/Cargo.toml
+++ b/csi/Cargo.toml
@@ -1,11 +1,11 @@
 [package]
 description = "gRPC bindings to the CSI spec"
-name = "bpfman-csi"
-version = "1.8.0"
 edition.workspace = true
 homepage.workspace = true
 license.workspace = true
+name = "bpfman-csi"
 repository.workspace = true
+version = "1.8.0"
 
 [dependencies]
 prost = { workspace = true, features = ["prost-derive", "std"] }
--- a/csi/Cargo.toml
+++ b/csi/Cargo.toml
@@ -1,10 +1,11 @@
 [package]
 description = "gRPC bindings to the CSI spec"
-edition.workspace = true
-license.workspace = true
 name = "bpfman-csi"
-repository.workspace = true
 version = "1.8.0"
+edition.workspace = true
+homepage.workspace = true
+license.workspace = true
+repository.workspace = true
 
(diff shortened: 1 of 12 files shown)


danielmellado:
Let's skip for now the first item as we can handle that within the spec submission to Fedora. Thanks for combining this all in this PR!

csi/Cargo.toml
```diff
@@ -1,10 +1,11 @@
 [package]
 description = "gRPC bindings to the CSI spec"
+name = "bpfman-csi"
+version = "1.8.0"
```
Billy99: Does this need `documentation.workspace = true`?

danielmellado: I think that's outdated and the current version doesn't have that anymore.

Billy99: You lost me. For all the other `Cargo.toml` files, `documentation.workspace = true` was added, but `csi/Cargo.toml` was the only one without it. I don't follow all the changes, just watching the pattern and noticed an anomaly. It's fine if it doesn't need it, just curious.

astoycos: So I intentionally left it out here because technically these bindings are independent of the bpfman project, i.e the crate could be used by anyone as rust CSI protobuf bindings.

Billy99: Good enough for me, just wanted to make sure it wasn't an oversight.

Issue #1018, opened 2024-03-05
Make the error message you get when bpfman times out more helpful


anfredette:
If a user tries to run a bpfman cli command while another instance of bpfman has the database locked,  they get the following message.

```
$ bpfman list
thread 'main' panicked at bpfman/src/main.rs:59:40:
Unable to open root database: Timed out
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'tokio-runtime-worker' panicked at /home/afredette/.cargo/registry/src/index.crates.io-6f17d22bba15001f/spin-0.5.2/src/once.rs:128:29:
Once has panicked
```

astoycos:
```
[INFO  bpfman] Database lock is already held, retrying after 2000 milliseconds
[INFO  bpfman] Database lock is already held, retrying after 2000 milliseconds
thread 'tokio-runtime-worker' panicked at bpfman/src/lib.rs:141:53:
Unable to open root database: DatabaseLockError
```

This is what it looks like now, I think it's good enough :) 

Additionally now that none of our processes hold the lock for an extended period of time this shouldn't be an issue anymore, I'm ready to close this if you are @anfredette

ZhangShuaiyi:
```
[INFO  bpfman::utils] Has CAP_BPF: true
[INFO  bpfman::utils] Has CAP_SYS_ADMIN: true
[INFO  bpfman] Database lock is already held, retrying after 1000 milliseconds
[INFO  bpfman] Database lock is already held, retrying after 1000 milliseconds
[INFO  bpfman] Database lock is already held, retrying after 1000 milliseconds
[INFO  bpfman] Database lock is already held, retrying after 1000 milliseconds
[INFO  bpfman] Database lock is already held, retrying after 1000 milliseconds
[INFO  bpfman] Database lock is already held, retrying after 1000 milliseconds
[INFO  bpfman] Database lock is already held, retrying after 1000 milliseconds
[INFO  bpfman] Database lock is already held, retrying after 1000 milliseconds
[INFO  bpfman] Database lock is already held, retrying after 1000 milliseconds
thread 'main' panicked at /root/go/src/github.com/bpfman/bpfman/bpfman/src/lib.rs:486:14:
Unable to open root database: DatabaseLockError
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrac
```
This is what it looks like in `481eee86`.  I think it can ouput the error message `Unable to open root database: DatabaseLockError` but no need to panic.

PR #1140, opened 2024-05-16
bpfman: don't panic if DatabaseLockError occurred

ZhangShuaiyi:
Currently `bpfman` will panic when DatabaseLockError occurred.
```
thread 'main' panicked at /root/go/src/github.com/bpfman/bpfman/bpfman/src/lib.rs:486:14:
Unable to open root database: DatabaseLockError
```
After this patch `bpfman` will print DatabaseLockError message and don't panic
```
Error: Failed to acquire database lock, please try again later
```
Refs: #1018

### DIFF
--- a/bpfman/src/lib.rs
+++ b/bpfman/src/lib.rs
@@ -479,12 +479,7 @@ fn get_programs_iter(root_db: &Db) -> impl Iterator<Item = (u32, Program)> + '_ 
 async fn setup() -> Result<(Config, Db), BpfmanError> {
     initialize_bpfman()?;
 
-    Ok((
-        open_config_file(),
-        init_database(get_db_config())
-            .await
-            .expect("Unable to open root database"),
-    ))
+    Ok((open_config_file(), init_database(get_db_config()).await?))
 }
 
 async fn add_multi_attach_program(


bpfman/src/lib.rs
```diff
@@ -481,9 +481,10 @@ async fn setup() -> Result<(Config, Db), BpfmanError> {
...
     Ok((
         open_config_file(),
-        init_database(get_db_config())
-            .await
-            .expect("Unable to open root database"),
+        init_database(get_db_config()).await.unwrap_or_else(|err| {
+            eprintln!("{}", err);
+            std::process::exit(1)
+        }),
     ))
```
anfredette, line 488: I suggest we return an error here instead of just exiting.  The following should work:

```suggestion
    Ok((open_config_file(), init_database(get_db_config()).await?))
```

anfredette (changes requested):
@ZhangShuaiyi, thank you very much for your contribution!  Your proposal would work, and is better than the current behavior.  However, we discussed this during the community meeting yesterday and thought it would be better to return an error than to just exit.  I made a suggestion.

ZhangShuaiyi:
@anfredette  Thank you for your suggestion. I've updated the patch.

anfredette (approved):
LGTM.  Thanks for your first contribution, @ZhangShuaiyi!

PR #1146, opened 2024-05-21
Have systemd-rpm-macros use bpfman service

danielmellado:
This PR changes the specfile so it uses the service directly for
systemd-rpm-macros. Otherwise we hit an issue in fedora-lint as this is
advised in the review [1]

[1] https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_scriptlets


Billy99 (approved):
LGTM

Issue #1145, opened 2024-05-21
DatabaseConfig max_retries not working as intended


anfredette:
The description for `max_retries` is "Maximum numbers of times to attempt to open the database after a failed attempt".  However, the way the for loop is implemented, bpfman actually makes max_retries-1 attempts.  For example, if you set it to 1, we try 0 times and fail immediately without trying.

PR #1147, opened 2024-05-21
Make DatabaseConfig max_retries work as defined.

anfredette:
The description for max_retries is "Maximum number of times to attempt to open the database after a failed attempt". However, the way the for loop is implemented, bpfman actually makes max_retries-1 attempts. For example, if you set it to 1, we try 0 times and fail immediately without trying.

Fixes: #1145


astoycos (approved):
Thanks for the catch @anfredette 👍

PR #1148, opened 2024-05-21
docs: update the RPM uninstall process

Billy99:
The RPM spec file is being updated to be compliant with Fedora (see #1146). One of the changes requires bpfman to be stopped before uninstalling the RPM, otherwise errors are logged and reinstall of the RPM fails.

Example Error:
```
May 21 10:41:49 nfvsdn-22-oot.lab.eng.rdu2.redhat.com systemd[1]: bpfman.socket: Socket unit configuration has changed while unit has been running, no open socket file descriptor left. The socket unit is not functional until restarted.
```

And subsequent RPM install fails and bpfman is not running:
```
$ sudo ./go-xdp-counter --iface eno3
2024/05/21 11:03:28 Using Input: Interface=eno3 Priority=50 Source=/home/bmcfall/src/bpfman/examples/go-xdp-counter/bpf_bpfel.o
2024/05/21 11:03:28 rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /run/bpfman-sock/bpfman.sock: connect: connection refused"
```

Updated the docs to indicate bpfman should be stopped.

### DIFF
--- a/docs/getting-started/running-rpm.md
+++ b/docs/getting-started/running-rpm.md
@@ -97,9 +97,12 @@ $ sudo rpm -qa | grep bpfman
 bpfman-0.4.0~dev-1.20240117143006587102.main.191.gda44a71.fc39.x86_64
 ```
 
-To uninstall the RPM:
+To stop bpfman and uninstall the RPM:
 
 ```console
+sudo systemctl stop bpfman.socket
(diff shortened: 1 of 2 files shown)


astoycos (approved):
LGTM

PR #1150, opened 2024-05-31
bpfman: add multiarch builds for bpfman binaries

Billy99:
Commit 1:
On each PR push, a set a builds and checks are performed on the codebase, which is managed by `.github/workflows/build.yml`. This PR updates this file to build all images for multiple architectures to make sure they build for each, but only tests on x86. While adding multiarch support, some changes were made to make the process faster and more
efficient.

Changes associated with this PR:
* arm64 build required openssl. It was already a dependancy in Cargo.lock, so it was added to Cargo.toml and bpfman/Cargo.toml.
* A matrix was to build each image for each architecture. bpfman won't build for ppc64le or s390x until AYA support for those archs is added, so they are currently commented out.
* Because build is a matrix, it means all steps in the job are run for each architecture. So steps that only need to run once (independent of the architecture), like lint and format, are only run in amd64 runs. Manpage and CLI Tab Completion output are pushed to artifacts so they can be used in integration tests.
* Building the eBPF code is not needed for the userspace binaries to build, so only needs to be built once (multiarch may need to follow here, but will be different). Output is pushed to to artifacts so it can be used in integration tests.
* Build the rust images for each architecture. amd64 debug build results are pushed to artifacts so it can be used in integration tests. All release build results are pushed to artifacts so they can be included in the release.
* Build the go images for each architecture. Go version was bumped everywhere from 1.21 to 1.22 but here because there were some linting issues.

Commit 2:
On every PR commit, all the container images managed by bpfman are built and pushed to Quay.io via .github/workflows/image-build.yaml. This PR updates this file to build all the Userspace containers as multiarch and
pushes the associated images and manifest files to Quay.io. This file also builds eBPF bytecode images. The multiarch builds for the bytecode images will follow later.

Changes:
* Break the Userspace builds into it's own job. They maybe recombined later when bytecode multiarch is implemented, so left some of the flags in place.
* Don't pull in go/rust tools into base VM because the Userspace containers build images within the container and pulled in there, so not needed in VM.
* Switched build of userspace from `redhat-actions/buildah-build@v2` to using `docker/build-push-action@v5`. docker populates TARGETARCH, TARGETOS and TARGETPLATFORM based on the what target is being built, which was needed in the bpfman build to install architecture specific tools. buildah doesn't pass any indication of target.
* docker and buildah both take a list of architectures (platforms) and loops internally through the architectures. bpfman can't currently build for ppc64le and s390x because AYA doesn't support those archs. All builds are only building amd64 and arm64 until AYA is patched. All other images were able to build for all archs in test runs.
* For the byecode builds, bumped go from 1.21 to 1.22.
* Remove bpf2go install, recent changes to eBPF source code changed the `generate go` command to include the bpf2go, so no longer need to install.
* Updated all the userspace container images to support multiarch.
* For bpfman image, don't need C eBPF bytecode for bpfman to build, so stripped relevant logic out. Added architecture specific commands.
* Updated go based images from 1.21 to 1.22 and removed bpf2go.


.github/workflows/image-build.yaml
```diff
@@ -175,6 +175,89 @@ jobs:
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Build and push
+        id: build-push-image
+        uses: docker/build-push-action@v5
+        with:
+          platforms: linux/amd64, linux/arm64
+          # platforms: linux/amd64, linux/arm64, linux/ppc64le, linux/s390x
```
astoycos, line 233: nit remove this comment maybe just add one saying "TBD linux/ppc64le, linux/s390x"

.github/workflows/build.yml
```diff
@@ -24,9 +24,28 @@ jobs:
         uses: apache/skywalking-eyes@cd7b195c51fd3d6ad52afceb760719ddc6b3ee91
 
   build:
+    strategy:
+      # amd64,arm64,ppc64le,s390x
+      fail-fast: false
```
astoycos, line 29: Does fail-fast default to true?

.github/workflows/build.yml
```diff
@@ -39,19 +58,34 @@ jobs:
             libssl-dev \
             gcc-multilib \
             libelf-dev \
-      - uses: actions/checkout@v4
-      - uses: actions/checkout@v4
+
+      - uses: taiki-e/setup-cross-toolchain-action@v1
```
astoycos, line 62: What does this action do? Does it distill down into just adding toolchains?

.github/workflows/build.yml
```diff
@@ -62,70 +96,157 @@ jobs:
       ## - uses: Swatinem/rust-cache@v2
 
       - name: Install cargo-llvm-cov
+        # Only need to run once
+        if: ${{ matrix.arch.arch == 'amd64' }}
```
astoycos, line 100: So we have this check everywhere... WDYT about having a separate workflow with all the new build steps that run first and handles ONLY the builds, then leave this existing one to just do everything else? 

This is kind of a nit, if it's too much of a PITA to deal with right now I understand

bpfman/Cargo.toml
```diff
@@ -53,6 +53,7 @@ oci-distribution = { workspace = true, default-features = false, features = [
     "native-tls",
     "trust-dns",
 ] }
+openssl = { workspace = true, features = ["vendored"] }
```
astoycos, line 56: Hrm I don't think we can rely on the vendored version of openssl for the RPM. see

https://github.com/bpfman/bpfman/pull/1023

@danielmellado Maybe you have an idea? 

I'm not quite sure why this is breaking us only on ARM

bpfman-operator/Makefile
```diff
@@ -109,6 +109,9 @@ ENVTEST ?= $(LOCALBIN)/setup-envtest
 CM_VERIFIER ?= $(LOCALBIN)/cm-verifier
 OPERATOR_SDK ?= $(LOCALBIN)/operator-sdk
 
+GOARCH ?= amd64
```
astoycos, line 112: Can you make this automatic based on the build system? 

i.e use the output of 
```
go env GOOS GOARCH
linux
amd64
```

examples/go-xdp-counter/container-deployment/Containerfile.go-xdp-counter
```diff
@@ -1,22 +1,29 @@
-FROM golang:1.21 as gocounter-build
+ARG BUILDPLATFORM=linux/amd64
```
astoycos, line 1: Does the makefile for the examples need to be updated? I guess we can do that in a follow up

Cargo.toml
```diff
@@ -55,6 +55,7 @@ netlink-packet-route = { version = "^0.19", default-features = false }
 netlink-sys = { version = "^0.8", default-features = false }
 nix = { version = "0.28", default-features = false }
 oci-distribution = { version = "0.10", default-features = false }
+openssl = { version = "0.10.29", default-features = false }
```
astoycos, line 58: see previus comment

astoycos (commented):
A bunch of questions and some comments, overall this looks good thanks for all the hard work 

1. Do we need that taki action? it seems to be only maintained by one person which worries me 
2. For the openssl dep I _think_ we could do something like 
```
[target.<architecture>]
pre-build = [
  "dpkg --add-architecture $CROSS_DEB_ARCH",
  "apt-get update && apt-get --assume-yes install libssl-dev:$CROSS_DEB_ARCH"
]
```
Rather than adding the vendored version of the crate since we need to be relying on the system's openssl version with the fedora packaging 

Otherwise the Golang stuff all looks good!

Billy99:
> Do we need that taki action? it seems to be only maintained by one person which worries me

Not sure which one you are referring too. grepped for `taki` and looked through all "actions/" and nothing caught my eye.

astoycos:
Sorry Taiki-e https://github.com/bpfman/bpfman/pull/1150/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R62

Billy99:
> Sorry Taiki-e https://github.com/bpfman/bpfman/pull/1150/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R62

I'll look for a replacement, but we were already using it in two places, I just added a 3rd.

.github/workflows/image-build.yaml
```diff
@@ -175,6 +175,89 @@ jobs:
...
 
+    name: Build Image (${{ matrix.image.image }})
+    steps:
+      - name: Checkout bpfman
+        uses: actions/checkout@v4
+
+      - name: Install cosign
+        uses: sigstore/cosign-installer@v3.5.0
+
+      - name: Generate olm bundle on disk
```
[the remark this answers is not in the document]
Billy99: I think there are several changes that are needed once we move to the new repo. Removing should be done in a separate PR.

.github/workflows/build.yml
Billy99: Yes. If one of the targets fail, the rest are cancelled. This allows us to continue with the other archs if one failed. Was definitely helpful when working on this. I think we still want it?

.github/workflows/build.yml
Billy99: That was my first implementation and it worked. But added ~10mins to the build. Things like lint and xtask commands build the code base,  so package install and builds are being duplicated.

.github/workflows/build.yml
Billy99: Looks like I was able to build without it, so I'll remove it.

bpfman/Cargo.toml
Billy99: Removed.

bpfman-operator/Makefile
Billy99: Done

Cargo.toml
Billy99: Done

.github/workflows/build.yml
Billy99: Done

.github/workflows/image-build.yaml
Billy99: Done

examples/go-xdp-counter/container-deployment/Containerfile.go-xdp-counter
Billy99: Yes, I missed that.

Done.

Billy99:
> > Sorry Taiki-e https://github.com/bpfman/bpfman/pull/1150/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R62
> 
> I'll look for a replacement, but we were already using it in two places, I just added a 3rd.

Were you worried about taiki-e/install-action in general, or just the `setup-cross-toolchain-action` (which I ended up removing)?

Billy99:
@astoycos I think I addressed or commented on everything. I'll squash the two commits that addressed your comments once you've had a look. The openssl change was a lot, so wanted to keep it separate until you had a look.

.github/workflows/build.yml
```diff
@@ -25,22 +25,25 @@ jobs:
...
   build:
     strategy:
-      # amd64,arm64,ppc64le,s390x
       fail-fast: false
       matrix:
         arch:
           - arch: amd64
             rust-target: x86_64-unknown-linux-gnu
             filename: linux-x86_64
+            command: cargo
```
astoycos, line 34: So we don't need cross here since we're building on an x86 system?

Billy99: Correct. That's how I understand it.

astoycos (approved):
LGTM

Thanks @Billy99 I'll keep and eye on the workflows after this merges

PR #1154, opened 2024-06-10
Add application object example

msherif1234:
Add application object example

```sh
make deploy-target
IMAGE_APP_BC="quay.io/mmahmoud/bpfman-bytecode/go-application-counter:all" make deploy-application

oc get bpfprograms
NAME                                                                                                            TYPE          STATUS         AGE
go-application-counter-example-kprobe-bpfman-deployment-control-plane-try-to-wake-up                            application   bpfmanLoaded   50s
go-application-counter-example-tc-bpfman-deployment-control-plane-eth0                                          application   bpfmanLoaded   40s
go-application-counter-example-tracepoint-bpfman-deployment-control-plane-syscalls-sys-enter-kill               application   bpfmanLoaded   45s
go-application-counter-example-uprobe-bpfman-deployment-control-plane--go-target-go-target-ds-psgp2-go-target   application   bpfmanLoaded   34s
go-application-counter-example-xdp-bpfman-deployment-control-plane-eth0                                         application   bpfmanLoaded   30s

```

### DIFF
--- a/examples/config/base/go-app-counter/bytecode.yaml
+++ b/examples/config/base/go-app-counter/bytecode.yaml
@@ -11,7 +11,7 @@ spec:
   bytecode:
     image:
       url: quay.io/bpfman-bytecode/go-application-counter:latest
-      imagepullpolicy: IfNotPresent
+      imagepullpolicy: Always
   programs:
     - type: Kprobe
       kprobe:
--- a/examples/config/base/go-app-counter/bytecode.yaml
+++ b/examples/config/base/go-app-counter/bytecode.yaml
@@ -1,0 +1,49 @@
+---
+apiVersion: bpfman.io/v1alpha1
+kind: BpfApplication
+metadata:
+  labels:
+    app.kubernetes.io/name: bpfapplication
+  name: go-application-counter-example
+spec:
+  # Select all nodes
+  nodeselector: {}
+  bytecode:
+    image:
+      url: quay.io/bpfman-bytecode/go-application-counter:latest
+      imagepullpolicy: IfNotPresent
+  programs:
+    - type: Kprobe
+      kprobe:
+        bpffunctionname: kprobe_counter
+        func_name: try_to_wake_up
+        offset: 0
+        retprobe: false
+    - type: Tracepoint
+      tracepoint:
+        bpffunctionname: tracepoint_kill_recorder
+        names: [syscalls/sys_enter_kill]
+    - type: TC
+      tc:
+        bpffunctionname: stats
+        interfaceselector:
+          primarynodeinterface: true
+        priority: 55
+        direction: ingress
+    - type: Uprobe
+      uprobe:
+        bpffunctionname: uprobe_counter
+        func_name: main.getCount
+        target: /go-target
+        retprobe: false
+        containers:
+          namespace: go-target
+          pods: {}
+          containernames:
+            - go-target
+    - type: XDP
+      xdp:
+        bpffunctionname: xdp_stats
+        interfaceselector:
+          primarynodeinterface: true
+        priority: 55
--- a/examples/config/base/go-app-counter/bytecode.yaml
+++ b/examples/config/base/go-app-counter/bytecode.yaml
@@ -4,7 +4,7 @@ kind: BpfApplication
 metadata:
   labels:
     app.kubernetes.io/name: bpfapplication
-  name: go-application-counter-example
+  name: app-counter
 spec:
   # Select all nodes
   nodeselector: {}
--- a/examples/go-app-counter/main.go
+++ b/examples/go-app-counter/main.go
@@ -4,8 +4,10 @@
 package main
 
 import (
+	"log"
 	"os"
 	"os/signal"
+	"sync"
 	"syscall"
 )
 
@@ -13,18 +15,26 @@ const (
 	DefaultByteCodeFile = "bpf_bpfel.o"
 )
 
+var initMutex = &sync.Mutex{}
+
 //go:generate go run github.com/cilium/ebpf/cmd/bpf2go -cc clang -no-strip -cflags "-O2 -g -Wall" bpf ./bpf/app_counter.c -- -I.:/usr/include/bpf:/usr/include/linux
 func main() {
 	stop := make(chan os.Signal, 1)
 	signal.Notify(stop, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
 
-	processKprobe(stop)
+	go processKprobe(stop)
 
-	processTracepoint(stop)
+	go processTracepoint(stop)
 
-	processTC(stop)
+	go processTC(stop)
 
-	processUprobe(stop)
+	go processUprobe(stop)
 
-	processXdp(stop)
+	go processXdp(stop)
+
+	//select {} // wait forever
+
+	<-stop
+
+	log.Printf("Exiting go-app-counter...\n")
 }
--- a/examples/go-app-counter/main.go
+++ b/examples/go-app-counter/main.go
@@ -1,0 +1,30 @@
+//go:build linux
+// +build linux
+
+package main
+
+import (
+	"os"
+	"os/signal"
+	"syscall"
+)
+
+const (
+	DefaultByteCodeFile = "bpf_bpfel.o"
+)
+
+//go:generate go run github.com/cilium/ebpf/cmd/bpf2go -cc clang -no-strip -cflags "-O2 -g -Wall" bpf ./bpf/app_counter.c -- -I.:/usr/include/bpf:/usr/include/linux
+func main() {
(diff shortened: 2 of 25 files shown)


examples/go-app-counter/go-app-counter
astoycos, line 1: We shouldn't track the binary right?

astoycos (commented):
So overall the idea is solid, I really don't like having to copy each bpf program AND userspace go program into the `examples/go-app-counter/` dir... in the future if any of them change now we have to update them twice 😨 I wonder if we could change the organization of the examples to make this easier 🤔 WDYT @Billy99?

At the very least we would need  an issue to let folks know about all the dubplication

examples/config/base/go-app-counter/bytecode.yaml
```diff
@@ -0,0 +1,49 @@
+---
+apiVersion: bpfman.io/v1alpha1
+kind: BpfApplication
+metadata:
+  labels:
+    app.kubernetes.io/name: bpfapplication
+  name: go-application-counter-example
+spec:
+  # Select all nodes
+  nodeselector: {}
```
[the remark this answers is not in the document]
msherif1234: I didn't have global because none of the current examples had global data and wanted to leverage all existing programs as is for the app

examples/go-app-counter/go-app-counter
msherif1234: right will remove

msherif1234:
> So overall the idea is solid, I really don't like having to copy each bpf program AND userspace go program into the `examples/go-app-counter/` dir... in the future if any of them change now we have to update them twice 😨 I wonder if we could change the organization of the examples to make this easier 🤔 WDYT @Billy99?
> 
> At the very least we would need an issue to let folks know about all the duplication

I agree that was the quickest way for me to spin an app with all ebpf hooks to test the new object, there is a slight renameing and diff for the combined case but yeah the vast majority is teh same but at the end this is more test codedoesn't have to be super clean I think ?

examples/config/base/go-app-counter/bytecode.yaml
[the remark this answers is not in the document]
astoycos: Ok that's alright for now but we need test coverage to ensure global works correctly across multiple programs (it should). 

Today we only have global data tests at the integration test level https://github.com/bpfman/bpfman/blob/main/tests/integration-test/src/tests/e2e.rs#L67

But we don't test the kubernetes flow

Billy99:
> So overall the idea is solid, I really don't like having to copy each bpf program AND userspace go program into the `examples/go-app-counter/` dir... in the future if any of them change now we have to update them twice 😨 I wonder if we could change the organization of the examples to make this easier 🤔 WDYT @Billy99?

Yes, not ideal to duplicate all the example logic under go-app-counter. One option is just to create a separate set of eBPF programs that the go-app-counter runs. We could make it simpler, removing all the CLI logic and it is only callable from K8s. The other option is to move the bulk of the code into pkg/config-mgmt directory and call from each example program. That might make the examples even harder to follow though.

Billy99:
> but yeah the vast majority is teh same but at the end this is more test codedoesn't have to be super clean I think ?

My only counter to this statement is the examples are there for two fold. They are primarily there to give others examples on how to interact with bpfman and how samples of each program type. Since we already had it, we also use them for test coverage.

Billy99:
We can write an issue to come back and figure out a way to not duplicate all the code. Just need to make it Urgent and make sure it doesn't get pushed out.

Billy99 (commented):
You need to go into the Settings TAB for:
https://quay.io/repository/bpfman-bytecode/go-application-counter
https://quay.io/repository/bpfman-userspace/go-application-counter

And give the Robot Account github_actions Write permission on the images. That way github can push updates images. Look at one of other images for examples.

examples/go-app-counter/main.go
```diff
@@ -0,0 +1,30 @@
...
+const (
+	DefaultByteCodeFile = "bpf_bpfel.o"
+)
+
+//go:generate go run github.com/cilium/ebpf/cmd/bpf2go -cc clang -no-strip -cflags "-O2 -g -Wall" bpf ./bpf/app_counter.c -- -I.:/usr/include/bpf:/usr/include/linux
+func main() {
+	stop := make(chan os.Signal, 1)
+	signal.Notify(stop, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
+
+	processKprobe(stop)
```
anfredette, line 21: Maybe I'm missing something, but the way this seems to work is that we start the Kprobe user space code, but never get to the rest of the userspace functions.  Something would need to signal processKprobe to stop before it would continue to processTracepoint, right?  Could we run each in a separate thread simultaneously?  Or better yet, combine them into one user space app that cycles through the different maps and logs the count for each.

anfredette (changes requested):

examples/go-app-counter/main.go
msherif1234: I started that way when there were only 2 programs but since u suggested on the other PR cover all or most the userspace would require more work to include all

anfredette:
> We can write an issue to come back and figure out a way to not duplicate all the code. Just need to make it Urgent and make sure it doesn't get pushed out.

Each of these examples stands on it's own, so even if Mohamed starts out by duplicating code, that doesn't mean that we need to keep it in sync if someone decides to change one of the other applications.

anfredette:
> We could make it simpler, removing all the CLI logic and it is only callable from K8s.

I think this is a good idea.  We really should have some k8s-only examples, and maybe at least one cli-only example to make it easier for the user to see the minimum needed for a given environment.  We might also include non-Kustomized yaml files (also to make it easier to see the minimum required).

examples/go-app-counter/main.go
anfredette: How do you get them all to run, though?

anfredette: I.e., if you did
```
	go processKprobe(stop)
	go processTracepoint(stop)
	go processTC(stop)
	go processUprobe(stop)
	go processXdp(stop)
```
Would they all run in parallel?

msherif1234: how best I can test the userspace I think all my validation was done for bc ?

examples/go-app-counter/main.go
```diff
@@ -0,0 +1,324 @@
...
+				paramData.ProgId = uint(kernelInfo.GetId())
+			} else {
+				conn.Close()
+				log.Printf("kernelInfo not returned in LoadResponse")
+				return nil, action, err
+			}
+			log.Printf("Program registered with id %d\n", paramData.ProgId)
+
+			// 2. Set up defer to unload program when this is closed
+			defer func(id uint) {
```
anfredette, line 283: Having this here unloads the program as soon as this function exits, so it doesn't work.  The deferred close needs to be in the function that runs the get stats loop or after it exits.

examples/go-app-counter/main.go
```diff
@@ -0,0 +1,324 @@
...
+
+			// 3. Get access to our map
+			mapPath, err = configMgmt.CalcMapPinPath(res.GetInfo(), mapIndex)
+			if err != nil {
+				log.Print(err)
+				return nil, action, err
+			}
+		} else {
+			// 2. Set up defer to close connection
+			defer func(id uint) {
```
anfredette, line 302: I think this deferred connection close will be a problem too.

msherif1234: yeah I agree for the above two comments the defer needs to move up one level not in the common function I think

examples/go-app-counter/main.go
```diff
@@ -0,0 +1,124 @@
...
+	log.Printf("Waiting for all goroutines to finish...\n")
+	wg.Wait()
+
+	log.Printf("Exiting go-app-counter...\n")
+}
+
+func getMapPinPath(progId uint, map_name string) (string, error) {
+	appMutex.Lock()
+	defer appMutex.Unlock()
+	return configMgmt.RetrieveMapPinPath(ctx, c, progId, map_name)
```
msherif1234, line 111: why do u need a mutex for reading the fs mount vol ?

msherif1234:
I guess u couldn't find a way to use the common API I had b4 while handle the defers per prog type ? I am fine with the changes anyway

examples/go-app-counter/main.go
anfredette: The mutexes on the new functions are to prevent all the threads from calling bpfman at the same time.  When they do that, some of the calls time out because it's blocked executing another request.  I was thinking of an alternative with a retry and a random sleep time, but thought I'd push it this way since it's working.

msherif1234:
lets add patch.yaml back here as well

anfredette (approved):
LGTM

astoycos (approved):
LGTM

PR #1141, opened 2024-05-16
Update image spec to support multiple images, add build tooling for bytecode images

astoycos:
start working on a design document which will
allow eBPF bytecode images to run on on avaliable
linux architecture.

PR #1155, opened 2024-06-11
Temporary work-around for multi-program container images

anfredette:
The BpfApplication CRD support is intended to allow an application to define a set of BPF programs required for operation.  This set of programs may included multiple types of BPF programs that are packaged in a single .o and single bytecode image file.  However, our currentl image spec is designed to support a single bpf program with a single type.

More importantly, the bpfman code has a check to validate that the name of program being loaded matches the name that is provided by the user.

As a temporary work-around, this pr just logs the mismatch and doesn't fail the load.

A more complete solution will be provided by pr #1141

### DIFF
--- a/bpfman/src/types.rs
+++ b/bpfman/src/types.rs
@@ -12,7 +12,7 @@ use std::{
 use aya::programs::ProgramInfo as AyaProgInfo;
 use chrono::{prelude::DateTime, Local};
 use clap::ValueEnum;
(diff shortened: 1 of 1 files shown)


astoycos (approved):
I'm alright with this as an intermediate fix prior to https://github.com/bpfman/bpfman/pull/1141 landing!! 

Thanks @anfredette

PR #1161, opened 2024-06-18
examples: cleanup patch.yaml files

Billy99:
In the examples directory, the patch.yaml files were never meant to be checked in. They are generated from the patch.yam.env files in the Makefile when `make deploy` is called. They were inadvertently checked in as part of the rename from bpfd to bpfman where lots of files were modified.
    
Once they were checked in, a change was made to remove quotes around the path value and `---` added to the top of the file. So these changes were propagated to the patch.yaml.env file.
    
Makefile needed updates to load from tags.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -189,13 +189,15 @@ deploy-prog: kustomize
 ifndef TAG
 	sed 's@URL_BC@$(IMAGE_BC)@' config/$(CONFIG_DIR)/patch.yaml.env > config/$(CONFIG_DIR)/patch.yaml
 	cd config/$(CONFIG_DIR) && $(KUSTOMIZE) edit set image quay.io/bpfman-userspace/$(PROG_NAME)=${IMAGE_US}
+	$(eval KUST_DIR=$(CONFIG_DIR))
 else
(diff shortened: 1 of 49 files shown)


examples/Makefile
```diff
@@ -189,13 +189,15 @@ deploy-prog: kustomize
 ifndef TAG
 	sed 's@URL_BC@$(IMAGE_BC)@' config/$(CONFIG_DIR)/patch.yaml.env > config/$(CONFIG_DIR)/patch.yaml
 	cd config/$(CONFIG_DIR) && $(KUSTOMIZE) edit set image quay.io/bpfman-userspace/$(PROG_NAME)=${IMAGE_US}
+	$(eval KUST_DIR=$(CONFIG_DIR))
 else
-	$(eval KUST_DIR=$(TAG))
```
Billy99, line 193: @anfredette This look OK to you? KUST_DIR was being set but not used. So loading from a release, like `TAG=v0.4.1 make deploy` was just loading latest.

astoycos (approved):
LGTM I'll let @anfredette Remove the hold

Thanks @Billy99!

anfredette (approved):
LGTM

PR #1164, opened 2024-06-21
migrate the bpfman-operator

astoycos:
For future ease of use and maintainability the
bpfman-operator has been moved into its own
github repository at https://github.com/bpfman/bpfman-operator.

All further operator development should continue
there.

See the upstream [k8s slack conversation](https://kubernetes.slack.com/archives/C04UJBW2553/p1719324083855869) for approval from Bob Killen (CNCF POC)


dave-tucker (approved):

Billy99 (commented):
Doesn't this need to change?
https://github.com/bpfman/bpfman/blob/main/.github/workflows/docs-build.yaml#L32

astoycos:
Yeah I think we do, good catch!

Billy99 (commented):
Another spot to take a look at:
https://github.com/bpfman/bpfman/blob/main/scripts/lint.sh#L26

astoycos:
I'm going to have to move the API document generation into the core repo I think

astoycos:
Ok @Billy99 I fixed it ptal at the most recent commit

astoycos:
Goes with https://github.com/bpfman/bpfman-operator/pull/34

astoycos:
> Another spot to take a look at: https://github.com/bpfman/bpfman/blob/main/scripts/lint.sh#L26 https://github.com/bpfman/bpfman/blob/main/scripts/verify-golint.sh#L33

I think these two scripts should be condensed into one

Billy99:
> I think these two scripts should be condensed into one... Done :)

Looks like `cargo xtask lint` is calling this script, but should `cargo xtask lint` be called from CI?

astoycos:
> > I think these two scripts should be condensed into one... Done :)
> 
> Looks like `cargo xtask lint` is calling this script, but shouldn't `cargo xtask lint` be called from CI?

All these steps individually are already happening in CI however we've split our workflow steps up by language so doing it all in one command wouldn't work at the moment 

I agree though it'd be better to run this command in CI rather than executing this 🤔

Billy99 (approved):
/lgtm

anfredette (approved):
lgtm

PR #1168, opened 2024-06-21
examples: Add patch.yaml file back

Billy99:
The patch.yaml files were removed from the examples/config directory in PR#1161. They were removed because they are generated by the Makefile. However, bpfman-operator/tests/integration calls Kustomize directly and is not going through the Makefile. So add them back.


msherif1234 (approved):

astoycos (approved):
LGTM 

Good catch thanks @Billy99!

PR #1183, opened 2024-07-08
only push dispatcher tag when building from main

astoycos:
Before this was always building/pushing the dispatcher image tags regardless of the branch. This change makes sure we only update the main tags when building from main.


msherif1234:
LGTM

msherif1234 (approved):

anfredette (approved):
LGTM

PR #1186, opened 2024-07-10
bpfman/lib: add function documentation

frobware:
This pull request documents several functions in the bpfman crate, with examples marked with `rust,no_run` to ensure proper compilation without execution.

Fixes: https://github.com/bpfman/bpfman/issues/1050

#### Rationale for `rust,no_run` Attribute

- **Compile-Time Checks**: By using `rust,no_run`, the examples are still compiled when running `cargo test --doc`. This ensures that the code snippets are syntactically correct and that any changes to the API will be caught at compile time.
- **Avoiding Runtime Dependencies**: The examples often require a running instance of `bpfman` and specific system conditions (e.g., loaded eBPF programs). These conditions are not guaranteed in a typical documentation test environment. Using `rust,no_run` prevents these examples from failing due to unmet runtime dependencies.
- **Safety and Stability**: Some functions can have side effects or require specific system states. By preventing the execution of these examples, we avoid potential issues such as modifying system state, interfering with running services, or causing unintended side effects during testing.
- **Focus on Usage Clarity**: The examples are intended to illustrate how to use the functions correctly. Using `rust,no_run` ensures that the focus remains on demonstrating the correct usage patterns without the need for setting up the entire runtime environment required for execution.

By compiling but not running the examples, we strike a balance between ensuring code correctness and avoiding the complexity and potential instability of executing environment-dependent code during documentation tests.

Tested with:

```sh
% cargo test --doc
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.19s
   Doc-tests bpfman

running 6 tests
test bpfman/src/lib.rs - pull_bytecode (line 633) - compile ... ok
test bpfman/src/lib.rs - remove_program (line 316) - compile ... ok
test bpfman/src/lib.rs - add_program (line 132) - compile ... ok
test bpfman/src/lib.rs - get_program (line 549) - compile ... ok
test bpfman/src/lib.rs - list_programs (line 439) - compile ... ok
test bpfman/src/types.rs - types::ProgramData::new (line 373) ... ok
```

### CI Integration
Documentation examples are now compiled in CI to ensure syntax correctness and catch API changes early.

### DIFF
--- a/bpfman/src/lib.rs
+++ b/bpfman/src/lib.rs
@@ -99,7 +99,98 @@ pub(crate) fn get_db_config() -> SledConfig {
     SledConfig::default().temporary(true)
 }
 
-/// Loads an ebpf program.
+/// Adds an eBPF program to the system.
+///
+/// This function takes a `Program` and performs the necessary loading
+/// and attaching operations to add it to the system. It supports
+/// various types of eBPF programs such as XDP, TC, TCX, Tracepoint,
+/// Kprobe, Uprobe, Fentry, and Fexit. The program can be added from a
+/// locally built bytecode file or a remote bytecode image. If the
+/// program is successfully added, it returns the updated [`Program`];
+/// otherwise, it returns a `BpfmanError`.
+///
+/// # Arguments
+///
+/// * `program` - The eBPF program to be added.
+///
+/// # Returns
+///
+/// * `Ok(Program)` - If the program is successfully added.
+/// * `Err(BpfmanError)` - If there is an error during the process.
+///
+/// # Example
+///
+/// ```rust,no_run
+/// use bpfman::add_program;
+/// use bpfman::errors::BpfmanError;
+/// use bpfman::types::{KprobeProgram, Location, Program, ProgramData};
+/// use std::collections::HashMap;
+///
+/// #[tokio::main]
+/// async fn main() -> Result<(), BpfmanError> {
+///     // Define the location of the eBPF object file.
+///     let location = Location::File(String::from("kprobe.o"));
+///
+///     // Metadata is used by the userspace program to attach additional
+///     // information (similar to Kubernetes labels) to an eBPF program
+///     // when it is loaded. This metadata consists of key-value pairs
+///     // (e.g., `owner=acme`) and can be used for filtering or selecting
+///     // programs later, for instance, using commands like `bpfman list
+///     // --metadata-selector owner=acme`.
+///     let mut metadata = HashMap::new();
+///     metadata.insert("owner".to_string(), "acme".to_string());
+///
+///     // Optionally, initialise global data for the program. Global data
+///     // allows you to set configuration values that are applied at runtime
+///     // during the loading of the eBPF program. The keys in global_data must
+///     // already exist in the eBPF program, otherwise, an error
+///     // will be encountered. For example:
+///     let mut global_data = HashMap::new();
+///     global_data.insert("global_counter".to_string(), vec![0; 8]);
+///
+///     // Optionally specify the owner of eBPF maps as a UID (e.g.,
+///     // Some(1001)).
+///     let map_owner_id = None;
+///
+///     // Create the program data with the specified location, name,
+///     // metadata, global data, and optional map owner ID. ProgramData
+///     // holds all necessary information for an eBPF program.
+///     let program_data = ProgramData::new(location, String::from("kprobe_do_sys_open"), metadata, global_data, map_owner_id)?;
+///
+///     // Create a kprobe program with the specified function name,
+///     // offset, and options. This sets up a probe at a specific point
+///     // in the kernel function named "do_sys_open".
+///     let probe_offset: u64 = 0;
+///     let is_retprobe: bool = false;
+///     let container_pid: Option<i32> = None;
+///     let kprobe_program = KprobeProgram::new(program_data, String::from("do_sys_open"), probe_offset, is_retprobe, container_pid)?;
+///
+///     // Add the kprobe program using the bpfman manager.
+///     let added_program = add_program(Program::Kprobe(kprobe_program)).await?;
+///
+///     // Print a success message with the name of the added program.
+///     println!("Program '{}' added successfully.", added_program.get_data().get_name()?);
+///     Ok(())
+/// }
+/// ```
+///
+/// # Errors
+///
+/// This function will return an error if:
+/// * The setup or initialization steps fail.
+/// * The program data fails to load.
+/// * The map owner ID is invalid or setting the map pin path fails.
+/// * The program bytes fail to set.
+/// * Adding the program (multi-attach or single-attach) fails.
+/// * The program is unsupported.
+///
+/// In case of failure, any created directories or loaded programs
+/// will be cleaned up to maintain system integrity.
+///
+/// # Asynchronous Operation
+///
+/// This function is asynchronous and should be awaited in an
+/// asynchronous context.
 pub async fn add_program(mut program: Program) -> Result<Program, BpfmanError> {
     let (config, root_db) = &setup().await?;
     let mut image_manager = init_image_manager().await;
@@ -166,7 +257,56 @@ pub async fn add_program(mut program: Program) -> Result<Program, BpfmanError> {
     }
 }
 
-/// Unloads and ebpf program.
+/// Removes an eBPF program specified by its ID.
+///
+/// This function attempts to remove an eBPF program that has been
+/// previously loaded by the `bpfman` tool. It performs the necessary
+/// cleanup and removal steps based on the type of program (e.g., XDP,
+/// Tc, Tracepoint, Kprobe, etc.).
+///
+/// # Arguments
+///
+/// * `id` - A `u32` kernel allocated value that uniquely identifies the eBPF program to be removed.
+///
+/// # Returns
+///
+/// * `Result<(), BpfmanError>` - Returns `Ok(())` if the program is successfully
+///   removed, or a `BpfmanError` if an error occurs during the removal process.
+///
+/// # Errors
+///
+/// This function will return a `BpfmanError` in the following cases:
+///
+/// * The program with the given ID does not exist or was not created
+///   by `bpfman`.
+/// * The program with the given ID is currently in use and cannot be
+///   deleted.
+/// * The program with the given ID has dependent resources that must
+///   be deleted first.
+/// * The user does not have sufficient permissions to delete the program.
+/// * The program type specified is invalid or unsupported for deletion.
+/// * The deletion operation encountered a database error.
+/// * The deletion operation was aborted due to a system timeout or
+///   interruption.
+///
+/// # Example
+///
+/// ```rust,no_run
+/// use bpfman::remove_program;
+///
+/// #[tokio::main]
+/// async fn main() {
+///     match remove_program(42).await {
+///         Ok(()) => println!("Program successfully removed."),
+///         Err(e) => eprintln!("Failed to remove program: {:?}", e),
+///     }
+/// }
+/// ```
+///
+/// # Asynchronous Operation
+///
+/// This function is asynchronous and should be awaited in an
+/// asynchronous context.
 pub async fn remove_program(id: u32) -> Result<(), BpfmanError> {
     let (config, root_db) = &setup().await?;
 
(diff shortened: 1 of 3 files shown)


bpfman/src/types.rs
```diff
@@ -398,6 +491,16 @@ impl ProgramData {
         )
     }
 
+    /// Retrieves the kind of program.
```
astoycos, line 494: as specified by the (ProgramType)[https://github.com/bpfman/bpfman/blob/main/bpfman/src/types.rs#L1534] structure

bpfman/src/types.rs
```diff
@@ -414,6 +527,16 @@ impl ProgramData {
         sled_insert(&self.db_tree, ID, &id.to_ne_bytes())
     }
 
+    /// Retrieves the ID of the program.
```
astoycos, line 530:
```suggestion
    /// Retrieves the kernel ID of the program.
```

bpfman/src/types.rs
```diff
@@ -449,6 +572,16 @@ impl ProgramData {
         })
     }
 
+    /// Retrieves the location of the program.
```
astoycos, line 575:
```suggestion
    /// Retrieves the location of the program. The Location of the program describes if the program's bytecode was provided via a fully qualified local path OR via an OCI compliant container image tag.
```

Or something like that

bpfman/src/types.rs
```diff
@@ -653,6 +836,16 @@ impl ProgramData {
      * Methods for setting and getting kernel information.
      */
 
+    /// Retrieves the name of the kernel.
```
astoycos, line 839:
```suggestion
    /// Retrieves the name of the program.
```

Sorry this is a bit confusing but please see the API documentation here https://docs.aya-rs.dev/aya/programs/struct.programinfo for all these fields

All APIs with the "kernel" substring in the name are just pulling from the aya ProgramInfo structure

astoycos (commented):
Looking good thanks for the detailed work here @frobware!!!

I think overall we might be able to move _some_ of the documentation into the docs themselves.

Additionally it'd be great if we got a preview of our rust docs on each PR like aya does https://github.com/aya-rs/aya/blob/main/netlify.toml It looks like netlify handles the lifting there, but I'm not 100% how it works (if that's too much work don't worry about it) 

Otherwise just some small comments, thanks again!

bpfman/src/lib.rs
```diff
@@ -93,7 +93,126 @@ pub(crate) fn get_db_config() -> SledConfig {
...
+/// Adds an eBPF program to the system.
+///
+/// This function takes a `Program` and performs the necessary setup
+/// and loading operations to add it to the system. It supports
+/// various types of eBPF programs such as XDP, TC, Tracepoint,
+/// Kprobe, Uprobe, Fentry, and Fexit. If the program is successfully
+/// added, it returns the updated `Program`; otherwise, it returns a
+/// `BpfmanError`.
+///
+/// The following steps are performed to integrate the eBPF program
```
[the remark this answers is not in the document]
frobware: > These are great docs....Open question though does it belong here or in our developer documentation?

Yeah, wasn't sure either. I attempted to be a client of the API by writing a simple program (which is where I derived the example code from for the docs) and recorded my understanding. From a callers perspective maybe some of the details are not important.

bpfman/src/lib.rs
```diff
@@ -217,7 +385,96 @@ pub async fn remove_program(id: u32) -> Result<(), BpfmanError> {
...
+/// * `debug!("BpfManager::list_programs()")` - Logged at the start of
+///   the function to indicate that the `list_programs` function has
+///   been called. This is only logged in debug builds or if the log
+///   level is set to `debug` or lower.
+///
+/// * `warn!("Unable to set kernel info for prog {prog_id}, error: {e}")` -
+///   Logged if there is an error while setting kernel information for
+///   a program. This is logged in both debug and release builds and
+///   helps in identifying issues with retrieving kernel information
+///   without stopping the listing of other programs.
```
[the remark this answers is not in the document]
frobware: As a caller of an API I would be surprised if logging (to stdout/stderr) takes place. As the caller I'd like to decide whether logging occurs or not. I just noticed logging could occur as I tried the API.

bpfman/src/types.rs
frobware: Just to be clear: The preference is for the text to read:

"Retrieves the kind of program as specified by [the] (ProgramType)[https://github.com/bpfman/bpfman/blob/main/bpfman/src/types.rs#L1534]"?

bpfman/src/types.rs
frobware: One thing I noted when viewing the converted documentation was that the first line shouldn't be too long. How about?

/// Retrieves the location of the program.
/// 
/// The location describes whether the program's bytecode was provided 
/// via a fully qualified local path or an OCI-compliant container image tag.

bpfman/src/lib.rs
```diff
@@ -93,7 +93,126 @@ pub(crate) fn get_db_config() -> SledConfig {
...
+///     metadata.insert("cpu_usage_threshold".to_string(), "80".to_string());
+///     metadata.insert("sampling_rate".to_string(), "1000".to_string());
+///     metadata.insert("enable_debug".to_string(), "true".to_string());
+///
+///     // Initialise global data for the program. Global data holds
+///     // variables that can be shared across multiple eBPF programs,
+///     // such as counters or aggregated statistics. The keys in
+///     // global_data must already exist in the eBPF program, otherwise,
+///     // an error will be encountered. For example:
+///     // global_data.insert("global_counter".to_string(), vec![0; 8]);
```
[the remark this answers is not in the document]
frobware: Will take another pass over text here w.r.t. global data for my own understanding.

bpfman/src/types.rs
frobware: Well, not with the line number because that will obviously change over time.

anfredette:
This is failing the DCO check because two commits aren't signed off.

frobware:
> This is failing the DCO check because two commits aren't signed off.

Fixed in https://github.com/bpfman/bpfman/compare/ddd7f0dcdb815e6af7909f137e8632e21baa1810..1d9ed28ebe1ba8ff5ba6907161336f0dc34df2c4. Thanks!

Billy99 (commented):
Thanks, great work.

Do we need to add `cargo test --doc` to CI?

bpfman/src/lib.rs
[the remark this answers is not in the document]
frobware: Having played around with the logging I see that it applies to other components too (e.g., h2, hickory*, etc). I will remove my commentary and move it to the developer guide, if it's not already covered.

bpfman/src/lib.rs
```diff
@@ -93,7 +93,126 @@ pub(crate) fn get_db_config() -> SledConfig {
...
+///     // configurations or properties related to the eBPF program, such
+///     // as threshold values, sampling rates, specific flags (e.g.,
+///     // enabling debug mode), and other parameters. These values can be
+///     // used by the userspace program.
+///     let mut metadata = HashMap::new();
+///     metadata.insert("cpu_usage_threshold".to_string(), "80".to_string());
+///     metadata.insert("sampling_rate".to_string(), "1000".to_string());
+///     metadata.insert("enable_debug".to_string(), "true".to_string());
+///
+///     // Initialise global data for the program. Global data holds
```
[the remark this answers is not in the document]
frobware: Updated in 6f070d1108586357738805ae9669f5f6cec6fe7c.

bpfman/src/lib.rs
```diff
@@ -93,7 +93,126 @@ pub(crate) fn get_db_config() -> SledConfig {
...
+///     // variables that can be shared across multiple eBPF programs,
+///     // such as counters or aggregated statistics. The keys in
+///     // global_data must already exist in the eBPF program, otherwise,
+///     // an error will be encountered. For example:
+///     // global_data.insert("global_counter".to_string(), vec![0; 8]);
+///     let global_data = HashMap::new();
+///
+///     // Summary:
+///     //
+///     // Metadata is used by the userspace program and does not need to
```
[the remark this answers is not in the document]
frobware: Updated in c5bdb296640c70f66222d6e208b6fc499e781ba9.

bpfman/src/types.rs
frobware: Updated in db1fb0d3a133b0da7411f409e2ab5d434b4b3805.

bpfman/src/lib.rs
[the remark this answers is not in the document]
frobware: Removed the detailed steps in 7a52e6c7b40c4e2398393ee192fc00808cc21538.

bpfman/src/lib.rs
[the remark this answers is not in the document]
frobware: Updated in 6f070d1108586357738805ae9669f5f6cec6fe7c.

bpfman/src/lib.rs
```diff
@@ -93,7 +93,126 @@ pub(crate) fn get_db_config() -> SledConfig {
...
-/// Loads an ebpf program.
+/// Adds an eBPF program to the system.
+///
+/// This function takes a `Program` and performs the necessary setup
+/// and loading operations to add it to the system. It supports
+/// various types of eBPF programs such as XDP, TC, Tracepoint,
+/// Kprobe, Uprobe, Fentry, and Fexit. If the program is successfully
+/// added, it returns the updated `Program`; otherwise, it returns a
+/// `BpfmanError`.
+///
```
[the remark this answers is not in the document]
frobware: Updated in ba1a4b461e6e7a6deca5a91fc4338244a1f8db85.

bpfman/src/types.rs
frobware: Done in 1cd04b01359f2f5a27acdf9947c20c00981f691b.

bpfman/src/lib.rs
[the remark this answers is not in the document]
frobware: Removed commentary about logging in 5919461387f8ce1daeffd52cf682bd8f29f33e09

frobware:
> Thanks, great work.
> 
> Do we need to add `cargo test --doc` to CI?

Yes, I think we should. Without it, there's a risk of drift unless everyone consistently runs the command before committing

frobware:
I pushed a lot of commits to address review comments. I think once we're happy with the content we should squash them back down to 3 or 4 commits.

/hold

frobware:
> Thanks, great work.
> 
> Do we need to add `cargo test --doc` to CI?

I added to the github workflow in 109d9419b64516b13d8e6067270cef22c3b7cf23.

frobware:
@Billy99 Thanks for the review. I've pushed changes for all suggestions from both you and @astoycos. Once we're satisfied with the content, I'll squash the commits down to four discrete ones.

The additional CI GitHub workflow that compiles the examples is expected to fail, and intentionally for the moment until I push:

```diff
commit 247aaeb8dc99aee1f0d49e2c2e215b6f329c1407 (HEAD -> 1050-add-doc-comments-and-examples)
Author: Andrew McDermott <amcdermo@redhat.com>
Date:   Wed Aug 14 15:43:37 2024 +0100

    Fix: cargo test --doc --all-features

diff --git a/bpfman/src/lib.rs b/bpfman/src/lib.rs
index 050f184e..ad518b45 100644
--- a/bpfman/src/lib.rs
+++ b/bpfman/src/lib.rs
@@ -140,8 +140,8 @@ pub(crate) fn get_db_config() -> SledConfig {
 ///     // during the loading of the eBPF program. The keys in global_data must
 ///     // already exist in the eBPF program's BPF maps, otherwise, an error
 ///     // will be encountered. For example:
+///     let mut global_data = HashMap::new();
 ///     global_data.insert("global_counter".to_string(), vec![0; 8]);
-///     let global_data = HashMap::new();
 ///
 ///     // Summary:
 ///     //
```

I haven't pushed this change yet because I want to first confirm that the GitHub workflow fails as expected before pushing a compilable example.

bpfman/src/lib.rs
```diff
@@ -93,7 +93,116 @@ pub(crate) fn get_db_config() -> SledConfig {
     SledConfig::default().temporary(true)
 }
 
-/// Loads an ebpf program.
+/// Adds an eBPF program to the system.
+///
+/// This function takes a `Program` and performs the necessary loading
+/// and attaching operations to add it to the system. It supports
+/// various types of eBPF programs such as XDP, TC, Tracepoint,
```
Billy99, line 100: Globally, go ahead and add TCX everywhere you mention TC, because it's coming and code has already started, so they will have missed your changes when grepping through the code for TC.

bpfman/src/lib.rs
```diff
@@ -93,7 +93,116 @@ pub(crate) fn get_db_config() -> SledConfig {
...
+///     // programs later, for instance, using commands like `bpfman list
+///     // --metadata-selector owner=acme`. Although initially designed to
+///     // assist with Kubernetes deployments, its use may vary depending
+///     // on the specific implementation.
+///     // the eBPF program.
+///     //
+///     // Global data is shared between the userspace program and the
+///     // eBPF program. The keys must be defined in the eBPF program's
+///     // BPF maps because the eBPF program directly interacts with these
+///     // data structures.
```
Billy99, line 161: You describe `global data` above so I don't think it is needed here, plus, this description is wrong.

bpfman/src/lib.rs
```diff
@@ -93,7 +93,116 @@ pub(crate) fn get_db_config() -> SledConfig {
...
+///
+///     // Initialise metadata for the program. Metadata contains custom
+///     // configurations or properties related to the eBPF program, such
+///     // as threshold values, sampling rates, specific flags (e.g.,
+///     // enabling debug mode), and other parameters. These values can be
+///     // used by the userspace program.
+///     let mut metadata = HashMap::new();
+///     metadata.insert("cpu_usage_threshold".to_string(), "80".to_string());
+///     metadata.insert("sampling_rate".to_string(), "1000".to_string());
+///     metadata.insert("enable_debug".to_string(), "true".to_string());
```
Billy99, line 136: Is this chunk of code needed (the metadata.insert() calls)? I'm not sure what it's purpose is? Plus, the description is wrong, the correct text is below.

bpfman/src/lib.rs
frobware: OK to drop the summary entirely? 

```rust
///     // Summary:
///     //
///     // Metadata is used by the userspace program to attach additional
///     // information (similar to Kubernetes labels) to an eBPF program
///     // when it is loaded. This metadata consists of key-value pairs
///     // (e.g., `owner=acme`) and can be used for filtering or selecting
///     // programs later, for instance, using commands like `bpfman list
///     // --metadata-selector owner=acme`.
///     //
///     // Global data is shared between the userspace program and the
///     // eBPF program. The keys must be defined in the eBPF program's
///     // BPF maps because the eBPF program directly interacts with these
///     // data structures.

```

As you mentioned, we've already described global data and, for that matter, metadata too.

bpfman/src/lib.rs
frobware: The insert calls were just examples. I can leave it at: 

```rust
let mut metadata = HashMap::new();
```

if you think that's sufficient.

bpfman/src/lib.rs
frobware: So the example, which includes two exemplary calls to insert would be: 

```rust
/// use bpfman::add_program;
/// use bpfman::errors::BpfmanError;
/// use bpfman::types::{KprobeProgram, Location, Program, ProgramData};
/// use std::collections::HashMap;
///
/// #[tokio::main]
/// async fn main() -> Result<(), BpfmanError> {
///     // Define the location of the eBPF object file.
///     let location = Location::File(String::from("kprobe.o"));
///
///     // Metadata is used by the userspace program to attach additional
///     // information (similar to Kubernetes labels) to an eBPF program
///     // when it is loaded. This metadata consists of key-value pairs
///     // (e.g., `owner=acme`) and can be used for filtering or selecting
///     // programs later, for instance, using commands like `bpfman list
///     // --metadata-selector owner=acme`.
///     let mut metadata = HashMap::new();
///     metadata.insert("sampling_rate".to_string(), "1000".to_string());
///
///     // Optionally, initialise global data for the program. Global data
///     // allows you to set configuration values that are applied at runtime
///     // during the loading of the eBPF program. The keys in global_data must
///     // already exist in the eBPF program, otherwise, an error
///     // will be encountered. For example:
///     let mut global_data = HashMap::new();
///     global_data.insert("global_counter".to_string(), vec![0; 8]);
///
///     // Optionally specify the owner of eBPF maps as a UID (e.g.,
///     // Some(1001)).
///     let map_owner_id = None;
///
///     // Create the program data with the specified location, name,
///     // metadata, global data, and optional map owner ID. ProgramData
///     // holds all necessary information for an eBPF program.
///     let program_data = ProgramData::new(location, String::from("kprobe_do_sys_open"), metadata, global_data, map_owner_id)?;
///
///     // Create a kprobe program with the specified function name,
///     // offset, and options. This sets up a probe at a specific point
///     // in the kernel function named "do_sys_open".
///     let probe_offset: u64 = 0;
///     let is_retprobe: bool = false;
///     let container_pid: Option<i32> = None;
///     let kprobe_program = KprobeProgram::new(program_data, String::from("do_sys_open"), probe_offset, is_retprobe, container_pid)?;
///
///     // Add the kprobe program using the bpfman manager.
///     let added_program = add_program(Program::Kprobe(kprobe_program)).await?;
///
///     // Print a success message with the name of the added program.
///     println!("Program '{}' added successfully.", added_program.get_data().get_name()?);
///     Ok(())
/// }
```

I can take out both calls to insert if they add no value.

Better?

bpfman/src/lib.rs
frobware: For the commentary I've added, I believe this is the only reference to TC. Updated in: f66f909491cb79fd1d4890b1a8a6d843fda8b877.

bpfman/src/lib.rs
Billy99: Yes, better. For metadata, I would change it to the example you use in the comment. The `sampling_rate` makes it appear like you are configuring something else.
```
/// metadata.insert("owner".to_string(), "acme".to_string());
```

Billy99 (commented):
One last nit.

anfredette:
@frobware, I rebased this pr to get it to pass the build check, however, I just realized that you still need to squash everything and sign it to pass the DCO check.  So, you should do a rebase yourself after you squash and then force push it.

Billy99:
Yes, feel free to go ahead and squash and have it ready to merge.

frobware:
Commits have been squashed, and I have rebased. Ready for final review.

Billy99 (approved):
/lgtm

Thanks. Sorry for the delay.

PR #1191, opened 2024-07-18
bpfman: XDP Mode Updates

Billy99:
Based on the Ingress Node Firewall integration with bpfman, if was found that the XDP Mode handling was not ideal. For context, the valid values are HW, DRV and SKB. HW is really only applicable for Netronome hardware.

Adding the following changes:
* The config file examples (bpfman.toml) for xdp_mode were invalid. Needed an `s` on interface.eth0.
* If config file value for a given interface is not entered, SKB was the default. Changing it to DRV and and retrying as SKB if DRV or HW fail.

### DIFF
--- a/docs/developer-guide/configuration.md
+++ b/docs/developer-guide/configuration.md
@@ -8,8 +8,8 @@ There is an example at `scripts/bpfman.toml`, similar to:
 
 ```toml
 [interfaces]
-  [interface.eth0]
-  xdp_mode = "hw" # Valid xdp modes are "hw", "skb" and "drv". Default: "skb".
+  [interfaces.eth0]
+  xdp_mode = "drv" # Valid xdp modes are "hw", "skb" and "drv". Default: "drv", but will fall back to "skb" on failure.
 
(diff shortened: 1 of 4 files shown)


docs/developer-guide/configuration.md
```diff
@@ -8,8 +8,8 @@ There is an example at `scripts/bpfman.toml`, similar to:
 
 ```toml
 [interfaces]
-  [interface.eth0]
-  xdp_mode = "hw" # Valid xdp modes are "hw", "skb" and "drv". Default: "skb".
+  [interfaces.eth0]
+  xdp_mode = "drv" # Valid xdp modes are "hw", "skb" and "drv". Default: "skb".
```
msherif1234, line 12: do u want to change `Default` here and below to `drv`

Billy99: Yes, good catch. Done.

msherif1234 (approved):

anfredette (approved):
LGTM

PR #1192, opened 2024-07-19
fixup bytecode image building

astoycos:
Ensure endianness is set correctly.

Ensure selecting bytecode from cilium project
is a bit more graceful/ works with different sub string ordering

### DIFF
--- a/bpfman/src/bin/cli/image.rs
+++ b/bpfman/src/bin/cli/image.rs
@@ -107,12 +107,21 @@ pub(crate) async fn execute_build(args: &BuildBytecodeArgs) -> anyhow::Result<()
         .unwrap()
         .split('=')
         .collect();
+
     let bc_file = PathBuf::from(first_arg[1]);
-    let (prog_labels, map_labels) = if first_arg[0].contains("EL") {
-        build_bpf_info_image_labels(&bc_file, Some(Endianness::Little))?
+
+    // Make sure bytecode matches host endian if we're building a host endian image
+    // otherwise determine correct endianness based on build argument naming scheme
+    let expected_endianess = if build_context.platforms.is_none() {
+        Endianness::default()
+    } else if first_arg[0].contains("EL") {
+        Endianness::Little
     } else {
-        build_bpf_info_image_labels(&bc_file, Some(Endianness::Big))?
+        Endianness::Big
     };
+
+    let (prog_labels, map_labels) =
+        build_bpf_info_image_labels(&bc_file, Some(expected_endianess))?;
 
     container_tool.build_image(
         &args.tag,
@@ -258,7 +267,9 @@ pub(crate) fn parse_bytecode_from_cilium_ebpf_project(
             let arch = if let Some(file_name) = file_name.to_str() {
                 GoArch::from_cilium_ebpf_file_str(file_name).map_err(|e| anyhow!(e))
             } else {
-                return Err(anyhow!("Could not parse file name in cilium/ebpf project"));
+                return Err(anyhow!(
+                    "Could not parse file name {file_name:?} in cilium/ebpf project"
+                ));
             }?;
             debug!("Found bytecode file for {arch:?} in cilium/ebpf project.");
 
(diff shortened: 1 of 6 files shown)


msherif1234:
Can we also have graceful handling for unsupported ebpf prog types by just adding a warning but proceed with the build ?

bpfman/src/bin/cli/image.rs
```diff
@@ -107,13 +107,22 @@ pub(crate) async fn execute_build(args: &BuildBytecodeArgs) -> anyhow::Result<()
...
+
     let bc_file = PathBuf::from(first_arg[1]);
-    let (prog_labels, map_labels) = if first_arg[0].contains("EL") {
-        build_bpf_info_image_labels(&bc_file, Some(Endianness::Little))?
+
+    // Make sure bytecode matches host endian if we're building a host endian image
+    // otherwise determine correct endianness based on build argument naming scheme
+    let expected_endianess = if build_context.platforms.is_none() {
+        Endianness::default()
+    } else if first_arg[0].contains("EL") {
```
Billy99, line 117: Does this need to be the following (based on the other change you made to allow --bytecode to work for me)?

```suggestion
    } else if first_arg[0].contains("EL") || first_arg[0].contains("BYTECODE_FILE") {
```

astoycos: Good catch, let me reanalyze this logic really quick

msherif1234 (changes requested):
what about this
https://github.com/bpfman/bpfman/blob/main/examples/Makefile#L139

examples/Makefile
```diff
@@ -134,10 +134,6 @@ build-us-images: build ## Build all example userspace images
 	docker buildx build -t ${IMAGE_GT_US} --platform ${PLATFORM} --load -f ./go-target/container-deployment/Containerfile.go-target ../
 	docker buildx build -t ${IMAGE_APP_US} --platform ${PLATFORM} --load  -f ./go-app-counter/container-deployment/Containerfile.go-app-counter ../
 
-.PHONY: build-bc-images
-build-bc-images: generate ## Build bytecode example userspace images
-	IMAGE_TC_BC=${IMAGE_TC_BC} IMAGE_TP_BC=${IMAGE_TP_BC} IMAGE_XDP_BC=${IMAGE_XDP_BC} IMAGE_KP_BC=${IMAGE_KP_BC} IMAGE_UP_BC=${IMAGE_UP_BC}  IMAGE_URP_BC=${IMAGE_URP_BC} IMAGE_APP_BC=${IMAGE_APP_BC} ./build-bytecode-images.sh
-
 build-all-images: build-bc-images build-us-images ## Build both bytecode and userspace images
```
msherif1234, line 137: we can't do bc now because u removed that target and the same for push bc right ?

bpfman/src/bin/cli/image.rs
anfredette: @Billy99, what was the use case that you think won't work?  The following seems to work for me without the change:
```
bpfman image build --bytecode ./go-kprobe-counter/bpf_x86_bpfel.o -f /home/afredette/go/src/github.com/bpfman/bpfman/Containerfile.bytecode --tag quay.io/afredette/go-kprobe-counter
```

Billy99: While reviewing #1141, a similar command as you entered was failing for me (using --bytecode). Andrew fixed in #1141 with:
https://github.com/bpfman/bpfman/pull/1141/files#diff-2229a2342526fa26a90e753aff04cbe5e9b50fe2ca0c302d208447c9f5148ab7R296
So I was asking if a similar fix was needed here.

anfredette: I'm going to leave this as it is.  Everything seems to work, but we can look at it later.

examples/Makefile
anfredette: Removed

anfredette (approved):
LGTM

Issue #1200, opened 2024-07-26
bpfman image generate-build-args --cilium-ebpf-project failing for some examples


anfredette:
`bpfman image generate-build-args --cilium-ebpf-project <DIRECTORY>` fails for the following examples: `go-tracepoint-counter`, `go-uprobe-counter`, `go-xdp-counter` with the following error:
```
Error: BTF error

Caused by:
    error parsing BTF header
```
However, it works for the following examples: `go-app-counter`,  `go-kprobe-counter`,  `go-target`,  `go-tc-counter`,  `go-uretprobe-counter`.

anfredette:
Update: It didn't have anything to do with the actual example code.  For generating build args from a cilium directory, we get the list of bpf object files in the directory and then use the first one in the list to generate the build args.  However, Aya Object::parse doesn't currently work for S390 files, so if an S390 file happens to be the first one in the list, it doesn't work.

PR #1201, opened 2024-07-26
Don't use S390 object code for generating build args

anfredette:
For generating build args from a cilium directory, we get the list of bpf object files in the directory and then use the first one in the list to generate the build program and map labels.  However, Aya Object::parse doesn't currently work for S390 files, so if an S390 file happens to be the first one in the list, it doesn't work.

This is a temporary fix that doesn't use S390 files, and if no other type of file exists, it returns an error.

Fixes: #1200


anfredette:
I'm merging this without a review to hopefully get the upstream bpfman-image-build action working again.

PR #1202, opened 2024-07-26
Use aya::Endianness in image.rs instead of object::Endianness

anfredette:
Pr #1197 bumped the version level for object, which introduced an incompatibility between object::Endianness and aya::Endianness and caused a compile error.

### DIFF
--- a/bpfman/src/bin/cli/image.rs
+++ b/bpfman/src/bin/cli/image.rs
@@ -10,6 +10,7 @@ use std::{
 };
 
 use anyhow::{anyhow, Context, Result};
+use aya::Endianness;
 use aya_obj::Object;
 use base64::{engine::general_purpose, Engine};
 use bpfman::{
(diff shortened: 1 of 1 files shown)


anfredette:
We use object::Endianness in a number of other places in the code that didn't have a conflict because we aren't comparing them to aya::Endianness values.  However, should we be using aya::Endianness everywhere?

msherif1234:
> We use object::Endianness in a number of other places in the code that didn't have a conflict because we aren't comparing them to aya::Endianness values. However, should we be using aya::Endianness everywhere?

Why would they go out of sync ? I think one source of truth should make life easier imo

anfredette:
I looked at this a little more closely.  It seems like both bpfman and our version of aya were using the same (or close enough) versions of object::Endianness.  Then, dependabot bumped our version of object, but we're still using the latest version of aya, which is using the older version, which is no longer compatible.

dave-tucker (approved):

PR #1209, opened 2024-08-01
Fix FromAsCasing warnings in container files

frobware:
This change addresses the FromAsCasing warnings issued by Docker during the build process by ensuring that 'as' is converted to 'AS' in the FROM directives.

Warnings observed:
- FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 5)

Reference: https://docs.docker.com/reference/build-checks/from-as-casing/

### DIFF
--- a/Containerfile.bpfman
+++ b/Containerfile.bpfman
@@ -1,7 +1,7 @@
 # We do not use --platform feature to auto fill this ARG because of incompatibility between podman and docker
 ARG BUILDPLATFORM=linux/amd64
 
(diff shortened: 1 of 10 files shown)


msherif1234 (approved):

msherif1234:
Can u do the same for the operator dockerfiles too while u are there ?

frobware:
> Can u do the same for the operator dockerfiles too while u are there ?

https://github.com/bpfman/bpfman-operator/pull/81

PR #1217, opened 2024-08-08
bpfman: dispatcher should point to release tag

Billy99:
When a release is created, XDP and TC code should use the dispatcher from image based on a release tag, not a latest (today they always point to "quay.io/bpfman/xdp-dispatcher:v2" and "quay.io/bpfman/tc-dispatcher:v1" which are essentially "latest"). Move both image strings to a common location for easier updating and update the release doc to indication they should be updated.
    
Also deprecated "quay.io/bpfman/xdp-dispatcher:v1". Left the source file in place, but no longer build it automatically. Also deprecated the tags v1 and v2 in quay and now just use latest.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -206,12 +206,14 @@ Once these steps are completed:
     - Updates the bpfman-operator version in it's [Makefile](https://github.com/bpfman/bpfman/blob/main/bpfman-operator/Makefile):
         - `VERSION ?= x.x.x`
     - Runs `make bundle` from the bpfman-operator directory to update the bundle version.
-    - Adds a new `examples` config directory for the release version
+    - Adds a new `examples` config directory for the release version.
+    - Modify the tag for `XDP_DISPATCHER_IMAGE` and `TC_DISPATCHER_IMAGE` in `bpfman/src/lib.rs` from `latest` to the
+      new release tag.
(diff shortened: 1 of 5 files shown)


RELEASE.md
```diff
@@ -206,12 +206,14 @@ Once these steps are completed:
     - Updates the bpfman-operator version in it's [Makefile](https://github.com/bpfman/bpfman/blob/main/bpfman-operator/Makefile):
         - `VERSION ?= x.x.x`
     - Runs `make bundle` from the bpfman-operator directory to update the bundle version.
-    - Adds a new `examples` config directory for the release version
+    - Adds a new `examples` config directory for the release version.
+    - Modify the tag for `XDP_DISPATCHER_IMAGE` and `TC_DISPATCHER_IMAGE` in `bpfman/src/lib.rs` from `latest` to the
+      new release tag.
```
anfredette, line 211: Add note:  After release is tagged, change the version tag back back to latest and push/merge pr.

anfredette (changes requested):
Just one nit on RELEASE.md.  Otherwise, LGTM.

anfredette (approved):
I can add the note to RELEASE.md when I update it, so I'm going to just approve this.  Thanks for chasing this down!

Issue #1221, opened 2024-08-12
bpfman: Failed dispatcher load corrupts DB


Billy99:
If the TC or XDP dispatcher fails to load, subsequent load commands fail with a panic in bpfman:

```
Aug 12 08:51:07 nfvsdn-21-oot.lab.eng.rdu2.redhat.com bpfman-rpc[2631646]: Loading program bytecode from file: /home/bmcfall/src/bpfman/bpfman/examples/go-xdp-counter/bpf_x86_bpfel.o
Aug 12 08:51:07 nfvsdn-21-oot.lab.eng.rdu2.redhat.com bpfman-rpc[2631646]: thread 'tokio-runtime-worker' panicked at bpfman/src/multiprog/mod.rs:111:18:
Aug 12 08:51:07 nfvsdn-21-oot.lab.eng.rdu2.redhat.com bpfman-rpc[2631646]: failed to get xdp_dispatcher num_extensions: DatabaseError("Database entry num_extension does not exist in tree \"xdp_dispatcher__4_1\"", "")
Aug 12 08:51:07 nfvsdn-21-oot.lab.eng.rdu2.redhat.com bpfman-rpc[2631646]: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Aug 12 08:51:20 nfvsdn-21-oot.lab.eng.rdu2.redhat.com bpfman-rpc[2631646]: Shutdown Unix Handler /run/bpfman-sock/bpfman.sock
```

To test, modify the dispatcher image to something that doesn't exist:
```
    pub(crate) const XDP_DISPATCHER_IMAGE: &str = "quay.io/bpfman/xdp-dispatcher:la-test";
    pub(crate) const TC_DISPATCHER_IMAGE: &str = "quay.io/bpfman/tc-dispatcher:la-test";
```

Then try to load TC or XDP program twice. The first fails with dispatcher load failure as expected, then the second attempt fails with the panic:

```
$ sudo ./go-xdp-counter -iface eno3 -file /home/$USER/src/bpfman/bpfman/examples/go-xdp-counter/bpf_x86_bpfel.o
2024/08/12 08:50:44 Using Input: Interface=eno3 Priority=50 Source=/home/bmcfall/src/bpfman/bpfman/examples/go-xdp-counter/bpf_x86_bpfel.o
[...]

PR #1223, opened 2024-08-12
bpfman: Failed dispatcher load corrupts DB

Billy99:
If the TC or XDP dispatcher fails to load, subsequent load commands fail with a panic in bpfman. To fix, removed an unwrap(), add a dispatcher delete in the Dispatcher.new() function when an error occurs, and mapped some errors to new BpfmanError for dispatcher failures instead of just returning lower level error.

Resolves: #1221


anfredette (approved):
LGTM

PR #1226, opened 2024-08-14
bpfman: restore caps for bpfman-rpc as a service

Billy99:
In scripts/bpfman.service, the capabilites granted to the service are commented out. This was done in #1053 commit#2, which updated the scripts to install bpfman-rpc instead of `bpfman system service`.


Billy99:
I don't remember why the caps were commented out. I ran locally as a systemd service and everything seems to run fine. Pushing a PR to run through CI and see if any issues are seen.

anfredette:
What was the use case that didn't work?

Billy99:
> What was the use case that didn't work?

Originally or now? I don't know originally, I don't remember why I commented them out. Now, I was locally changing logging from "Info" to "Debug" in the file and noticed the caps were commented out. The reasons for the caps in the bpfman.service file is to limit the set of capabilities granted to a process.

anfredette (approved):
LGTM

PR #1235, opened 2024-08-28
Move Fedora back to release 40 in Containerfile.bpfman.local

anfredette:
This was bumped up to 41 by Kustomize.  However, 41 isn't released yet, and there's no good reason to use an unreleased version.


Billy99 (approved):
/lgtm

PR #1240, opened 2024-09-04
bpfman: Make whether to use cosign configurable

anfredette:
This pr adds an option to the `Config.signing`  called `verify_enabled` that
controls whether bpfman checks the signatures on images.

`verify_enabled` is currently defaulted to `false`, due to the current issue
with TUF (https://github.com/bpfman/bpfman/issues/1241).  It should be defaulted to 
`true` after the issue is resolved.

Reading the config and setting defaults is cleaned up as well.

Fixes: https://github.com/bpfman/bpfman/issues/1239

### DIFF
--- a/bpfman/src/oci_utils/image_manager.rs
+++ b/bpfman/src/oci_utils/image_manager.rs
@@ -8,7 +8,7 @@ use std::{
 
 use anyhow::anyhow;
 use flate2::read::GzDecoder;
-use log::{debug, trace};
+use log::{debug, info, trace};
 use object::{Endianness, Object};
 use oci_distribution::{
     client::{ClientConfig, ClientProtocol},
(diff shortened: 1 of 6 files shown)


bpfman/src/oci_utils/image_manager.rs
```diff
@@ -93,9 +97,11 @@ impl ImageManager {
...
         let image: Reference = image_url.parse().map_err(ImageError::InvalidImageUrl)?;
 
-        self.cosign_verifier
-            .verify(image_url, username.as_deref(), password.as_deref())
-            .await?;
+        if let Some(cosign_verifier) = &mut self.cosign_verifier {
+            cosign_verifier
+                .verify(image_url, username.as_deref(), password.as_deref())
+                .await?;
+        }
```
Billy99, line 104: Add and else and a info!() log indicating Cosign is being skipped.

Billy99 (commented):
Thanks Andre. I think we should log that Cosign is being skipped.

bpfman/src/oci_utils/image_manager.rs
anfredette: Good point.  Done!

Billy99 (approved):
/lgtm

PR #1242, opened 2024-09-04
Add an integration test for adding a program with cosign disabled

anfredette:


tests/integration-test/src/tests/basic.rs
```diff
@@ -493,3 +493,29 @@ fn test_load_unload_fexit() {
...
+fn test_load_unload_cosign_disabled() {
+    let _cosign_guard = disable_cosign();
+
+    debug!("Installing kprobe program with cosign disabled");
+
+    let globals = vec!["GLOBAL_u8=63", "GLOBAL_u32=0D0C0B0A"];
+
+    let mut loaded_ids = vec![];
+
+    for lt in LOAD_TYPES {
```
Billy99, line 507: Do we need to test the load type "File" with cosign disabled? Wouldn't cosign be skipped anyway? Would save a little bit of time on the total test time.

anfredette: Good point.  I removed the loop and am just using LoadType::Image.  Thanks, Andre

Billy99 (approved):
/lgtm

PR #1244, opened 2024-09-06
Fix BUILDPLATFORM redefinition issue in bpfman Containerfile

msherif1234:
unblock https://github.com/bpfman/bpfman/pull/1214


Billy99 (approved):
/lgtm

PR #1256, opened 2024-09-10
ci: pin rust to nightly-2024-09-09 until public-api fixed

Billy99:
The public-api (cargo xtask public-api) is failing with the latest rust nightly package. Temporarily pin nightly to last known working version until source of the issue can be resolved.


anfredette (approved):
LGTM

Billy99:
Tracking via #1260

PR #1258, opened 2024-09-11
Set proper cargo config for aya

danielmellado:
This commit sets up cargo to vendor the specific aya commit so builds work
without trying to fetch it from github repo. While this may work locally
with Cargo build it'll break all rpm builds on offloine mode.


dave-tucker:
Closing since we're on track to do a release of bpfman that brings in Aya 0.13.0

PR #1284, opened 2024-09-27
ci: add --toolchain to public-api to make pinning easier

Billy99:
Currently, "nightly" is hard-coded in the xtask public-api code. Add a --toolchain parameter so it can be set from commandline. Also update build.yml to pass the current toolchain as a parameter to public-api call.


anfredette:
The following works for me with this change, but shouldn't it fail?
```
cargo xtask public-api --toolchain nightly-2024-09-10 --bless
```

PR #1292, opened 2024-10-08
cI: Create scorecard.yml

dave-tucker:
Adds the OpenSSF Scorecard action to GitHub Actions

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,0 +1,73 @@
+# This workflow uses actions that are not certified by GitHub. They are provided
+# by a third-party and are governed by separate terms of service, privacy
+# policy, and support documentation.
+
+name: Scorecard supply-chain security
+on: # yamllint disable-line rule:truthy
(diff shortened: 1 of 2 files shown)


anfredette (commented):
Question: Did you derive this yaml from Kepler, or somewhere else?  I ask because scorecard has the following sample yaml:
https://github.com/ossf/scorecard/blob/main/.github/workflows/scorecard-analysis.yml
It's pretty similar, though it has the actions pinned to some newer versions which we probably want to update to.

dave-tucker:
> Question: Did you derive this yaml from Kepler, or somewhere else? I ask because scorecard has the following sample yaml: https://github.com/ossf/scorecard/blob/main/.github/workflows/scorecard-analysis.yml It's pretty similar, though it has the actions pinned to some newer versions which we probably want to update to.

This came from the GitHub UI by following the instructions to add it from the OpenSSF website.
Dependabot should take care of doing the bumps after merge.

anfredette (approved):
LGTM.  Let's see how it works!

PR #1293, opened 2024-10-08
ci: Move clippy to a code scanning action

dave-tucker:
There's support for clippy -> sarif and therefore integration into the GitHub Code Scanning infrastructure. Let's try this out.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -91,7 +91,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           toolchain: ${{ env.NIGHTLY_VERSION }}
-          components: rustfmt, clippy, rust-src
+          components: rustfmt, rust-src
           override: false
 
       # Only install cross if we need it
@@ -157,12 +157,6 @@ jobs:
         if: ${{ matrix.arch.arch == 'amd64' }}
         run: |
           cargo +${{ env.NIGHTLY_VERSION }} fmt --all -- --check
-
-      - name: Run clippy
-        # Only need to run once
-        if: ${{ matrix.arch.arch == 'amd64' }}
-        run: |
-          cargo +${{ env.NIGHTLY_VERSION }} clippy --all -- --deny warnings
 
       - name: Check public API
         # Only need to run once
--- a/.github/workflows/rust-clippy.yaml
+++ b/.github/workflows/rust-clippy.yaml
@@ -1,0 +1,54 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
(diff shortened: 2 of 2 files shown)


Billy99:
Just a couple of questions/notes:
* What additional functionality is this providing? I looked up clippy-sarif and it converts to output to SARIF to be uploaded to github. But how is that going to be leveraged in github? Are clippy errors just shown in the code now?
* If we decide we want it, we need to update the CONTRIBUTING.md to show how to run locally ([here](https://github.com/bpfman/bpfman/blob/main/CONTRIBUTING.md?plain=1#L181)) when an error is encountered. `cargo +nightly clippy --all -- --deny warnings` should have been there currently but it wasn't.
* Update docs [here](https://github.com/bpfman/bpfman/blob/main/docs/getting-started/building-bpfman.md?plain=1#L78) for new packages that are needed to run locally.

anfredette:
I was just looking into this too:
https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/about-integration-with-code-scanning

anfredette:
I was also wondering what we get this this change.  The results are uploaded and then can be used by other tools, but I'm not sure which ones we would use.  Maybe errors will show up in the code review tool?

> Are clippy errors just shown in the code now?

I think rust-analyzer shows us most of the errors in vscode now, but it's not exactly the same as when you run `cargo +nightly clippy --all -- --deny warnings`.

I'll wait for Dave's response.

dave-tucker:
> What additional functionality is this providing? I looked up clippy-sarif and it converts to output to SARIF to be uploaded to github. But how is that going to be leveraged in github? Are clippy errors just shown in the code now?

This should make clippy run as a code scanning tool - so any alerts will also show up here - alongside any other code scanning tools we might add:
https://github.com/bpfman/bpfman/security/code-scanning

I also hope it can give a better experience. i.e it should show the lines that violate the clippy rule in the UI.

> If we decide we want it, we need to update the CONTRIBUTING.md to show how to run locally ([here](https://github.com/bpfman/bpfman/blob/main/CONTRIBUTING.md?plain=1#L181)) when an error is encountered. cargo +nightly clippy --all -- --deny warnings should have been there currently but it wasn't.

I'll open an issue about using [precommit](https://pre-commit.com/) as it can automate all these pre-commit checks.
Then we'll have less docs to update 😄 

> Update docs [here](https://github.com/bpfman/bpfman/blob/main/docs/getting-started/building-bpfman.md?plain=1#L78) for new packages that are needed to run locally.

No changes required as no new packages required.

I think it's worth trying out and if we hate it, we can back it out again.

anfredette (commented):
LGTM.  Let's give it a try.

@Billy99, I'll leave it to you to approve.

Billy99 (approved):
/lgtm

Now that the release is out, let give it a try.

PR #1296, opened 2024-10-08
Allow the scorecard workflow to be manually triggered

anfredette:
Add `workflow_dispatch` to the scorecard workflow so it can be manually triggered as needed while we try to improve the score.


Billy99 (approved):
/lgtm

PR #1298, opened 2024-10-09
Upgrade to aya version 0.13.0 and aya-obj version 0.2.0

anfredette:
Changes include:
- Update bpfman TC code to new Aya APIs
- The new Aya release now returns Aya enums for ProgramType and MapType
  instead of u32. Added code to convert these enums back to u32 values 
  to reduce change in the bpfman code.
- Added an enum for the new Netfilter program type introduced in Aya.

### DIFF
--- a/bpfman/src/types.rs
+++ b/bpfman/src/types.rs
@@ -9,7 +9,10 @@ use std::{
     time::SystemTime,
 };
 
-use aya::programs::ProgramInfo as AyaProgInfo;
+use aya::{
+    maps::MapType as AyaMapType,
+    programs::{ProgramInfo as AyaProgInfo, ProgramType as AyaProgramType},
+};
 use chrono::{prelude::DateTime, Local};
 use clap::ValueEnum;
 use log::{info, warn};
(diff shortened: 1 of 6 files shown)


bpfman/src/types.rs
```diff
@@ -2198,11 +2246,54 @@ impl std::fmt::Display for ProgramType {
             ProgramType::Lsm => "lsm",
             ProgramType::SkLookup => "sk_lookup",
             ProgramType::Syscall => "syscall",
+            ProgramType::Netfilter => "netfilter",
         };
         write!(f, "{v}")
     }
 }
 
+impl From<AyaProgramType> for ProgramType {
```
dave-tucker, line 2255: 👀 why?

bpfman/src/types.rs
```diff
@@ -1958,6 +1959,49 @@ impl std::fmt::Display for MapType {
     }
 }
 
+impl From<AyaMapType> for MapType {
```
dave-tucker, line 1962: 👀 why?

dave-tucker (changes requested):
Wondering why we're mapping Aya types -> Bpfman types vs using the aya types directly 🤔

bpfman/src/types.rs
anfredette: Pretty much the same reason for both AyaMapType and AyaProgramType.  Aya used to return the kernel id (u32) for both, but now it returns an enum and gives us no way to get at the id.  So, I mappped from the Aya type to the bpfman type, and then I can convert to either a u32 or string using our existing code.

bpfman/src/types.rs
anfredette: See above.

bpfman/src/types.rs
anfredette: And, Rust wouldn't let me map directly from the Aya type to a u32 or string directly because we don't define either.

dave-tucker: > Pretty much the same reason for both AyaMapType and AyaProgramType. Aya used to return the kernel id (u32) for both, but now it returns an enum and gives us no way to get at the id. So, I mappped from the Aya type to the bpfman type, and then I can convert to either a u32 or string using our existing code.

ugh yeah iirc this is because we need to handle the u32s in the backend - i.e store to database.
I guess this is fine for now, but I think we owe a follow up PR in aya to allow conversion into u32 somehow.

dave-tucker (approved):

PR #1253, opened 2024-09-10
cross-build (POC)

frobware:
This PR enables cross-compiling `bpfman` without using Docker-in-Docker. 

### Dockerfile:
- **Stage 1:** Builds `bpfman` using a Rust base image for the specified target architecture (`TARGETARCH`).
- **Stage 2:** Packages the built binaries in a minimal Debian image for runtime.

### `cross-build.sh`:
- Sets up cross-compilation for supported architectures (e.g., `aarch64`, `ppc64le`).
- Installs required toolchains and dependencies.
- Configures the environment and builds `bpfman` for the target using Cargo and the appropriate Rust target.

This approach simplifies cross-compilation by eliminating the need for nested containers.

```sh
% uname -m
x86_64

% docker buildx build -f Containerfile.cross-build --build-arg TARGETARCH=aarch64 -t bpfman-aarch64 .
[+] Building 94.0s (18/18) FINISHED                                                                                 docker:default
 => [internal] load build definition from Containerfile.cross-build                                                           0.1s
 => => transferring dockerfile: 760B                                                                                          0.0s
 => [internal] load metadata for docker.io/library/debian:bullseye-slim                                                       0.8s
 => [internal] load metadata for docker.io/library/rust:latest                                                                0.8s
 => [auth] library/debian:pull token for registry-1.docker.io                                                                 0.0s
 => [auth] library/rust:pull token for registry-1.docker.io                                                                   0.0s
 => [internal] load .dockerignore                                                                                             0.0s
 => => transferring context: 65B                                                                                              0.0s
 => [bpfman-build 1/5] FROM docker.io/library/rust:latest@sha256:fcd390e0a3a6bfcf26969861efbe7b864df052aa71a361cf3cd7c5c585b  0.0s
 => [internal] load build context                                                                                             0.4s
 => => transferring context: 287.21MB                                                                                         0.4s
 => [stage-1 1/5] FROM docker.io/library/debian:bullseye-slim@sha256:00558f781b91e90469812bad32002f311ab26ef241b4a1996f66006  0.0s
 => CACHED [bpfman-build 2/5] WORKDIR /usr/src/bpfman                                                                         0.0s
 => [bpfman-build 3/5] COPY . .                                                                                               0.6s
 => [bpfman-build 4/5] RUN ./scripts/cross-build.sh aarch64                                                                  91.5s
 => [bpfman-build 5/5] RUN mkdir bin && cp target/aarch64-unknown-linux-gnu/debug/bpfman* bin                                 0.2s
[...]

PR #1313, opened 2024-10-15
image: bpfman image build needs same build and base image

Billy99:
bpfman images are built in github actions. The binary is built using .github/wokflows/image-build.yml which is currently using ubuntu:latest. Then the image is build by copoying the binary into a container. The base image for the container is redhat/ubi9-minimal, which used glibc 2.34. Recently, ubuntu:latest change from ubuntu 22.04 (glibc 2.34) to ubuntu 24.04 (glibc 2.39). So now bpfman won't run in a redhat/ubi9-minimal base image.

This PR moves the build image back to ubuntu 22.04 (glibc 2.34). A follow-up PR #1253 will change how this works and use the same build and base images.

### DIFF
--- a/.github/workflows/image-build.yml
+++ b/.github/workflows/image-build.yml
@@ -18,7 +18,7 @@ jobs:
   # "build-and-push-bytecode-images" step. It will be used to generate
   # build args with the "bpfman image generate-build-args" command.
   build-bpfman-for-build-arg-gen:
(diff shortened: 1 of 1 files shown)


Billy99:
@anfredette I was able to load and run with the test build we generated.

anfredette (approved):
lgtm

PR #1325, opened 2024-10-21
Move bpfman container back to redhat/ubi9-minimal

anfredette:
We're having failures in pulling/loading bpf programs with the new image.  It's not fully debugged, but this is the likely change.


Billy99 (approved):
/lgtm

PR #1326, opened 2024-10-22
image: Move bpfman container back to Ubuntu 24.04

Billy99:
Backout #1325, which moved image building from Ubuntu 24.04 (which is needed for TCX) to ubi9-minimal. #1325 was needed because images originally built with Ubuntu 24.04 were failing to load eBPF programs because of Cosign failures. Turns out that those failures were due to ca-certicates not being installed in Ubunut 24.04 base images. So this PR also adds ca-certicates to the bpfman base image.

To allow local building of bpfman images that are consistent with github built images, this PR also moves Containerfile.bpfman.local to build with Ubuntu 24.04.


anfredette (approved):
/LGTM.

PR #1327, opened 2024-10-22
Modify filter() to just get TC programs

anfredette:
The filter() function used when loading and unloading multi attach programs was getting both TC and TCX programs because they both have the same ProgramType's.  I modified it to just get TC programs and renamed it to get_multi_attach_programs()

This was causing adding tc programs to fail if there were already tcx progrms installed.

I also removed a few more unwrap()s and handled the errors in the process.


Billy99 (approved):
/lgtm

PR #1340, opened 2024-11-09
WIP: Netns support for XDP, TC, and TCX

anfredette:
This is an initial POC with support for just TCX.

Questions/Comments:
1. Does the bpfman API look okay?  The following option was added for XDP, TC, and TCX:
```
  -n, --netns <NETNS>
          Optional: The name of the target network namespace
```

2. Unlike when we set the mount namespace for uprobes, it's possible to set the network namespace directly in bpfman (i.e., without using a separate process).  Should we still use the separate process?

3.  We'll need to host mount `/var/run/netns` on Kubernetes to access the node netns files.


msherif1234:
can u share in this PR or in the description example creating custom netns and use the cli to attach an ebpf prog to an interface in the new netns and show the outcomes with ip link show ?
or better yet to add it somewhere in the docs ?

anfredette:
>can u share in this PR or in the description example creating custom netns and use the cli to attach an ebpf prog to an interface in the new netns and show the outcomes with ip link show ?
or better yet to add it somewhere in the docs ?

Done.

xtask/public-api/bpfman.txt
```diff
@@ -1708,6 +1711,13 @@ pub const bpfman::utils::SOCK_MODE: u32
 pub fn bpfman::utils::create_bpffs(directory: &str) -> anyhow::Result<()>
 pub fn bpfman::utils::set_dir_permissions(directory: &str, mode: u32)
 pub fn bpfman::utils::set_file_permissions(path: &std::path::Path, mode: u32)
+pub macro bpfman::tc_did_string!
```
Billy99, line 1714: Should these macros be public?

bpfman/src/multiprog/tc.rs
```diff
@@ -165,19 +172,22 @@ impl TcDispatcher {
...
 
         self.loader = Some(loader);
         self.set_num_extensions(extensions.len())?;
         self.set_program_name(TC_DISPATCHER_PROGRAM_NAME)?;
 
         self.attach_extensions(&mut extensions)?;
-        self.attach(root_db, old_dispatcher).await?;
+
+        if let Some(netns) = netns {
+            let _netns_guard = set_netns(netns)?;
```
Billy99, line 185: I'm assuming `drop()` is called automatically. Is it when the code leaves this if statement, the function, or other? May want to sprinkle some comments around this usage, because there is also the `set_netns()` function that writes the value to the DB. Someone skimming the code might miss that these are two different functions and this one is entering a network namespace.

scripts/del-bpfman-qdiscs.sh
```diff
@@ -1,20 +1,25 @@
...
+}
 
-for i in "${interfaces[@]}";
-do
-	# delete the clsact qdisc if it exists
-	sudo tc qdisc del dev $i clsact 2>/dev/null;
-	# delete the ingress qdisc if it exists
-	sudo tc qdisc del dev $i ingress 2>/dev/null;
-done
+# Delete qdiscs in the default namespace
```
Billy99, line 18: Based on thread on slack, may need to wrap this logic in a function so it can be called from `sudo ./scripts/setup.sh uninstall`. That may require breaking it across two files so cleanup can be called directly, or we just assume it is only used with `sudo ./scripts/setup.sh uninstall`.

scripts/nsadd.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
Billy99, line 1: Do wee need a `nsdel.sh`?

Billy99 (commented):
Still reviewing, but wanted to post what I had so far.

bpfman/src/multiprog/tc.rs
anfredette: Yes, drop is called automatically when _netns_guard goes out of scope.  In this case, that's just inside the if statement.  I should probably change the function name so that there's no confusion.

scripts/del-bpfman-qdiscs.sh
anfredette: I'll add the qdisc cleanup to `./scripts/setup.sh uninstall`, but I may do it in a separate PR.  I think we have an issue in the has_qdisc() code that is independent from the new netns functionality added in this PR.

scripts/nsadd.sh
anfredette: I added one.

bpfman/src/utils.rs
```diff
@@ -301,3 +312,132 @@ pub(crate) fn initialize_bpfman() -> anyhow::Result<()> {
...
+                tc_base, netns, $if_index, $revision
+            )
+        } else {
+            format!("{}/dispatcher_{}_{}", tc_base, $if_index, $revision)
+        }
+    }};
+}
+pub(crate) use tc_dispatcher_rev_path;
+
+macro_rules! tc_dispatcher_link_id_path {
```
dave-tucker, line 434: These macros should really be functions IMHO.
The benefit of using macros here would be if we were able to do something like:

```
dispatcher_db_key!(tc, direction, netns, if_index); // equivalent to did_string
disptacher_db_key!(xdp, if_index, netns);

dispatcher_link!(tc, direction, netns, if_index, revision); // equivalent to dispatcher_rev_path
dispatcher_link!(tc, direction, netns, if_index, revision, id); // equivalent to dispatcher_link_id_path
```

i.e making use of variadic arguments and/or not having to specify a type for the arguments

bpfman-ns/src/main.rs
```diff
@@ -111,11 +111,7 @@ fn execute_uprobe_attach(args: UprobeArgs, bpfman_pid: u32) -> anyhow::Result<()
         .context("failed to get UProbe from pin file")?;
 
     // Set namespace to target namespace
-    set_ns(
-        target_mnt_file,
-        CloneFlags::CLONE_NEWNS,
-        args.container_pid as u32,
-    )?;
+    enter_mnt_ns(target_mnt_file, args.container_pid as u32)?;
```
dave-tucker, line 114: nit: This doesn't need to be renamed in this PR

scripts/install.sh
```diff
@@ -195,6 +195,12 @@ uninstall() {
     del_svc "${SVC_BPFD_SVC}"
     del_bin "${BIN_BPFD}"
     del_bin "${BIN_BPFCTL}"
+
+    # Delete clsact and ingress qdiscs, in the default namespace and all network
+    # namespaces.  This has the side-effect of removing any tc_dispatchers (or
+    # other programs) that happen to be attached to them.
+    echo "Delete qdiscs:"
+    delete_all_bpfman_qdiscs
```
dave-tucker, line 203: what if we didn't create those qdiscs but something else did?

dave-tucker (changes requested):

dave-tucker (changes requested):
This looks good overall. A couple of nits and a few comments on the API, when we switch into the target netns and usage of macros.

bpfman/src/utils.rs
```diff
@@ -38,8 +40,17 @@ pub(crate) fn read<P: AsRef<Path>>(path: P) -> Result<Vec<u8>, BpfmanError> {
     Ok(data)
 }
 
-pub(crate) fn get_ifindex(iface: &str) -> Result<u32, BpfmanError> {
-    match if_nametoindex(iface) {
+pub(crate) fn get_ifindex(iface: &str, netns: Option<String>) -> Result<u32, BpfmanError> {
```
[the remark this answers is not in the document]
anfredette: I originally tried to move the get_ifindex down into the code that was run after we entered the netns.  However, I ran into problems because there was code that assumed ifindex was already set in the database, so there were problems.  I though it was messier to try to re-arrange the code than to leave the order as is, and just enter the netns to get the ifindex.  Also, the only operations for which we need to be in the netns are when we get the ifindex, and we we do the actual attach, so I thought it would be better to limit the amount of time that the code is in the other netns to just those operations.  I can try to re-work this if you really want, but I'd prefer to leave it as is.

bpfman/src/bin/cli/args.rs
```diff
@@ -155,6 +155,10 @@ pub(crate) enum LoadCommands {
         /// [default: pass, dispatcher_return]
         #[clap(long, verbatim_doc_comment, num_args(1..))]
         proceed_on: Vec<String>,
+
+        /// Optional: The name of the target network namespace.
+        #[clap(short, long)]
+        netns: Option<String>,
```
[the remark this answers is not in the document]
anfredette: Good suggestion.  This will be better because it handles both cases.  I've also been thinking about the kubernetes api for this.  If an app is attaching a program in a network namespace directly on the node, then it probably can know the path, but if the app needs to attach to an interface in the network namespace of a kubernetes pod/container, the app probably doesn't know the path.  However, I think we can use the existing container selector, and our code can figure it out like we do for uprobes.  So, for kubernetes, I was thinking we'd need to provide the user two options: the path to a netns, or a container selector.  What do you think?

scripts/install.sh
anfredette: This is from a script I had to clean up the tc dispatchers if something went wrong and it left them hanging.  They also get left hanging if someone simply calls uninstall when they exist.  However, as you probably noted, it does wipe out those qdiscs, and everything on them, so it's more destructive than it needs to be.  I think this is fine for development, or the case where we know we're the only users of the qdiscs; however, maybe not a default part of uninstall.  With a little work I think I can just detach the programs attached to the priority we use.  That's still not perfect, so with some more work I can probably figure out how to just detach any tc dispatchers that get left hanging.

This change is independent from this pr, so I can either leave it as is (it probably helps more than it hurts right now), or back it out.  In either case, we should still improve it to just detach the hanging tc dispatchers. Let me know.

bpfman/src/utils.rs
```diff
@@ -301,3 +312,132 @@ pub(crate) fn initialize_bpfman() -> anyhow::Result<()> {
...
+    debug!("Switching to the target network namespace: {}", netns);
+    setns(target_netns_file, CloneFlags::CLONE_NEWNET)
+        .map_err(|e| BpfmanError::Error(format!("setns error: {}", e)))?;
+
+    Ok(NetnsGuard {
+        original_ns: bpfman_netns_file,
+    })
+}
+
+macro_rules! xdp_did_string {
```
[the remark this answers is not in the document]
anfredette: I didn't love it either, but it worked.  I'll implement the change you suggest.  However, can you clarify what to do overall?  Should I change them all to functions with the above key change, or get rid of them and go back to custom coding a format!() string for each use?  I personally think it makes things more clear and maintainable to have functions.

bpfman/src/bin/cli/args.rs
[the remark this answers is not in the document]
dave-tucker: The user-facing K8s API should only be selector-based.
bpfman's API will require a path.
bpfman-agent will have to figure out the mapping from container selector -> container ID -> network namespace

scripts/install.sh
dave-tucker: It's fine, just wanted to confirm that it's only developer-facing and we're not requiring users to run it.

bpfman/src/multiprog/tc.rs
```diff
@@ -62,12 +63,14 @@ impl TcDispatcher {
...
         if_index: u32,
         if_name: String,
+        netns: Option<String>,
         revision: u32,
     ) -> Result<Self, BpfmanError> {
         let db_tree = root_db
             .open_tree(format!(
-                "{}_{}_{}_{}",
-                TC_DISPATCHER_PREFIX, if_index, direction, revision
+                "{}_{}",
```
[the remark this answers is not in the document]
anfredette: Not quite.  It's the tree name, which is a little different from the path.  I added a function for it.

bpfman-ns/src/main.rs
anfredette: Removed.

bpfman/src/bin/cli/args.rs
```diff
@@ -183,6 +187,10 @@ pub(crate) enum LoadCommands {
         /// [default: ok, pipe, dispatcher_return]
         #[clap(long, verbatim_doc_comment, num_args(1..))]
         proceed_on: Vec<String>,
+
+        /// Optional: The name of the target network namespace.
+        #[clap(short, long)]
+        netns: Option<String>,
```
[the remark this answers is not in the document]
anfredette: Done

bpfman/src/bin/cli/args.rs
```diff
@@ -203,6 +211,10 @@ pub(crate) enum LoadCommands {
         /// [default: 1000]
         #[clap(short, long)]
         priority: i32,
+
+        /// Optional: The name of the target network namespace.
+        #[clap(short, long)]
+        netns: Option<String>,
```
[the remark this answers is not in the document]
anfredette: Done

bpfman/src/utils.rs
```diff
@@ -301,3 +312,132 @@ pub(crate) fn initialize_bpfman() -> anyhow::Result<()> {
...
+impl Drop for NetnsGuard {
+    fn drop(&mut self) {
+        // Switch back to the original network namespace
+        debug!("Switching back to the original network namespace");
+        setns(&self.original_ns, CloneFlags::CLONE_NEWNET)
+            .expect("Failed to switch back to the original namespace");
+    }
+}
+
+pub(crate) fn enter_netns(netns: String) -> Result<NetnsGuard, BpfmanError> {
```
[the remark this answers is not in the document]
anfredette: Done.

bpfman/src/utils.rs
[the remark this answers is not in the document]
anfredette: This has been reworked to use the inode number of the network namespace file for all casses, which is what the kernel uses as a network namespace identifier.

bpfman/src/utils.rs
anfredette: They are functions now.

anfredette:
@dave-tucker I think I've addressed all of your comments.

PR #1345, opened 2024-11-18
config: rework how default values are managed

Billy99:
Access functions to the config (configuration data read from bpfman.toml) return `Option<>`. So currently, `open_config_file()` checks for None and then call `default()` on each substructure. Update the data structures such that if they values are not in the bpfman.toml file, then default data is automatically used.


anfredette (approved):
/LGTM

Issue #1362, opened 2024-12-18
Error when deleting tc and xdp programs inside a netns after netns deleted


anfredette:
How to reproduce:
1. Attach a tc or xdp program to an interface inside a network namespace.
2. Delete the network namespace
3. Attempt to delete the program

You get the following error:
```
[DEBUG bpfman] Getting dispatcher with id: Xdp(DispatcherInfo(Some(\"/var/run/netns/bpfman-int-test\"), 55, None))
[ERROR bpfman] Error: failed to unload xdp program named \"pass\": An error occurred. Failed to get metadata for namespace path: No such file or directory (os error 2)
[DEBUG sled::pagecache::logger] IoBufs dropped
Error: An error occurred. Failed to get metadata for namespace path: No such file or directory (os error 2)
```
When a tc or xdp program is deleted, we need to update the dispatcher.  However, the namespace id is part of dispatcher id, and we get the namespace id from the file path which is gone if you delete the namespace first.

PR #1363, opened 2024-12-18
Fix Error deleting tc and xdp programs after netns deleted

anfredette:
This fix saves the network namespace ID when the program is loaded so it can be retrieved later when the program is being deleted even if the network namespace is deleted first.

Fixes: #1362


Billy99 (approved):
/lgtm

PR #1371, opened 2025-01-06
revert softprops/action-gh-release

Billy99:
Revert softprops/action-gh-release from v2.2.0 to v2.1.0 due to "Error: Request body length does not match content-length header"

See: https://github.com/softprops/action-gh-release/issues/556


anfredette (approved):
/LGTM

Issue #1282, opened 2024-09-27
ci: clippy failing with lastest rust nightly


Billy99:
Getting the following clippy error with updated rust nightly:

```
rror: unneeded `return` statement
   --> bpf-metrics-exporter/src/main.rs:273:7
    |
273 |     Ok(())
    |       ^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_return
    = note: `-D clippy::needless-return` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::needless_return)]`
help: remove `return`
    |
273 |     OkOk(())
    |       ~~~~~~

error: could not compile `bpf-metrics-exporter` (bin "bpf-metrics-exporter") due to 1 previous error
warning: build failed, waiting for other jobs to finish...
Error: Process completed with exit code 101.
```

Issue tracked by: https://github.com/rust-lang/rust-clippy/issues/13458

Pinning rust to nightly-2024-09-24. Unpin when issue is resolved.

PR #1376, opened 2025-01-08
unpin cargo nightly

Billy99:
Cargo has been pinned to nightly-2024-09-24 for a while. It was originally pinned because of clippy failures (see #1282). But once those were resolved, it was decided to leave pinned and move the pin roughly monthly. Because that hasn't been happening, and it is hard to document this for new comers to the project, move back to nightly. May move back to pinning if clippy type errors slow the project down.

Resolves: #1282


CONTRIBUTING.md
```diff
@@ -176,24 +176,67 @@ When you submit your pull request, or you push new commits to it, our automated
...
 passes these checks, but we also have more criteria than just that before we can
 accept and merge it. We recommend that you check the following things locally
-before you submit your code:
+before you submit your code.
+
+### bpfman Pinned Rust Toolchain
+
+bpfman is coded in Rust and uses the latest `nightly` Rust toolchain for some of
+the tools.
+There are periods where the Rust toolchain may be pinned to a fixed version dues to
```
anfredette, line 185:
```suggestion
There are periods where the Rust toolchain may be pinned to a fixed version due to
```

anfredette (changes requested):
Just one nit.

CONTRIBUTING.md
Billy99: Done

anfredette (approved):
/LGTM

PR #1377, opened 2025-01-10
cli: fix arm64 arch recongnization with --cilium-ebpf-project flag

junotx:
fix that arm64 bytecode file was recognized as an arm arch when using `bpfman image generate-build-args --cilium-ebpf-project <project>`


Billy99 (approved):
/lgtm

Thanks for the pull-request!

PR #1390, opened 2025-01-29
Create a single bytecode image with all supported program types

anfredette:


tests/integration-test/bpf/app-test.bpf.c
```diff
@@ -0,0 +1,97 @@
...
+  return 0;
+}
+
+SEC("classifier/tc_pass_test")
+int tc_pass_test(struct __sk_buff *skb) {
+  print_globals(" TC");
+  return TC_ACT_OK;
+}
+
+SEC("classifier/tcx_pass_test")
```
msherif1234, line 49: can we have tcx_ingress and tcx_egress

tests/integration-test/bpf/app-test.bpf.c
```diff
@@ -0,0 +1,97 @@
...
+#include <linux/pkt_cls.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+// clang-format on
+
+volatile const __u8 GLOBAL_u8 = 0;
+volatile const __u32 GLOBAL_u32 = 0;
+
+void print_globals(char *prefix) {
+  bpf_printk("%s: GLOBAL_u8: 0x%02X, GLOBAL_u32: 0x%08X", prefix, GLOBAL_u8,
```
msherif1234, line 15: why u bother with those 2 globals if there is no userspace to set them ?

tests/integration-test/bpf/app-test.bpf.c
anfredette: I'm not sure what you mean.  These functions can be attached in either the ingress or egress directions or both.

tests/integration-test/bpf/app-test.bpf.c
anfredette: We set the globals in integration tests and in sample programs.  That way we can `cat /sys/kernel/debug/tracing/trace_pipe` to confirm they are working as expected.

msherif1234: I see the PR showed only kernel side so I didn't realize userspace already populate them

tests/integration-test/bpf/app-test.bpf.c
msherif1234: since u were generating one BC for all different hook and I care about bidirection TC and TCX hooks hence was suggesting adding them for completeness

tests/integration-test/bpf/app-test.bpf.c
anfredette: I'm using these bpf functions in the sample programs in my load/attach split pr which hasn't merged yet.  At some point, we'll probably create a bpfman integration test that uses them all and an example program/k8s-integration test that uses them all.  However, for now, I just needed something to test all the program types in my draft pr.

tests/integration-test/bpf/app-test.bpf.c
anfredette: I created one for each action.  They can be attached to any hook point in either direction, so I was just keeping them generic.

msherif1234 (approved):

PR #1394, opened 2025-01-30
docs: building docs after a PR merges fails

Billy99:
When a PR merges, a job from docs-build.yml is run. This is currently failing with:

```
  pip install -r requirements.txt
  shell: /usr/bin/bash -e {0}
  :
Reading package lists...
Reading package lists...
Building dependency tree...
Reading state information...
Package python is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  2to3 python-is-python3

E: Package 'python' has no installation candidate
E: Unable to locate package install
Error: Process completed with exit code 100.
```

### DIFF
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -11,6 +11,13 @@ jobs:
     runs-on: ubuntu-24.04
     timeout-minutes: 3
     steps:
+      - name: Install dependencies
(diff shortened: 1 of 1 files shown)


anfredette (approved):
/LGTM

Issue #1402, opened 2025-02-09
Errors when deleting tc and xdp programs in network namespaces


anfredette:
**Error 1:** When the last tc program attached in a network namespace
is deleted, the dispatcher isn't detached and isn't unloaded. This
happens because bpfman needs to enter the network namespace to do
the detach and we aren't currently doing that.

**Error 2:** If a network namespace is deleted with more than one tc
and/or xdp program still attached, bpfman can't delete the
programs. This happens because when one program is deleted, bpfman
rebuilds the dispatcher with the remaining program(s) and tries to
re-attach it. However, if the network namespace is gone, you can't
re-attach the rebuilt dispatcher and the code errors out.

PR #1403, opened 2025-02-09
Fix errors deleting tc and xdp programs in network namespaces

anfredette:
For TC Programs:
When a TC program is attached within a network namespace, switch to that namespace to detach the TC dispatcher after removing the last program.

For TC and XDP Programs:
If TC or XDP programs are attached within a network namespace, avoid attempting to rebuild and reattach the dispatcher if the namespace has already been deleted when the programs are removed.

Fixes: #1402

### DIFF
--- a/bpfman/src/multiprog/tc.rs
+++ b/bpfman/src/multiprog/tc.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Authors of bpfman
 
-use std::{fs, mem, path::PathBuf};
+use std::{ffi::OsStr, fs, mem, os::unix::ffi::OsStrExt, path::PathBuf};
 
 use aya::{
     programs::{
@@ -32,8 +32,8 @@ use crate::{
     },
     utils::{
         bytes_to_string, bytes_to_u16, bytes_to_u32, bytes_to_u64, bytes_to_usize, enter_netns,
-        nsid, should_map_be_pinned, sled_get, sled_get_option, sled_insert,
-        tc_dispatcher_db_tree_name, tc_dispatcher_link_id_path, tc_dispatcher_rev_path,
+        should_map_be_pinned, sled_get, sled_get_option, sled_insert, tc_dispatcher_db_tree_name,
+        tc_dispatcher_link_id_path, tc_dispatcher_rev_path,
     },
 };
 
@@ -51,6 +51,7 @@ const NUM_EXTENSIONS: &str = "num_extension";
 const PROGRAM_NAME: &str = "program_name";
 const HANDLE: &str = "handle";
 const NSID: &str = "nsid";
+const NETNS: &str = "netns";
 
 #[derive(Debug)]
 pub struct TcDispatcher {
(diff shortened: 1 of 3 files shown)


bpfman/src/multiprog/tc.rs
```diff
@@ -32,8 +32,8 @@ use crate::{
     },
     utils::{
         bytes_to_string, bytes_to_u16, bytes_to_u32, bytes_to_u64, bytes_to_usize, enter_netns,
-        nsid, should_map_be_pinned, sled_get, sled_get_option, sled_insert,
-        tc_dispatcher_db_tree_name, tc_dispatcher_link_id_path, tc_dispatcher_rev_path,
+        should_map_be_pinned, sled_get, sled_get_option, sled_insert, tc_dispatcher_db_tree_name,
```
msherif1234, line 35: did u intentionally deleted `nsid` here ?

bpfman/src/multiprog/tc.rs
```diff
@@ -428,26 +434,50 @@ impl TcDispatcher {
...
+                    "TC dispatcher {}, {}, {}, {} successfully detached",
+                    if_name, direction, priority, handle
+                ),
+                Err(_) => debug!(
+                    "TC dispatcher {}, {}, {}, {} not attached when detach attempted",
+                    if_name, direction, priority, handle
+                ),
+            }
+        }
+    }
```
msherif1234, line 479: what about doing the same for TCX and XDP programs ?

bpfman/src/multiprog/tc.rs
anfredette: I did.  We were calling `nsid(netns.clone())?` from nsid below, but the nsid is already saved to the database, so I replaced that with `self.get_nsid()?` which doesn't require the caller to be in the netns to work.

bpfman/src/multiprog/tc.rs
```diff
@@ -428,26 +434,50 @@ impl TcDispatcher {
...
+                            old_handle,
+                            direction,
+                        );
+                    } else {
+                        debug!("enter_netns failed.  The netns may have been deleted.");
+                    };
+                } else {
+                    self.detach_dispatcher(&if_name, attach_type, priority, old_handle, direction);
+                };
             };
```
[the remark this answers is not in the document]
anfredette: enter_netns handles that for us.  It creates a `NetnsGuard` structure with a `drop` function that sets the namespace back to the original when `_netns_guard` goes out of scope.  `_netns_guard` has an `_` because it's not being used in the function, but serves the pupose of executing the `drop` function for us.  This is a common pattern in Rust.  It works kind of like `defer` in Go.

bpfman/src/multiprog/tc.rs
anfredette: We don't have this particular problem with TCX, XDP and Uprobe programs becasue simply deleting the `pin` files causes the programs to be deleted.  TC is different because we have to manage the detaching ourselves when the dispatcher is no longer needed.

anfredette: And, you can delete the pin files without entering the network namespace.

Billy99:
/lgtm
I'll let Mohamed approve since he had some comments

msherif1234 (approved):

PR #1407, opened 2025-02-12
rpm: RPM builds failing

Billy99:
Upstream RPM builds are failing:

```
  --- stderr
  Missing dependency: cmake
  thread 'main' panicked at /builddir/build/BUILD/bpfman-3b1aa1d0c42bd6f2dfef1d7cf27f4b7362307e49/vendor/aws-lc-sys/builder/main.rs:382:40:
  called `Result::unwrap()` on an `Err` value: "Required build dependency is missing. Halting build."
```

aws-lc-sys was added as a dependency in #1406 which requires cmake. Adding cmake to the RPM builds.


bpfman.spec
```diff
@@ -4,18 +4,18 @@
 %global crate bpfman
 
 ## START DO NOT OVERRIDE: These are generated at runtime by .packit.sh
-%global commit 0
-%global shortcommit 0
-%global base_version 0
-%global prerelease 0
+%global commit 3b1aa1d0c42bd6f2dfef1d7cf27f4b7362307e49
```
anfredette, line 7: I'm not sure how this works, but should these values be filled in?

bpfman.spec
```diff
@@ -4,18 +4,18 @@
...
+%global shortcommit 3b1aa1d0
```
anfredette, line 8: ditto

bpfman.spec
```diff
@@ -4,18 +4,18 @@
...
+
 ## END DO NOT OVERRIDE
 
 %global package_version %{base_version}%{?prerelease:~%{prerelease}}
 %global upstream_version %{base_version}%{?prerelease:-%{prerelease}}
 
 Name:           bpfman
 Version:        %{package_version}
-Release:        %autorelease
+Release: 1.20250212105330904645.billy99.rpm.cmake.42.3b1aa1d0%{?dist}
```
anfredette, line 18: ditto

bpfman.spec
Billy99: Removed

bpfman.spec
Billy99: Removed

bpfman.spec
Billy99: Removed

Billy99:
Added a HOLD: Currently, I don't know if this change builds because there is a different error when running locally.

anfredette (approved):
/LGTM

PR #1412, opened 2025-02-13
bpfman release v0.5.6

Billy99:
v0.5.6 release for bpfman repository.


msherif1234:
/LGTM
we no longer need to change the dispatched images as we used to do in the past right? I checked the lib.rs and it uses sha now so we should be good, right ?

Billy99:
> /LGTM we no longer need to change the dispatched images as we used to do in the past right? I checked the lib.rs and it uses sha now so we should be good, right ?

Correct.

dave-tucker (approved):

PR #1417, opened 2025-02-14
adjust building with aws-lc-sys

Billy99:
PR #1406 bumped the version of sigstore from 0.10.0 to 0.11.0. That bump also brought in a dependency on aws-lc-rs. Several subsequent PRs fixed RPM build failures do to aws-lc-rs (#1407 and #1410). Now downstream OpenShift builds of bpfman are failing on certain arches. Attempting a workaround to the downstream failures by changing the asw-lc-rs feature flags.

Ultimately, aws-lc-rs should not be a dependency, see: https://github.com/sigstore/sigstore-rs/issues/436


Billy99:
No longer needed.

PR #1420, opened 2025-02-18
ci: Don't use cross for cross builds

dave-tucker:
Instead of using cross, which uses some images that haven't been updated in over a year, we can instead do the same thing in a matrix build. We run the build in a container, passing the optional `--platform` argument.


Billy99:
> which uses some images that haven't been updated in over a year

Which images haven't been updated in over a year?
I hope you can get this to work, but I never could, hence the use of `cross`.

Issue #1384, opened 2025-01-22
xtask depends on bpfman


dave-tucker:
We need to adjust `xtask` so that it doesn't depend on bpfman.
This was found while working on #1354 and iterating on protobuf definitions.

For example:
- Add a new RPC into `bpfman.proto`
- Run `cargo xtask build-proto`
- At this point bpfman won't compile since the new rpc isn't implemented
- Change the RPC again and run `cargo xtask build-proto`
- This fails since `bpfman` won't compile and it's a dependency of `xtask`

This is due to the manpage and completions generation code that uses `include!` to include a file from `bpfman`
That file is not self-contained 😱 meaning that we add `bpfman_api` and `bpfman` as dependencies.

The fix here is either one of a few options:

1. Make that file self-contained
1. Find a different way of doing manpage/completion generation from `clap`

PR #1422, opened 2025-02-19
chore: Move completion/manpages to hidden cmds

dave-tucker:
Previously we had decided that manpage and completions generation should happen in xtask. This required that bpfman/cli/args.rs be completely standalone such that it can be imported cleanly in xtask.

This hasn't been the case for a while - which means that xtask must depend on bpfman... Which causes xtask to break if bpfman fails to compile (i.e you're working on it) which is not ideal.

Moving this to (hidden) subcommands solves the problem, the only downside being another couple of dependencies in bpfman. This seems like a reasonable trade-off.

Fixes: #1384


scripts/install.sh
```diff
@@ -101,14 +93,8 @@ del_cli_tab_completion() {
...
-    if [ -d ${SRC_MANPAGE_PATH} ] && [ "$(ls -A ${SRC_MANPAGE_PATH})" ]; then
-    #if [ -d ${SRC_MANPAGE_PATH} ] && [ -z "$(find ${SRC_MANPAGE_PATH} -mindepth 1 -maxdepth 1)" ]; then
-        echo "  Copying \"${SRC_MANPAGE_PATH}/*\" to \"${DST_MANPAGE_PATH}/.\""
-        rm ${DST_MANPAGE_PATH}/bpfman*.1  &>/dev/null
-        cp ${SRC_MANPAGE_PATH}/bpfman*.1 ${DST_MANPAGE_PATH}/.
-    else
-        echo "  CLI Manpage files not generated yet. Use \"cargo xtask build-man-page\" to generate."
-    fi
+    del_manpages
+    ${BIN_BPFMAN} man ${DST_MANPAGE_PATH}
```
Billy99, line 97: Below, where `copy_cli_tab_completion` and `copy_manpages` are called, moved those calls to the end of the function. If `setup.sh install` is called after a `setup.sh uninstall`, there is no bpfman CLI binary installed and the commands fail.

Move these down:

```
    echo "Copy CLI TAB Completion files:"
    copy_cli_tab_completion

    echo "Copy Manpage files:"
    copy_manpages
```

bpfman/src/bin/cli/args.rs
```diff
@@ -36,6 +36,10 @@ pub(crate) enum Commands {
     /// eBPF Bytecode Image related commands.
     #[command(subcommand)]
     Image(ImageSubCommand),
+    #[clap(hide = true)]
```
Billy99, line 39: On the fence, wondering if the commands should be hidden or not? We could expose later if we want, so not a blocker.

bpfman/src/bin/cli/args.rs
```diff
@@ -36,6 +36,10 @@ pub(crate) enum Commands {
...
+    Man(crate::manpage::Args),
+    #[clap(hide = true)]
```
Billy99, line 41: I may come back later and add helptext to both commands. Even though it is hidden, manpages are still generated and `sudo bpfman man --help` still outputs data.

Billy99 (commented):
Only comment that needs addressing is in the install.sh file.

scripts/install.sh
dave-tucker: Done

bpfman/src/bin/cli/args.rs
dave-tucker: I did it quickly. I also skipped generating man pages for hidden commands.
Completions will only work for hidden commands if you start typing them, otherwise they aren't suggested.

Billy99 (approved):
/lgtm

Issue #1398, opened 2025-02-04
Add version option or command to bpfman CLI tools


frobware:
Currently, the bpfman CLI tools do not provide a way to check their version (e.g., via bpfman --version or bpfman version). This makes it difficult to identify which version/commit of the tools is running both during development and when users report issues.

frobware:
I explored what build information would be useful and prototyped it here: [build-rs-example](https://github.com/frobware/build-rs-example). This example uses a single build.rs to embed BUILD_VERSION, BUILD_DATE, and BUILD_TOOLCHAIN. Do we need more than that for bpfman? What's unclear to me is how to integrate a single build.rs across all the bpfman binaries.

dave-tucker:
This issue is still relevant

PR #1431, opened 2025-02-19
chore: Add version flag to all binaries

dave-tucker:
Fixes: #1398

Note: I believe this significantly slows down the build process, but it's hard to argue with the utility of it.
You can do `./bpfman -V` to get version information like so:

```
bpfman 0.5.6
branch:add-version
commit_hash:fedf5f3f
build_time:2025-02-19 20:44:35 +00:00
build_env:rustc 1.84.0 (9fc6b4312 2025-01-07),stable-x86_64-unknown-linux-gnu
```


frobware:
I played around with shadow-rs and was surprised about the compile-time hit. I tried just recording a minimum amount in https://github.com/frobware/build-rs-example.

frobware:
Does this work if you were to compile from a source tarball?

PR #1354, opened 2024-12-05
split load and attach

dave-tucker:
This is still a draft PR.
Complete list of remaining work in `TODO.md`.
Opening to let maintainers see the upcoming API changes.

PR #1432, opened 2025-02-19
chore: Unasync all the things

dave-tucker:
When we moved bpfman to be a oneshot commandline program instead of being a daemon we noted that we should run sync and not async given that a lot of the work we do is synchronous.

The blocking factor here was that sigstore and oci_distribution required us to use tokio.

This solve this, we introduce a new wrapper called BlockingWrapper that lets us run:

- Async code from a sync context (i.e oci_utils from bpfman)
- Async code from a sync context from an async context (i.e bpfman-rpc -> bpfman -> oci_utils)

In effect, bpfman can remain sync, while using dependnecies that might require an async runtime like tokio.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -450,7 +450,6 @@ version = "0.5.6"
 dependencies = [
  "anyhow",
  "env_logger",
- "futures",
  "hex",
  "log",
  "netlink-packet-audit",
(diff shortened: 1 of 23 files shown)


dave-tucker:
This was split out of #1354, which already has a LGTM, but we're holding.
Merging this now to help reduce merge conflicts on ☝️

PR #1434, opened 2025-02-20
chore(integration-test): Use rust test framework

dave-tucker:
This uses the standard rust test framework. This requires less dependencies than using our own macros etc... The only downside is that the process to run them (cargo test --no-run and then executing the test binary with sudo -E) is a bit more involved. That said, it's all handled by xtask anyway.

One other benefit would be that we can use multiple test binaries if we wanted to.

Furthermore, all tests no longer "exec" bpfman directly, but instead use the bpfman library to properly test behaviour. In order for this to work correctly database setup had to be moved out of each of the public API functions and the DB and Config passed in by reference.


dave-tucker:
Also pulled out of #1354 = Merging since that has an LGTM and hopefully this will sort out the operator.

PR #1437, opened 2025-02-22
fix(bpfman-rpc): bpfman-rpc performance issues

dave-tucker:
@Billy99 reported that bpfman-rpc was performing poorly since the changes to use only sync code inside bpfman.

There were 2 reasons for this:

1. There's a possibility that the CSI code and the RPC code can both acquire the database lock at the same time. The lock is now protected with a Mutex to ensure this can't happen
2. Within the BlockingWrapper, calling `block_on` with the future performed poorly. Instead, we spawn the future on the runtime and then await it within the block_on

With these changes, bpfman-rpc performance is back where it should be. In debugging this I also noticed that error logs don't propagate to the bpfman-rpc logs in Kubernetes. I've adjusted the code there to ensure that errors are appropriately logged before being returned to the RPC client.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -383,7 +383,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "regex",
- "rustc-hash 1.1.0",
+ "rustc-hash",
  "shlex",
  "syn 2.0.98",
  "which 4.4.2",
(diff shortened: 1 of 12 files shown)


Billy99 (approved):
/lgtm

PR #1440, opened 2025-02-24
fix: correct links to install the crds and operator

hanshal101:
### Description
This PR fixes the old manifests links to install the crds and operator for `bpfman`

### DIFF
--- a/docs/getting-started/running-release.md
+++ b/docs/getting-started/running-release.md
@@ -79,14 +79,14 @@ Next, deploy the bpfman CRDs:
 
 ```console
 export BPFMAN_REL=0.5.6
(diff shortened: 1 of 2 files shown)


Billy99 (approved):
/lgtm

Thank you! Hope this didn't slow you down too much.

Billy99:
Sorry, this can be a bit confusing. At the very top of the docs pages, upper left, is the bpfman logo (icon and "bfpman" text), then there is the release of the docs. That is a pull down. This change did not make the v0.5.6 release, so it is only in main. Change that to "main". I'll see if there is a way to default our docs to main instead of the latest release, which is the natural default.

Billy99:
See #1446

ShyunnY:
Never mind, luckily I found this PR when the installation failed and it solved my problem :)

By the way: it is common to use `latest` as the default version for opening documents, instead of using a fixed version as the default behavior of bpfman currently. We can change this.

PR #1496, opened 2025-03-17
chore: Remove aws-lc-rs dependency

dave-tucker:
See: https://aws.github.io/aws-lc-rs/requirements/linux.html

We can install bindgen-cli on any target that requires it. This avoid having to manage a direct dependency on aws-lc-rs.


Billy99 (commented):
I think `aws-lc-rs` is still in the root Cargo.toml file.

Billy99 (approved):
/lgtm

PR #1497, opened 2025-03-20
chore: Update to Rust 2024 Edition

dave-tucker:
This allows us to take advantage of new language features. More importantly, the next Aya release will have a MSRV of 1.85 since it's also moved to 2024 edition rust, so we need to migrate to remain compatible.


Billy99 (approved):
/lgtm

PR #1498, opened 2025-03-21
bytecode image need to be in "https://quay.io/...." format

msherif1234:
byte code image need to be
`url: https://quay.io/bpfman-bytecode/app-test:latest`
so we can validate the URL at the APIs level


Billy99 (commented):
@msherif1234 So I started reviewing this and my first comment was there are other places that need to change. Search for "quay.io". Lots of docs, examples/Makefile, Makefile Kustomize and config file handling. But then it got me thinking, the standard `pod.spec.container[].image` doesn't use `https://`. I think `url:` should follow `image:`. Is there an API to validate `image`. I looked for 30 secs and didn't see one, but it may be out there. I think we may be opening a can of worms here.

PR #1501, opened 2025-03-24
fix: add binaries to container PATH

### LINKED ISSUE #1451: improve error message for when bpfman-ns is not found
Found this while making a container image from `bpfman` and changing the working directory.

Let's say I have a bpf app with a uprobe program, and the following containerfile:

```dockerfile
FROM quay.io/bpfman/bpfman:v0.56
WORKDIR /mynewdir
COPY myapp myapp_x86_bpfel.o /mynewdir/
```

And I build it and attempt to run it like so:

```sh
docker build -t myapp .
docker run -d --rm --name app --privileged -v /proc:/host/proc -v /sys/fs/bpf:/sys/fs/bpf -e RUST_LOG="Debug" myapp:latest
```

I'll get the following error message:

```
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "./bpfman-rpc": stat ./bpfman-rpc: no such file or directory: unknown.
```

This can be a bit confusing since it mentions  `./bpfman-rpc: no such file or directory`. Turns out this is due to `bpfman-ns`, as `bpfman` looks for it in `"./target/debug/bpfman-ns"`, `"./bpfman-ns"`, and finally in `PATH`: https://github.com/bpfman/bpfman/blob/main/bpfman/src/lib.rs#L1493-L1534 but since the workdir has changed and `/` is not in `PATH`, there is indeed no such file or directory.

Perhaps mentioning the specific file that was not found, or may be catching that specific type of error and returning another message for it could help with this.

monrax:
This closes #1451 

Two changes:

- `Containerfile.bpfman.local`: copy binaries to `/usr/local/sbin/` instead of `/`. This means the binaries are always found, regardless of changes to the current `WORKDIR`.
  - The entrypoint was updated accordingly.
  - For some reason, the build was failing for me unless I also installed `cmake` and `clang` so I went ahead and added those two packages to the last `apt-get install` command.

- `bpfman/src/lib.rs`: print out one additional message when the `bpfman-ns` binary is not found.

### DIFF
--- a/Containerfile.bpfman.local
+++ b/Containerfile.bpfman.local
@@ -54,6 +54,6 @@ RUN apt-get update && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/\* /tmp/\* /var/tmp/*
 
-COPY --from=bpfman-build  ./usr/src/bpfman/bpfman .
+COPY --from=bpfman-build  ./usr/src/bpfman/bpfman /usr/local/sbin/
 
-ENTRYPOINT ["./bpfman-rpc", "--timeout=0"]
+ENTRYPOINT ["bpfman-rpc", "--timeout=0"]
--- a/bpfman/src/lib.rs
+++ b/bpfman/src/lib.rs
@@ -1530,6 +1530,9 @@ pub(crate) fn attach_single_attach_program(root_db: &Db, l: &mut Link) -> Result
                         }
                         Err(e) => {
                             info!("bpfman-ns returned error: {:?}", e);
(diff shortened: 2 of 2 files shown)


Billy99:
@monrax This needs a rebase. Once you rebase you will see that `cmake` and `clang` were already added to the Containerfile. Get it rebased and I'll give it a whirl and merge it.

Billy99 (approved):
/lgtm

PR #1504, opened 2025-03-27
chore(vscode): use path relative to workspace

### LINKED ISSUE #1475: Clean up vscode settings
This can be removed 😆 
```yaml
"yaml.schemas": {
    "https://json.schemastore.org/github-workflow.json": "file:///home/dave/dev/bpfman/.github/workflows/build.yml"
  },
```

Unsure if there's a way to use a path relative to the workspace root here instead, but `/home/dave` doesn't mean anything to anyone other than me - or other daves

monrax:
This closes #1475 

According to RedHat's [vscode-yaml](https://github.com/redhat-developer/vscode-yaml?tab=readme-ov-file#associating-a-schema-to-a-glob-pattern-via-yamlschemas), the path is relative to the workspace and with this change, yaml validation seems to work for me. Please take a look at it and let me know if it works for you as well...

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,7 @@
 {
   "rust-analyzer.check.allTargets": true,
   "yaml.schemas": {
-    "https://json.schemastore.org/github-workflow.json": "file:///home/dave/dev/bpfman/.github/workflows/build.yml"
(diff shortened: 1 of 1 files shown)


monrax:
closing for kubecon

PR #1512, opened 2025-04-01
bpfman: invalid link fixes

Billy99:
If processing a `bpfman attach` failed in certain scenarios, the link would be added to the program but not to the DB. Once that happens, the program cannot be unloaded.

Example, load an XDP program and then try to attach to an invalid interface.

Step one is in the attach_program() function, once `prog.add_link()` is called, don't exit the function without calling `prog.remove_link()`.

Step two, when in the Program object and walking through the links to build up a list of link objects, if the link is not in the DB, skip over it because there is nothing we can about it at that point. We don't have enough data to detach.

### DIFF
--- a/bpfman/src/lib.rs
+++ b/bpfman/src/lib.rs
@@ -460,12 +460,14 @@ pub fn attach_program(
     let name = prog.get_data().get_name().unwrap_or("not set".to_string());
     info!("Request to attach {kind} program named \"{name}\" with id {id}");
 
-    // Write attach info into the database
+    // Write attach info into the database. Once written to the database,
+    // DO NOT EXIT with a failure without calling prog.remove_link().
     let mut link = prog.add_link()?;
-    link.attach(attach_info)?;
 
-    let result = attach_program_internal(config, root_db, &prog, link.clone());
-
+    let result = match link.attach(attach_info) {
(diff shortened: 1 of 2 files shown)


bpfman/src/lib.rs
```diff
@@ -460,12 +460,14 @@ pub fn attach_program(
...
+    // DO NOT EXIT with a failure without calling prog.remove_link().
     let mut link = prog.add_link()?;
-    link.attach(attach_info)?;
-
-    let result = attach_program_internal(config, root_db, &prog, link.clone());
 
+    let result = match link.attach(attach_info) {
+        Ok(_) => attach_program_internal(config, root_db, &prog, link.clone()),
+        Err(e) => Err(e),
+    };
```
anfredette, line 470:
```suggestion
let result = link.attach(attach_info)
    .and_then(|_| attach_program_internal(config, root_db, &prog, link.clone()));
```
Just a nit.  What you have works.  I think this would work too and would be a little cleaner.  Your choice.

anfredette (commented):
Ok.  LGTM otherwise.  Do you want to try the suggestions, or just submit as-is?

anfredette (approved):

Issue #1479, opened 2025-03-13
Use `uv` for the docs


dave-tucker:
- Add a `pyproject.toml` similar to: https://github.com/aya-rs/book/blob/main/pyproject.toml
- Remove `requirements.txt` since we'll rely on `uv.lock`
- Similar changes as per this PR to use `uv` in Netlify + bump the Python Version: https://github.com/aya-rs/book/commit/09319a48b09df9232c4a37794f0bafc504b455a2
- Update documentation for usage of `uv` https://github.com/bpfman/bpfman/blob/main/docs/developer-guide/documentation.md
- Enable the experimental support for `uv.lock` in CI: https://github.com/aya-rs/book/blob/main/.github/dependabot.yml

mdaffad:
Hi, can i contribute on this issue?

dave-tucker:
Hi @mdaffad, yes please! We'd be happy to accept contributions

dave-tucker:
Closed via #1530

PR #1530, opened 2025-04-20
ci: Integrate uv as package manager for documentation build

mdaffad:
Ref: #1479 

- [X] Add a pyproject.toml similar to: https://github.com/aya-rs/book/blob/main/pyproject.toml
- [X] Remove requirements.txt since we'll rely on uv.lock
- [X] Similar changes as per this PR to use uv in Netlify + bump the Python Version: https://github.com/aya-rs/book/commit/09319a48b09df9232c4a37794f0bafc504b455a2
- [X] Update documentation for usage of uv https://github.com/bpfman/bpfman/blob/main/docs/developer-guide/documentation.md
- [X] Enable the experimental support for uv.lock in CI: https://github.com/aya-rs/book/blob/main/.github/dependabot.yml


dave-tucker:
@mdaffad thanks for the contributuion. CI is failing due to the toml formatting.
https://github.com/bpfman/bpfman/actions/runs/14559803067/job/40846188475?pr=1530
If you could fix that then this should be good to go!

.github/workflows/build.yml
```diff
@@ -415,16 +415,14 @@ jobs:
       - name: Checkout bpfman
         uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # @v4
 
-      - name: Install Python3
-        uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # @v5
+      - name: Install required-version defined in pyproject.toml
+        uses: astral-sh/setup-uv@v5
         with:
-          python-version: 3.x
+          pyproject-file: "./pyproject.toml"
```
dave-tucker, line 421: nit: Is this required?

mdaffad:
> @mdaffad thanks for the contributuion. CI is failing due to the toml formatting. https://github.com/bpfman/bpfman/actions/runs/14559803067/job/40846188475?pr=1530 If you could fix that then this should be good to go!

Okay, thanks for your reply. I will fix it asap

.github/workflows/build.yml
mdaffad: I will see their workflows docs first. I think it should be the default

mdaffad:
Updated, checked with pre-commit

dave-tucker (approved):

Issue #386, opened 2023-03-31
Rearchitect the Bpfd Agent 


astoycos:
With the addition of new interface selectors in #362 the complexity of the bpfd-agent is starting to balloon.  This Issue will track a redesign which will hopefully make the code much less error prone and much easier to parse. The below image shows some preliminary design for this: 

<img width="487" alt="image" src="https://user-images.githubusercontent.com/23533892/229195742-8b51dcc6-c8bb-431b-8889-7f0332017a04.png">

astoycos:
Being fixed in #342

PR #1533, opened 2025-04-21
fix: change program key to application key to align with CRD

mdaffad:
Changes:
1. Follow fix on #386


dave-tucker (approved):
Code changes LTGM. @anfredette is there an easy way to test this against the k8s operator to check that everything still works as expected?

dave-tucker:
added hold label until we've tested this with the k8s operator too to make sure everything still works

mdaffad:
Hi, let me know if you need something from me. I'd like to help with my best.

PR #1536, opened 2025-04-23
chore: Fix docs build

dave-tucker:
This fixes the docs publish pipeline to use UV instead of pip. It also removes transitive dependencies from the pyproject.toml to tidy it up.


dave-tucker:
Skipping CI and merging since this PR only pertains to the docs build, which is triggered on merge only,

Issue #1537, opened 2025-04-23
Dispatchers and BPF state not always cleaned up after CRD delete


anfredette:
This issue was found when testing on a ROSA cluster with config/samples/bpfman.io_v1alpha1_clusterbpfapplication_with_intf_discovery.yaml.

Intermittent errors were found when applying and deleting the above yaml.  There are usually some attach errors due to issue #1528, but typically the XDP program seems to be attached to most discovered interfaces.  However, sometimes the program and some attachments are not completely deleted when the yaml is deleted.

PR #1539, opened 2025-04-23
Fix get_dispatcher() function

anfredette:
The issue was that the `contains` check in `get_dispatcher` could match multiple dispatchers. For example, a Dispatcher ID of `xdp_dispatcher_4026531840_2` is contained in `xdp_dispatcher_4026531840_20` and `xdp_dispatcher_4026531840_25`.  As a result, the code could get the wrong dispatcher, causing dispatchers to be unexpectedly deleted or hijacked, and that caused problems. Whether it happened depended on the order in which the interfaces were discovered and processed.

We didn't see this issue before because we never tested many interfaces in any given network namespace.

The above bug also exposed a couple of corner cases in `detach_multi_attach_program`, which were tightened up.  These cases should never have happened, but they did because of the above bug.

Fixes: #1537


bpfman/src/lib.rs
```diff
@@ -1801,25 +1801,29 @@ fn detach_multi_attach_program(
...
+    debug!(
+        "netns_deleted = {netns_deleted}, number_attached_programs = {number_attached_programs}"
+    );
 
-    let next_available_id = if netns_deleted {
+    // Typically, we'd expect the number of attached programs to be greater than
+    // 0 because we're calling the detach function.  However, if
+    // num_attached_programs() can't find the dispatcher for the given did, it
+    // will return 0, so we need to handle that case.
+    let next_available_id = if number_attached_programs == 0 {
```
dave-tucker, line 1813: This isn't easy to read  even with the comment.
You could write this as:
```rust
let next_available_id = num_attached_programs(&did, root_db)?.saturating_sub(1);
```
That would require num_attached_programs to return a `u32` - which I don't think is unreasonable since we'd never expect a negative number of programs. IIRC it was a usize because we use it to index something at some point an wanted to avoid the `as` cast.

bpfman/src/lib.rs
```diff
@@ -1801,25 +1801,29 @@ fn detach_multi_attach_program(
...
-        num_attached_programs(&did, root_db)? - 1
+        number_attached_programs - 1
     };
     debug!("next_available_id = {next_available_id}");
 
     let mut old_dispatcher = get_dispatcher(&did, root_db)?;
 
-    if let Some(ref mut old) = old_dispatcher {
-        if next_available_id == 0 {
+    if next_available_id == 0 || netns_deleted {
```
dave-tucker, line 1822: Do we even need `netns_deleted` at all?
If so, a comment here explaining this condition would be helpful.
Perhaps move the comment from above to here where it's more relevant.

dave-tucker (changes requested):

bpfman/src/lib.rs
anfredette: Done.

bpfman/src/lib.rs
anfredette: Done.

anfredette:
@dave-tucker I reworked `detach_multi_attach_program` to hopefully make it even more clear, but you can be the judge of that :-).

dave-tucker (approved):

PR #1673, opened 2026-07-07
Import the Go implementation of bpfman

frobware:
This PR makes the Go rewrite of bpfman the implementation this repository builds, tests, and publishes. After it merges, `quay.io/bpfman/bpfman:latest` is the Go daemon. It delivers the direction [proposed and agreed here](https://kubernetes.slack.com/archives/C04UJBW2553/p1769528836071779), also raised in the community meeting with no objections.

The change is too large for a line-by-line review, so the series and this description are structured to be checkable instead: a provable import, followed by small integration commits you can review conventionally, plus an explicit statement of what is preserved and what is known to differ.

## How the import is structured

The series separates "what we imported" from "what we changed to integrate it":

1. **Import** -- a pristine snapshot of the Go implementation at `github.com/frobware/go-bpfman@6397a5c5`. Every file is byte-for-byte identical to that commit: extract `git archive` of both trees and diff to verify. Nothing else in this commit.
2. **Absorb the example applications into the root module** -- the repository's go.mod previously existed to build the Go-based examples; the import replaces it with the Go implementation's, which carries far more dependencies. Rather than giving the examples a nested module of their own, both the implementation and the examples build from the one top-level go.mod, which therefore takes on the examples' operator and Kubernetes dependencies (controller-runtime pinned where the operator's preference cannot compile against the newer k8s floor). bpf2go is declared via the go.mod tool directive so vendor-mode `go generate` resolves it, and the example image builds move to a Go 1.25 base to match the module's toolchain directive.
3. **Merge the Rust ignore rules back into the imported .gitignore.**
4. **Reconcile the lint configurations** -- the Go lint settings are adopted from the import: the existing `.golangci.yaml` gives way to the imported configuration (the one place a repository convention changes); yamllint and taplo are scoped around test fixtures and the vendor tree.
5. **Remove the Rust builds from the build workflow** -- pruned in place, keeping the workflow and surviving job names so the documented release process is untouched. The release tarballs become static cross-compiled Go builds under the same artefact names.
6. **Adapt image-build.yml to the Go implementation** -- same workflow name, job structure, image names, tag rules, and cosign signing; the Go build substitutes where the Rust build fed it. Details below.
7. **Run the Go CI checks on pushes to main** -- so merged commits get the same coverage as pull requests.

Each integration commit's message carries the full rationale.

To verify the import commit yourself:

```
git archive <import-commit> | tar -x -C /tmp/import
git -C <go-bpfman clone> archive 6397a5c5 | tar -x -C /tmp/go-bpfman
(cd /tmp/go-bpfman && find . -type f | while read -r f; do
  cmp -s "$f" "/tmp/import/$f" || echo "differs: $f"
done)
```

This prints nothing when every imported file is byte-identical.

## The Rust implementation stays (for now)

This PR stops building the Rust implementation but does not delete it. Removing ~25k lines of Rust in the same change that adds the Go implementation would make an already large PR unreviewable, and keeping the source in place makes rollback trivial while the rewrite soaks. A follow-up PR removes the Rust tree after a soak period. Two safety notes: the operator pins released images, so nothing consumes `latest` in anger, and we can cut a `last-known-working-bpfman` branch before merging as a belt-and-braces rollback point. The real test is the next release.

The bytecode pipeline is the one place the Rust toolchain remains in use: the integration-test and dispatcher bytecode objects are still built with cargo xtask, unchanged.

## Compatibility

What the PR preserves, stated directly since no diff can show it:

- **External behaviour.** The CLI grammar moved to a noun-verb structure ([discussed here](https://kubernetes.slack.com/archives/C04UJBW2553/p1770805319115229)); note this is a breaking change for CLI users.
- **gRPC API.** `proto/bpfman.proto` is byte-identical between the two implementations, and the `clients/gobpfman/v1` stubs keep their import path, so the operator's integration is unchanged.
- **Images and release artefacts.** Multi-arch images (amd64, arm64, ppc64le, s390x) under the same names and tag rules; release tarballs under the same `bpfman-linux-*.tar.gz` names; cosign signing retained.
- **Configuration and deployment.** Configuration is read from the same place in the same format (`/etc/bpfman/bpfman.toml`, TOML, with built-in defaults when absent), and the operator deploys and drives the Go daemon unchanged.

## Known differences

From a driven comparison of the two CLIs ([June 2026 parity run](https://github.com/frobware/go-bpfman/blob/1e1b964a728cf4cf5f98419f6300343e70d5a67e/README.md), re-checked against the imported tree). Behavioural differences first, direction noted where it is settled:

- Map sharing via `--map-owner-id` actually shares the kernel map in the Go implementation; in Rust it records the owner relationship but loads a fresh map (verified with bpftool: same `map_ids` on Go's owner and sharer, distinct on Rust's). This one's direction is settled: Rust is broken, Go behaves as the flag documents.
- Equal-priority dispatcher ordering: Go's tie-break is deterministic across runs; Rust's is not. Equal-priority ordering is implementation-defined either way -- use distinct priorities for a defined order.
- Error exit codes differ for some failure classes. Both implementations reject the same invalid inputs and end in the same state, but the numeric codes are not aligned (e.g. Rust's panic 101 and clap usage 2 against Go's codes).
- Bytecode extracted from image pulls is not yet cleaned up. This is also an open issue in the Rust implementation; the rework is planned alongside making managed programs own their bytecode.

CLI capabilities the Go implementation adds (supersets; nothing Rust accepts is rejected):

- `-o json` / `-o text` on every get/list verb; Rust is text-only.
- `--global` accepts `0x`-prefixed hex values as well as the bare form Rust requires.
- `--container-pid` attachment works; Rust marks it not currently supported.

Findings from the parity run that need no action: the XDP test objects now use the canonical `SEC("xdp")` form both loaders accept; Rust's XDP `--priority` help text says `1-1000` but both implementations accept 0 identically (a Rust documentation nit, no divergence); and the duplicate-TCX-attach failure has the same class and exit status on both sides, with Go giving the more actionable message.

### What's now in the image

The old image (Containerfile.bpfman.multi.arch) carried three binaries -- `bpfman`, `bpfman-ns`, `bpfman-rpc` -- on an ubuntu:24.04 base, with `bpfman-rpc --timeout=0` as the entrypoint. The new image carries a single `/bpfman` binary with `/bpfman` as the entrypoint. The published image is the dynamically linked binary on fedora-minimal, whose glibc matches the Fedora builder; building with `STATIC=1` switches the base to ubi-minimal, which a static binary can ride because it has no runtime libc dependency. Note the base distribution changes from Ubuntu to Fedora -- relevant if you scan images against a distribution CVE feed or debug in-cluster (microdnf, not apt). The daemon is the `bpfman serve` subcommand, replacing the separate bpfman-rpc binary; the bpfman-ns namespace-entry mechanism is integrated into the binary rather than shipped beside it. The operator drives the new image unchanged.

The release tarball follows the same consolidation. The v0.6.0 `bpfman-linux-*.tar.gz` shipped five binaries (`bpfman`, `bpfman-rpc`, `bpfman-ns`, `bpf-metrics-exporter`, `bpf-log-exporter`); the tarball this PR produces ships the one `bpfman` binary. The two exporters were never part of the images, only the tarball; if they are still in use somewhere, restoring them is a follow-up -- flagging that as a gap until confirmed either way.

The daemon state store is SQLite behind a single driver (mattn/go-sqlite3). The schema is managed by goose migrations embedded in the binary and run at startup, so the database upgrades in place rather than being wiped on restart, and nothing has to package migration files alongside the binary.

## Testing

The Go implementation keeps the gRPC server and conventional Go unit tests, but the centre of gravity of the test corpus is `e2e/scripts`: 123 `.bpfman` scripts that drive the system through the bpfman CLI and assert on its output. This raises the abstraction level of the tests to the surface users actually touch -- load, attach, list, inspect via the CLI against real kernels -- rather than asserting on internals. That matters for where the project is heading: we anticipate the gRPC server eventually going away, and CLI-level scripts remain a stable test surface across that change.

In CI this shows up as: the unit and e2e matrix on amd64 and arm64 with and without the race detector, the script suite driven from both file and image bytecode sources, cross-architecture (s390x, ppc64le) transport tests under QEMU, a nix build gate, and lint/format/vet/vendor checks. All of it runs on pull requests and on pushes to main.

## Developer workflow

The imported Makefile drives both development and CI, and `make help` catalogues it. The CI jobs are deliberately thin: each docker-using job in the Go CI workflow is a shell over the corresponding `make ci-*` target, so what passes in CI is what passes locally (`make ci` runs the lot) and a CI failure reproduces on your machine with one command. Day to day: `make build-all` and `make test` for the edit-compile-test loop, `make build-image-dev` for a fast current-arch image from a host-built binary, `make build-image` for the canonical cross-compiled image the release pipeline ships, and the `make test-e2e*` tiers (root required) up to `make test-all` as the pre-push gate.

## Image build and verification

The daemon image is built by `make build-image`, which bakes attestation metadata into the binary (image reference, signer identity, OIDC issuer), so `bpfman version` prints a ready-to-run `cosign verify` command for the image it shipped in. After every push to main, verification jobs pull the published image per architecture, assert the version stamp against the commit, and verify the cosign signature both directly and via the attestation the binary carries. Every image reference derives from a single `IMAGE_REGISTRY` workflow variable pointing at quay.io.

The pipeline was proven end to end from a fork before this PR: the full image set (daemon, userspace examples, bytecode, dispatchers -- 32 images, each a four-arch index) published to a scratch registry, signed, and verified.

The daemon image it produced was then verified by hand, as direct evidence that the multi-arch story survives the switch to the Go build.

### Multi-arch manifest check

```
skopeo inspect --raw docker://ttl.sh/bpfman/bpfman:latest | jq .
```

The raw manifest is an OCI image index (`application/vnd.oci.image.index.v1+json`) containing four platform manifests:

```
linux/amd64
linux/arm64
linux/ppc64le
linux/s390x
```

The index also carries four `unknown/unknown` entries. These are the buildx attestation manifests (provenance/SBOM), one paired with each platform manifest. They are expected and harmless.

### Running each platform variant

With qemu binfmt handlers registered, each variant was run with `version` as the command:

```
podman run --rm ttl.sh/bpfman/bpfman:latest version
podman run --rm --platform linux/arm64 ttl.sh/bpfman/bpfman:latest version
podman run --rm --platform linux/ppc64le ttl.sh/bpfman/bpfman:latest version
podman run --rm --platform linux/s390x ttl.sh/bpfman/bpfman:latest version
```

All four ran successfully, each reporting its own platform, confirming the index contains genuinely distinct per-arch binaries rather than one binary duplicated across manifests:

```
Platform:    linux/amd64
Platform:    linux/arm64
Platform:    linux/ppc64le
Platform:    linux/s390x
```

Beyond the platform line, all four printed identical build metadata, confirming a single workflow run produced all variants:

```
Version:     a956f1b
Git commit:  a956f1bf8eb4daebe8c7bbd69a5694c6bb274000
Git branch:  main
Git state:   dirty
Build date:  2026-07-07T10:20:57+01:00
Go version:  go1.25.11 X:nodwarf5
Attestation: cosign verify ttl.sh/bpfman/bpfman:latest --certificate-identity 'https://github.com/frobware/bpfman/.github/workflows/image-build.yml@refs/heads/main' --certificate-oidc-issuer https://token.actions.githubusercontent.com
```

### Operator integration

The same image was used in a KIND cluster to run the bpfman-operator's `test-integration-local` make target, which passed.

### Open question

Every variant reports `Git state: dirty`. A CI build from a clean checkout should not have a dirty tree, so something in the build (generated files, bpf2go output, vendoring) modifies the tree before the version stamp is computed. Cosmetic -- it does not affect the commit, date, or attestation stamps -- but worth investigating in the workflow.

## Follow-up work

Planned as separate PRs after this one: removal of the Rust tree once the soak period passes; a documentation rewrite, since docs/ still describes the Rust daemon (bpfman-rpc, the old CLI grammar) and the release process notes need updating for the consolidated binary; the exporters, if they turn out to be in use; and the `Git state: dirty` stamp investigation above.

## Review guide

Suggested layers, roughly matching the commits: the import mechanics above (commit 1 is verifiable, not reviewable); CI and release mechanics (commits 5-7); the module and lint integration (commits 2-4); then the implementation itself, best approached top-down from `cmd/bpfman`, the manager package, and `e2e/scripts` as executable documentation of behaviour.


dave-tucker (approved):
LGTM! Thanks for all the work on this @frobware.
I know this snowballed quickly from "lets use SQLite" to "actually lets rewrite it in Go" but appreciate the persistence. Hoping this will help unlock some new features and community contributions! 😎

PR #1676, opened 2026-07-09
Add support for eBPF LSM programs

frobware:
Adds eBPF LSM program support -- load, attach, detach, and list a `BPF_PROG_TYPE_LSM` program whose hook (for example `file_open`) is fixed at load time from the `lsm/<hook>` ELF section. This ports the LSM work [Dave Tucker](https://github.com/dave-tucker) did in [frobware/go-bpfman#109](https://github.com/frobware/go-bpfman/pull/109); the feature is his.

The go-bpfman core has moved a long way since that PR -- `ProgramType` is a plain string now, the store is driven by goose migrations, and discovery and the shell were reorganised -- so this is a re-implementation against the current tree rather than a cherry-pick. It is also the first change to alter the schema and exercise the goose migration path end to end: `00002_add_lsm.sql` adds a `link_lsm_details` table, and that is the whole migration -- `program_type` and `links.kind` are plain `TEXT` columns whose permitted values the domain enforces, so adding a type needs no constraint change. Rather than add Dave's separate `hook_name` column, the hook reuses the `attach_func` field fentry and fexit already thread from ELF section to `AttachTo`, so the programs table is untouched.

## Show and tell

The type is represented faithfully at every boundary. `TestLsm_LoadAndGet.bpfman` pins the load record exhaustively -- an added or renamed field fails the test:

```
assert $prog.record.load matches exhaustive {
    ...
    program_name: test_lsm
    program_type: lsm
    attach_func:  file_open
    ...
}
```

and the link carries the hook, matched exhaustively too:

```
assert $link.record.details matches exhaustive {
    hook_name: file_open
}
```

Behaviour is proven, not taken on faith. `file_open` fires for every open on the box, so a global counter is meaningless. The new `lsm` shell fixture leases a probe -- a unique marker comm and a target file -- and opens that file under the marker; the program counts only that comm, so the count is exactly the fixture's opens and nothing else. `TestLsm_LinkRoundTrip.bpfman` fires 50 opens, checks the count, detaches, fires again, and checks the count is frozen:

```
lsm hook=file_open marker=xwfzelqa count=50 want=50
lsm after-detach count=50 want=50
```

Zero before, exactly 50 after, still 50 once detached: the program observed precisely the fixture's opens, and detach really stopped it. Both scripts pass:

```
--- PASS: TestBPFManScripts/scripts/TestLsm_LoadAndGet.bpfman
--- PASS: TestBPFManScripts/scripts/TestLsm_LinkRoundTrip.bpfman
```

## Test plan

- `make test` -- the LSM manager lifecycle (load/attach/detach/unload), the link and program exhaustiveness tests, and the sqlite test that applies `00002_add_lsm.sql` forward and back.
- `make test-e2e-scripts TEST='TestBPFManScripts/scripts/TestLsm_'` -- both scripts, shown above.
- `make test-e2e-grpc` -- the parallel lifecycle test gains an `lsm` spec cycling the type through the real daemon.
- Attaching an LSM program needs `bpf` in the active LSM list (`/sys/kernel/security/lsm`). Hosted CI runners ship `CONFIG_BPF_LSM=y` but not the active entry, and cannot add it without a reboot, so the gRPC spec skips at runtime and the scripts carry a `requires-bpf-lsm` pragma the runner skips on -- they run wherever the kernel supports it and skip cleanly where it does not.
- The `nested-vm-e2e / e2e (nested Fedora VM, LSM)` job boots a guest kernel with bpf in the active LSM list, so the LSM tests run for real in CI rather than skipping. On this PR's head the job runs the scripts corpus with `STRESS_COUNT=5`: both scripts passed on all five iterations, the gRPC `lsm` spec passed, and the log contains no lsm skips:

  ```
  --- PASS: TestParallel_GRPC/lsm (38.14s)

  scripts_test.go:222: lsm hook=file_open marker=bhnpuqgm count=50 want=50
      lsm after-detach count=50 want=50
  --- PASS: TestBPFManScripts/scripts/TestLsm_LoadAndGet.bpfman (0.53s)
  --- PASS: TestBPFManScripts/scripts/TestLsm_LinkRoundTrip.bpfman (1.12s)
  ```

### DIFF
--- a/attach_kind.go
+++ b/attach_kind.go
@@ -23,6 +23,8 @@ func attachVerb(spec AttachSpec) (verb string, accepts []ProgramType) {
 		return "fentry", []ProgramType{ProgramTypeFentry}
 	case FexitAttachSpec:
 		return "fexit", []ProgramType{ProgramTypeFexit}
+	case LsmAttachSpec:
+		return "lsm", []ProgramType{ProgramTypeLsm}
 	case XDPAttachSpec:
 		return "xdp", []ProgramType{ProgramTypeXDP}
 	case TCAttachSpec:
--- a/attach_spec.go
+++ b/attach_spec.go
@@ -234,6 +234,27 @@ func (FexitAttachSpec) attachSpec() {}
 // ProgramID returns the kernel ID of the program to attach.
 func (s FexitAttachSpec) ProgramID() kernel.ProgramID { return s.programID }
 
+// LsmAttachSpec specifies how to attach an LSM program.
+// Note: the hook comes from the program's stored metadata (set at load
+// time from the lsm/<hook> ELF section), not from user input.
+type LsmAttachSpec struct {
+	attachMetadata
+	programID kernel.ProgramID
+}
+
+// NewLsmAttachSpec creates a LsmAttachSpec with validated fields.
+func NewLsmAttachSpec(programID kernel.ProgramID) (LsmAttachSpec, error) {
+	if programID == 0 {
+		return LsmAttachSpec{}, errors.New("programID is required")
+	}
+	return LsmAttachSpec{programID: programID}, nil
+}
+
+func (LsmAttachSpec) attachSpec() {}
+
+// ProgramID returns the kernel ID of the program to attach.
+func (s LsmAttachSpec) ProgramID() kernel.ProgramID { return s.programID }
+
 // XDPAttachSpec specifies how to attach XDP.
 type XDPAttachSpec struct {
 	attachMetadata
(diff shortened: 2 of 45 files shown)


dave-tucker:
@frobware needs a rebase now! but overall LGTM, thanks for porting this over 🫡

frobware:
Rebased and CI is green again.

frobware:
/hold

Want to rebase this on to #1688.

frobware:
This PR depends on #1690 for CI coverage of the lsm tests: GitHub-hosted runners boot without `bpf` in the active LSM list, so the lsm e2e tests skip there. #1690 adds a nested Fedora VM job whose kernel has bpf-LSM active; once it merges and this PR rebases onto it, the `TestLsm_` scripts and the lsm gRPC lifecycle run in CI with no further wiring -- the VM job's driver already runs the full corpus.

PR #1686, opened 2026-07-21
Reset the attach type when loading programs as extensions

frobware:
bpfman loads XDP and TC programs as `BPF_PROG_TYPE_EXT` (freplace) bound to a dispatcher, so each user program replaces a stub function in the dispatcher rather than attaching to a hook directly. cilium/ebpf resolves that freplace target from the program's `(Type, AttachType)` pair and recognises only `(Extension, AttachNone)`, then looks the function up by name in the target BTF. It takes `AttachType` from the ELF section, and the `xdp.frags` section yields `AttachType == XDP`. That pair has no mapping, so cilium refuses the load with `attach Extension/XDP: unrecognized attach type` before any syscall, and a program compiled into an `xdp.frags` section fails to load at all. Plain `xdp` sections are unaffected, since their attach type is already `AttachNone`.

This clears the attach type when converting a program to an extension, so the freplace target resolves by function name. bpfman-rust does not need this: aya loads freplace components through its dedicated `EbpfLoader::extension()` API, which resolves the target by BTF function name and loads the program as `BPF_PROG_TYPE_EXT` without carrying an attach type. cilium/ebpf instead reuses the XDP `ProgramSpec` -- attach type and all -- as the extension, so go-bpfman must clear the attach type itself. The same `xdp.frags` fixture that go-bpfman rejects today loads cleanly under bpfman-rust 0.6.0, confirming this is a go-bpfman-specific gap rather than a kernel or frags limitation.

The change is independent of XDP multi-buffer (frags) support proper and of the ERANGE generic-mode fallback: it is a load-path fix, whereas those are attach-path concerns. It does not make bpfman honour frags on the dispatcher; it only lets an `xdp.frags` program load and run.

## Test plan

A new e2e script, `TestLoad_XDPFragsProgram.bpfman`, loads an `xdp.frags` program, attaches it to a normal-MTU netns veth, confirms it is the dispatcher's sole member, then pings across the veth and asserts the program's per-CPU packet counter advances -- proving it is on the packet path, not merely wired up. Without the reset the script fails at the load step with `unrecognized attach type`. Run with:

`make test-e2e-scripts-file TEST='TestBPFManScripts/scripts/TestLoad_XDPFragsProgram\.bpfman$'`

Plain XDP, TC, and TCX load and attach through the same extension path is unchanged (`TestAttach_ResolvesIfaceInNetns` still passes).

### DIFF
--- a/e2e/scripts/TestLoad_XDPFragsProgram.bpfman
+++ b/e2e/scripts/TestLoad_XDPFragsProgram.bpfman
@@ -1,0 +1,57 @@
+# -*- mode: bpfman -*-
+#
+# Given: an XDP program compiled into the "xdp.frags" section (so it
+# carries BPF_F_XDP_HAS_FRAGS) that counts packets, and a normal-MTU
+# netns veth pair
+#
+# When: it is loaded and attached through bpfman
+#
+# Then: the load succeeds, the program is the dispatcher's sole member,
+# and it counts traffic on the packet path. bpfman loads XDP programs as
+# BPF_PROG_TYPE_EXT against a dispatcher; the loader must reset the
+# section-derived attach type when it does so, or the load is refused
+# with "unrecognized attach type" because the xdp.frags section leaves a
+# concrete attach type on the spec that has no extension mapping in
+# cilium/ebpf.
+
+guard veth <- net netns-veth-pair
+defer net release $veth
+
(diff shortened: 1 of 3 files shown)


frobware:
cc @alebedev87

PR #1690, opened 2026-07-24
Add a nested Fedora VM e2e harness and deterministic link-detach teardown

frobware:
This PR does three things: it makes `bpfman link detach` deterministic for link types the kernel releases asynchronously, it runs the whole e2e suite inside a nested Fedora VM in CI so that bpf-LSM tests become runnable, and it makes that VM a general-purpose developer tool you can drive from the command line.

## Deterministic detach

For link types with no kernel-side detach operation (perf-event and tracing links: kprobe, uprobe, tracepoint, fentry/fexit, LSM) the kernel tears the link down asynchronously after the pin is removed and the last FD closed, so a program can keep firing on its hook after `bpfman link detach` returns. Under CPU contention on a 4-vCPU runner that window stretched far enough for an e2e counter to record all fifty post-detach probe fires. `DetachLink` now captures the kernel link ID and polls `BPF_LINK_GET_FD_BY_ID` until the object is gone, distinguishing alive (an FD comes back), dying (`EAGAIN`: refcount zero, deferred release pending) and gone (`ENOENT`); the wait is bounded and skipped entirely for types whose kernel `Detach()` already disconnected the program synchronously.

Even a fully released link ID cannot prove the program is off its hook -- the trampoline unlink drains behind RCU grace periods with no userspace-visible completion -- so the script corpus stops trusting teardown timing altogether. The fixed `sleep 0.1` settles after detach become quiescence polls that fire the script's own probe and retry until the detached program's counter stops moving; the multi-prog wave scripts convert their fixed whole-run totals into snapshot deltas around each poll (a strictly stronger assertion); and the TCX duplicate-attach script reads kernel truth through the `linkinfo` builtin, which looks the link up by id via the `bpf()` syscall and is immune to the enumeration race against links dying elsewhere in the parallel corpus.

## A nested VM in CI

GitHub-hosted runners boot without `bpf` in the active LSM list and cannot be rebooted mid-job, so nothing that needs a bpf-LSM kernel can run on them. The new `nested-vm-e2e.yml` workflow boots a Fedora guest -- stock Fedora kernels ship bpf-LSM active -- and runs the build, the unit tests, the full gRPC e2e suite and the whole `.bpfman` script corpus inside it. Nothing here is LSM-specific; the VM simply provides the kernel the lsm tests in #1676 need, so once that PR rebases onto this one its `TestLsm_` scripts run with no further wiring.

The workflow has no triggers of its own: `CI` calls it as a job, so it runs on pull requests and pushes to main and counts towards `ci-complete`. Provision, build and test run as separate steps, each booting a cached provisioned disk. On the runner it builds virtiofsd 1.13.2 from crates.io (noble packages 1.10, which predates `--translate-uid`), opens `/dev/kvm`, lifts Ubuntu 24.04's AppArmor confinement of unprivileged user namespaces and raises file limits. A steady-state run restores the base image, the provisioned disk and the guest Go build cache from `actions/cache` and completes in about six minutes.

## The VM as a developer tool

The harness underneath, `hack/fedora-vm.sh`, is not CI-shaped -- it is a general way to run anything in a throwaway Fedora guest with your working tree shared in at its host path:

```
hack/fedora-vm-host-deps.sh        # host tooling, once (Fedora and Ubuntu/Debian)

# Interactive shell in the guest.
hack/fedora-vm.sh --provision hack/install-fedora-deps.sh

# The CI run, locally.
hack/fedora-vm.sh --provision hack/install-fedora-deps.sh --run hack/nested-vm-e2e.sh

# One-off command; exit status propagates.
hack/fedora-vm.sh --run 'uname -r && cat /sys/kernel/security/lsm'

# Share extra host directories, docker -v style.
hack/fedora-vm.sh -v /path/on/host:/path/in/guest:ro --run '...'
```

The guest disk is a per-boot overlay, so only the shares persist: packages you want permanently belong in `install-fedora-deps.sh`, whose content keys the provisioned disk so it rebuilds itself on change, and `GOCACHE` lives on the share so builds typed at the guest prompt stay incremental across boots. Shares are writable by any guest identity, including root under sudo, because virtiofsd squashes guest uids/gids to the invoking user via `--translate-uid`/`--translate-gid` -- the reason virtiofsd 1.13+ is required. The harness locates the daemon itself, probing PATH, `~/.local/bin` and `/usr/libexec` for a binary that supports the option (`VIRTIOFSD=<path>` overrides), so neither it nor the installer ever symlinks anything onto PATH. The harness also runs on aarch64 hosts (probing the UEFI firmware locations distributions disagree on), and `VM_ARCH` gives a TCG-emulated cross-architecture smoke test. The remaining knobs (`VIRTFS_FAST`, `VM_CPUS`, `VM_MEMORY`, `VM_FIRMWARE`, `SERIAL_LOG`, `SSH_PORT`) and the full details are documented in `hack/README.md`.

## Testing

Native host (bpf-LSM kernel): the full script corpus at `STRESS_COUNT=5`, 590/590 subtests passing. Nested VM: the complete driver -- build, unit tests, gRPC e2e suite, corpus at five iterations -- green end to end, and the workflow ran repeatedly green on the development branch, including four concurrent runs to recreate runner contention. The detach fix is falsified both ways: with the kernel-link wait disabled the after-detach flake reproduces in CI (counter exactly doubled); with the wait plus the quiescence polls it has not been observed since. The aarch64 path is exercised on a Fedora 44 Asahi host: all three CI stages run green in the nested guest, 605 subtests passing, the 55 skips being the `external`-labelled image scripts the default selector excludes.

### DIFF
--- a/.github/workflows/go-ci.yaml
+++ b/.github/workflows/go-ci.yaml
@@ -22,7 +22,7 @@ env:
 jobs:
   check-vendor:
     name: Check go.mod and vendor
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v7.0.1
 
@@ -38,7 +38,7 @@ jobs:
 
   check-fmt:
     name: Check Go formatting
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v7.0.1
 
@@ -54,7 +54,7 @@ jobs:
 
   check-goimports:
     name: Check Go imports
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v7.0.1
 
@@ -70,7 +70,7 @@ jobs:
 
   check-vet:
     name: Check go vet
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v7.0.1
       - uses: docker/setup-buildx-action@v4
@@ -78,7 +78,7 @@ jobs:
 
   check-gofix:
     name: Check go fix modernisers
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v7.0.1
       - uses: docker/setup-buildx-action@v4
@@ -86,7 +86,7 @@ jobs:
 
   check-bpfman-shell-fmt:
     name: Check .bpfman formatting
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v7.0.1
       - uses: docker/setup-buildx-action@v4
@@ -100,7 +100,7 @@ jobs:
       matrix:
         include:
           - arch: amd64
-            runner: ubuntu-latest
+            runner: ubuntu-24.04
           - arch: arm64
             runner: ubuntu-24.04-arm
     steps:
@@ -116,7 +116,7 @@ jobs:
 
   lint:
     name: Lint
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v7.0.1
       - uses: docker/setup-buildx-action@v4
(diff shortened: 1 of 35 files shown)


frobware:
cc @alebedev87

alebedev87 (approved):
LGTM

PR #1693, opened 2026-07-28
Drop in-process link tracking 

frobware:
The daemon kept every probe-style link FD it attached in an in-memory map so detach could close it later. That map was broken: state that could not survive the process holding it, in a daemon whose store and bpffs are supposed to be the truth. Nothing rebuilt it after a restart, so the same link took a different detach route depending on the daemon's uptime history, and the held FDs made the daemon a silent co-owner of every tracked link's liveness, recorded nowhere. This PR removes the map and makes detach's contract -- returned means the kernel link object is gone -- both enforced and observable.

With the map gone, the bpffs pin is a link's only reference in daemon and CLI mode alike: attach closes its FD once the pin exists, and detach always reopens an FD from the pin, holds it across the pin removal, and closes it last. That ordering is what makes teardown synchronous. In the kernel (`kernel/bpf/syscall.c`, verified at v6.6 and v6.18-rc7), an FD close frees the link synchronously inside `close(2)` via `bpf_link_put_direct` -- hook detach included -- while every other reference drop, a pin removal among them, defers the free to a workqueue. Holding our FD across the pin's put means only our own close can drop the last reference, so by the time detach returns the program is off its hook. The old split had two bugs this structure removes: the reopen-from-pin route closed its FD before the pin removal, handing the last drop to the deferred path so a detached program could keep firing briefly after detach returned, and the tracked route could keep a zombie alive -- after an out-of-band unpin, a daemon-held FD left the link attached and firing, invisible to bpfman's own state, until the daemon exited.

Detach now also fails when it cannot honour that contract: a pin that will not open for any reason other than ENOENT, or a failed Info read, aborts the detach with an error rather than warn-and-continue, and the link record survives for a retry. Attach-time link closes log their error rather than discarding it -- on cilium's tracefs fallback path a failed close leaks the kprobe_events/uprobe_events entry with no owner left to reap it.

The post-close wait (`waitKernelLinkGone`) is reworked around the same kernel facts. Its old three-state model was inverted: EAGAIN from `BPF_LINK_GET_FD_BY_ID` means a link mid-creation, not one dying; a dying link reports ENOENT while its free is still pending, and the free removes the link's ID before it detaches the hook, so the old loop could return during exactly the window it was supposed to wait out. The wait now keys on the first probe: ENOENT there means our own close already completed the teardown, and the wait returns at once; any later ENOENT is ambiguous -- an external holder may have released the link through the deferred put, and an earlier EAGAIN or transient error may have masked the state -- and observes a short settle. The caller states whether its close provably performed the last drop, so the branch where the pin vanished out-of-band disables the fast path outright. Transient lookup errors log once per wait, the deadline warning fires only when the link is genuinely still present, the poll backoff respects its cap, and the tuning lives in an injectable `linkWaitParams`.

On the e2e side, the net detach-quiescence probe no longer fails the whole script on a single lost ping: probe failure retries within the poll window, the 0.3s ping timeout fits several attempts where one second fitted at most two, and the retry message carries the ping's exit code and stderr so a broken data path names itself. A new `assert_kernel_link_gone` helper probes, via the `linkinfo` builtin, that the link's kernel ID is no longer openable the instant `bpfman link detach` returns, with preconditions that stop it passing vacuously when no kernel ID was recorded; it is applied once per distinct detach path (the per-kind round-trip scripts and netns variants, one multi-prog wave, one dispatcher lifecycle script -- 17 sites). Its reach is deliberately bounded: script bpfman commands run as short-lived subprocesses, so the probe catches references that outlive the command -- a surviving pin, or a daemon-held FD in a daemon-mode run -- while post-detach firing through the kernel's deferred-release window remains the quiescence polls' domain. The instruments divide the labour accordingly: the polls prove the counter stops moving under stimulus, the delta assertions keep exact attribution while attached, and `linkinfo` guards the reference accounting.

The gRPC e2e suite gains the assertions only a long-lived daemon can support. The script corpus's subprocess model means process exit releases any leaked reference before an assertion can observe it, so reference-lifetime bugs are structurally invisible there; the gRPC suite runs one persistent `bpfman serve` process, where they are not. Three detach-contract tests read the daemon directly: its fd table must hold no bpf link FDs while a link is attached (with a self-check that opens a pin in the test process and requires the matcher to count it, so the zero can never go vacuous); the kernel link ID must be unopenable the instant the Detach RPC returns; and an out-of-band pin removal must release the link rather than leave a zombie. A fourth, end-state check runs on the parallel stress phase's cleanup: once every lifecycle completes, the daemon must hold no bpf object FDs -- programs and maps included -- and manage no programs. Leaks accumulate, so a reference leaked by any of the ~1000 lifecycles, in any interleaving, is still visible when the suite goes quiet, and failures name the held objects via fdinfo.

That end-state check immediately found daemon-held state predating this PR: the adapter cached one test dispatcher per network type as the verifier target for extension loads, so a quiescent daemon carried two dispatcher program FDs and two map FDs for its lifetime. The cache is removed -- the target now loads per extension load and closes when the load returns, the kernel holding its own reference (`prog->aux->dst_prog`) for as long as the extension exists. A quiescent daemon now holds no bpf object FDs at all, which is the invariant the end-state check pins permanently.

One architectural observation worth recording. The reference-persistence bug family -- the liveLinks map, the dispatcher cache, any future leaked FD -- exists only because the daemon is a single long-lived process; the CLI topology is immune by construction, since process exit releases everything. With this PR the daemon holds no state beyond the store and bpffs, so an exec-per-request daemon (the gRPC surface forking the CLI per mutation) becomes a viable simplification that would eliminate the family outright, at the cost of a fork/exec per control-plane operation. It would not remove the need for the detach contract itself -- the synchronous teardown ordering matters identically in a short-lived process -- but it would make leak persistence impossible rather than merely tested-for. Not done here; noted as a direction the now-stateless daemon leaves open.

## Testing

Unit suite green (race detector on) and golangci-lint clean. The full `.bpfman` script corpus passes natively as root against this branch, and the full gRPC suite passes at default scale (1024 lifecycles across eight program types) including the end-state quiescence check. The wait behaviours -- the first-probe fast path with the production constants, the settle after an alive observation, after EAGAIN, after a transient error and on the pin-vanished branch, and the deadline, gone-at-deadline, transient-error and cancellation paths -- are pinned by unit tests driving the loop through injected fakes. The kernel-ordering claims are read directly from `kernel/bpf/syscall.c` at v6.6 and v6.18-rc7.

The new gRPC tests are red/green demonstrated rather than assumed: the while-attached fd scan and the zombie check fail against a daemon built from the pre-change implementation (one bpf_link FD held per attached link; the kernel link still alive three seconds after an out-of-band unpin), the detach-return check fails against a build that leaks the detach FD, and the end-state check fails on both implementations until the dispatcher cache is removed -- the pre-change daemon is otherwise tidy at rest, which is exactly why the while-attached probe exists. The nested-VM job has not been run against this branch and should gate merging.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -841,7 +841,7 @@ $(BIN_DIR)/e2e.test: $(DISPATCHER_BPF_EMBEDS) $(E2E_BPF_OBJECTS) | $(BIN_DIR)
 .PHONY: test-e2e
 test-e2e: $(BIN_DIR)/e2e.test
 	$(Q)$(MAKE) e2e-kmod-reload
-	sudo BPFMAN_E2E_BYTECODE_DIR=$(abspath e2e) $(call forward-env,BPFMAN_E2E_ISOLATED_RUNTIME BPFMAN_E2E_POLICY_RULE_PREF BPFMAN_LOG) $(BIN_DIR)/e2e.test -test.v -test.failfast -test.count=$(STRESS_COUNT) $(if $(filter-out 0,$(PARALLEL)),-test.parallel $(PARALLEL)) $(if $(TEST),-test.run $(TEST))
+	sudo BPFMAN_E2E_BYTECODE_DIR=$(abspath e2e) $(call forward-env,BPFMAN_E2E_ISOLATED_RUNTIME BPFMAN_E2E_POLICY_RULE_PREF BPFMAN_LOG) $(BIN_DIR)/e2e.test -test.v -test.failfast -test.count=$(STRESS_COUNT) $(if $(filter-out 0,$(PARALLEL)),-test.parallel $(PARALLEL)) $(if $(TEST),-test.run "$(TEST)")
 
 # Parallel gRPC e2e: stands up a real `bpfman serve` subprocess and
 # fans goroutines through load/get/attach/detach/unload over the
@@ -904,7 +904,7 @@ run-e2e-grpc:
 	    BPFMAN_E2E_BYTECODE_DIR=$(E2E_GRPC_BYTECODE_DIR) \
 	    $(call forward-env,$(E2E_GRPC_FORWARD_VARS)) \
 	    $(E2E_GRPC_TEST_BIN) -test.v -test.failfast \
-	    -test.count=$(STRESS_COUNT) $(if $(TEST),-test.run $(TEST))
+	    -test.count=$(STRESS_COUNT) $(if $(TEST),-test.run "$(TEST)")
 
 .PHONY: test-e2e-grpc
 test-e2e-grpc: build-e2e-grpc
--- a/e2e/grpc/detach_contract_test.go
+++ b/e2e/grpc/detach_contract_test.go
@@ -1,0 +1,274 @@
+//go:build e2e
+
+// Detach-contract tests: daemon-process properties that only a
+// long-lived `bpfman serve` process can exhibit. The script corpus
+// runs every bpfman command as a short-lived subprocess, so process
+// exit releases any leaked FD before an assertion can observe it;
+// here the daemon outlives every RPC, so reference-lifetime bugs are
+// visible in its /proc/<pid>/fd table and in how long kernel link
+// objects survive.
+//
+// Three properties, one test each:
+//
+//   - The daemon holds no bpf link FDs while a link is attached: the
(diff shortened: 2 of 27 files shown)


frobware:
cc @alebedev87

PR #1694, opened 2026-07-28
Modernise the examples with go fix

frobware:
The bpfman-gofix target has excluded the example applications since they were absorbed into the root module, with the stated intent of keeping them as their authors wrote them. This PR ends that grace period: the go1.26.4 fix modernisers run over `examples/...`, dropping the legacy `// +build` constraint lines that the `//go:build` form superseded and rewriting `go-app-counter`'s six-goroutine fan-out from the manual `Add`/`Done` pattern to `sync.WaitGroup.Go` (the comment describing the old counter bookkeeping goes with it). The second commit removes the package-list filter from the Makefile target, so the moderniser gate in CI keeps the examples current from here on.

## Testing

`go build ./examples/...` and `go vet ./examples/...` are clean, and re-running the now-inclusive `make bpfman-gofix` produces no further changes, so the CI moderniser gate passes by construction.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1251,8 +1251,7 @@ bpfman-vet: $(DISPATCHER_BPF_EMBEDS)
 # applications are excluded: they stay as their authors wrote them.
 .PHONY: bpfman-gofix
 bpfman-gofix: $(DISPATCHER_BPF_EMBEDS)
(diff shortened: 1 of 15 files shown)


frobware:
cc @alebedev87

PR #1695, opened 2026-07-29
Re-check cached images when the signing policy changes

frobware:
The OCI puller consults the image cache before it reaches the signature verifier, and the cache entry records nothing about the policy that admitted it. An image pulled whilst `allow_unsigned = true` therefore keeps loading after that permission is withdrawn: the verifier is never asked, and nothing is logged to say a policy decision was skipped. Only `PullAlways`, or clearing the node's cache, produces the intended refusal. The same applies to `[[signing.trusted_identities]]` -- adding a constraint to exclude a signer does not evict what that signer's images already put in the cache.

This matters for any move to make signature verification mandatory by default, which is currently under discussion. `allow_unsigned` defaults to `true` in `config/default.toml` and in what bpfman-operator ships, so flipping it would appear to take effect whilst leaving every already-pulled image untouched -- which on a running cluster is every image in use.

Two properties would have to hold for skipping verification on a cache hit to be sound, and neither does. The entry records no verification state, so "signature verified", "unsigned, allowed by policy" and "verification disabled" are indistinguishable after the fact. And the cache is keyed by a hash of the image reference rather than by content, so a hit contacts no registry at all.

The fix gives `SignatureVerifier` a `PolicyID()` fingerprinting the decision it makes -- whether unsigned images pass, and which signers count -- recorded in the cache entry. On a hit, an entry admitted under the policy still in force is served untouched, which is the steady state and costs nothing; anything else is verified again and the new decision recorded, so the check happens once per policy change rather than once per load. Verifying on every load would be simpler but puts Rekor and the certificate transparency log in the hot path of every program load, which is presumably why the check sits where it does.

Comparing fingerprints for equality rather than ordering policies by strictness avoids needing a definition of when one set of trusted identities is stricter than another. Re-verification asks about the digest the verifier was originally asked about, which for a multi-platform image is the index digest cosign signs, not the platform child the cache holds -- verifying the child would reject an image the policy should accept. An entry recording no policy is never admitted, so entries cached before this existed, or by a verifier that names its policy with the empty string, are pulled again and put through the full check. `PullNever` refuses rather than pulling, since serving bytecode the current policy has never passed is the one outcome worse than failing.

A verifier must name its policy: `NewPuller` refuses one whose `PolicyID()` is empty, since the cache reserves the empty policy for images admitted by no verifier at all, and a verifier answering with it could not tell its own decisions from those.

Recording the new policy after a re-verification is best effort. It is a read-modify-write on the cache entry with nothing serialising it -- image pulls run outside the writer flock by design, so two loads of the same image can overlap -- and a lost update costs a repeated verification on the next load rather than admitting anything unverified. The wider race, where concurrent pulls of one image both download, both write the bytecode and both rewrite the metadata, predates this change and is tracked in #1696.

## Test plan

`TestPullerVerifiesCachedImageAgainstCurrentPolicy` reproduces the original bug: pull an image through a permissive verifier, then pull it again through one that refuses everything, and assert both that the verifier is consulted and that the load fails. Before the fix it failed on both counts; with the second pull switched to `PullAlways` it passed, confirming the pull policy was the only variable.

`TestPullerServesCachedImageWithoutRepeatingVerification` pins the steady state -- two loads under one unchanged policy consult the verifier once, so sigstore stays off the load path. `TestPullerReverifiesTheDigestThatWasSigned` publishes a two-platform image and asserts the re-verified subject matches the originally verified one; it fails loudly if the index and child digests coincide, so it cannot silently prove nothing. `TestPullerDoesNotAdmitUnrecordedEntryToAnEmptyPolicy` caches an image with no verifier, then loads it through a verifier whose `PolicyID()` is empty, and asserts the verifier is consulted.

Each new test was checked against a reverted version of the clause it covers, to confirm it fails for the intended reason. `make lint-go` reports no issues; `go build ./...` and `go test ./platform/... ./config/... ./cmd/...` pass.

Not covered: none of this runs against real sigstore infrastructure -- every test uses a stub verifier, so the cosign integration itself is unchanged and unexercised here.

### DIFF
--- a/platform/image/oci/puller.go
+++ b/platform/image/oci/puller.go
@@ -39,6 +39,18 @@ type cachedMetadata struct {
 	Programs map[string]string `json:"programs,omitempty"`
 	Maps     map[string]string `json:"maps,omitempty"`
 	PulledAt time.Time         `json:"pulled_at"`
+
+	// Policy is the SignatureVerifier.PolicyID that admitted this
+	// entry. Empty on entries cached before the policy was
+	// recorded, or with no verifier configured, which no policy
+	// matches, so those entries are pulled again on the next hit.
+	Policy string `json:"policy,omitempty"`
+
+	// VerifiedDigest is the image the verifier was asked about. For
+	// a multi-platform image that is the index digest, which cosign
+	// signs, and so differs from Digest above once a platform child
+	// has been selected.
+	VerifiedDigest string `json:"verified_digest,omitempty"`
 }
 
 // puller implements ImagePuller using ORAS for OCI registry access.
@@ -87,6 +99,13 @@ func NewPuller(cache fs.EnsuredImageCache, opts ...Option) (platform.ImagePuller
 		}
 	}
 
+	// The cache reserves the empty policy for images admitted by no
+	// verifier, so a verifier answering with it could never record a
+	// reusable entry. Refuse here rather than degrade quietly later.
+	if p.verifier != nil && p.verifier.PolicyID() == "" {
+		return nil, fmt.Errorf("signature verifier reports an empty policy id")
+	}
+
 	p.logger.Debug("initialising OCI puller", "cache_dir", p.cache.Root())
 
 	return p, nil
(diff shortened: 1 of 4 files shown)


frobware:
cc @alebedev87

